PR #10, opened 2018-05-19
Export route tables on Network

### LINKED ISSUE #3: Expose route tables on Network
We are currently not exposing the route tables for a network as properties.  This has led to problems for a user who was trying to create VPC Endpoints for S3 on top of a `Network`, which requires access to the route tables.

The workaround we shared was below - but the middle part of this is complex and should not be required:

```javascript
let network = new awsinfra.Network(`${prefix}-net`, {
    numberOfAvailabilityZones: config.availabilityZones, // Create subnets in many AZs
    usePrivateSubnets: true,                                // Run compute inside private subnets in each AZ
});

// Define the policy for exposing an S3 VPC endpoint
let endpointPolicy = {
    Version: "2012-10-17",
    Statement: [{
        Action: "*",
        Effect: "Allow",
        Principal: "*",
        Resource: "*",
    }, {
        Action: "*",
        Effect: "Allow",
        Principal: "*",
        Resource: [
            "arn:aws:s3:::*.amazonaws.com",
            "arn:aws:s3:::*.amazonaws.com/*",
        ],
    }],
}

// Lookup the route table for a subnet
async function getRouteTableIdForSubnet(subnetId) {
    let res = await aws.ec2.getRouteTable({
        subnetId: subnetId,
    });
    return res.routeTableId;
}

// Find all route tables for the Network (one for each private subnet, and the shared route table for the public subnets)
let routeTableIds = pulumi.all([...network.subnetIds, network.publicSubnetIds[0]]).apply(subnetIds => {
    return Promise.all(subnetIds.map(getRouteTableIdForSubnet));
});

// Create a VPC Endpoint in the VPC applied to the 
let vpcEndpointS3 = new aws.ec2.VpcEndpoint("s3-endpoint", {
    policy: JSON.stringify(endpointPolicy),
    routeTableIds: routeTableIds,
    vpcId: network.vpcId,
    serviceName: `com.amazonaws.${aws.config.region}.s3`,
});
```

lukehoban:
Fixes #3.

### DIFF
--- a/nodejs/aws-infra/network.ts
+++ b/nodejs/aws-infra/network.ts
@@ -40,6 +40,9 @@ export interface NetworkVpcArgs {
     readonly publicSubnetIds: pulumi.Input<string>[];
 }
 
+// The lazily initialized default network instance.
+let defaultNetwork: Network;
+
 /**
  * Network encapsulates the configuration of an Amazon VPC.  Both [VPC with Public
  * Subnet](https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Scenario1.html) and [VPC with Public and Private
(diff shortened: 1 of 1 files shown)


joeduffy:
@lukehoban Should we merge this?

houqp:
@lukehoban do you know why this got removed from the latest master? We happened to need this for setting up vpc peering.

lukehoban:
@houqp Which piece specifically was removed? I know there's some reorganization here that is not yet released, but we definitely can't regress overall functionality here.

cc @CyrusNajmabadi

houqp:
We would need `publicRouteTableId` or `privateRouteTableIds` to set the routes for both vpcs.

PR #16, opened 2018-06-12
Add aws-infra Network implementation in Python

swgillespie:
This code came out of last Thursday's hackathon and serves as an example and test that we can write non-trivial components in Python.

This code is translated pretty much verbatim from the TypeScript implementation in this repo. All of this code is known to work with tip-of-master builds of `pulumi` and `pulumi-aws`, but we'll need to do another pass once the `v0.14` SDK drops.

Follow-up PRs will do the plumbing required to publish this package to PyPI.

### DIFF
--- a/python/pulumi_aws_infra/network.py
+++ b/python/pulumi_aws_infra/network.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """
 The network module provides the Network class, which is a Python component that creates
 a VPC-based network that conforms with AWS best practices.
@@ -21,6 +20,7 @@ from pulumi.resource import ComponentResource, ResourceOptions
 from pulumi.errors import RunError
 from pulumi_aws import ec2
 from .util import get_aws_az
+
 
 class Network(ComponentResource):
     """
@@ -259,7 +259,7 @@ class Network(ComponentResource):
             subnets = ec2.get_subnet_ids(vpc_id=vpc.id)
             default_security_group = ec2.get_security_group(
                 name="default", vpc_id=vpc.id)
-            net = Network.from_vpc(
+            net = Network(
                 "default-vpc",
                 vpc_id=vpc.id,
                 subnet_ids=subnets.ids,
@@ -269,6 +269,5 @@ class Network(ComponentResource):
                 opts=resource_options)
 
             Network._default_network = net
-            return net
 
         return Network._default_network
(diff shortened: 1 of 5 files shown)


swgillespie:
I'm reviving this today. I'll push a new iteration in a little bit.

swgillespie:
I think I've addressed all of the feedback from last week. I've tried to rework all of the code to adhere to `pylint` and the Google style guide for python. The code is all formatted using https://github.com/google/yapf.

python/pulumi_aws_infra/network.py
```diff
@@ -114,7 +114,7 @@ def _create_from_arguments(self, name, number_of_availability_zones):
         :return: None
         """
         number_of_availability_zones = number_of_availability_zones or 2
-        if number_of_availability_zones < 1 or number_of_availability_zones > 4:
+        if number_of_availability_zones < 1 or number_of_availability_zones >= 4:
```
joeduffy: This is driven by the AZs available in a region, right?  Or is there some other limitation here?

swgillespie: Yes, that is true.

lukehoban: Opened https://github.com/pulumi/pulumi-aws-infra/issues/21 to track potentially changing this to compute it from the target region.  It's not entirely trivial though, as we would then need to make an async call (which will likely in the future be a call that returns an `Output<T>`) before running the loop that creates each resource.  Interestingly, in Python that may not be as big a deal as it is in JavaScript - but I suspect we'll want to be consistent here across the two.

python/setup.py
```diff
@@ -0,0 +1,27 @@
...
+setup(
+    name='pulumi_aws_infra',
+    version='${VERSION}',
+    description='Pulumi Amazon Web Services (AWS) infrastructure components.',
+    keywords='pulumi aws aws-infra',
+    url='https://github.com/pulumi/pulumi-aws-infra',
+    project_urls={'Repository': 'https://github.com/pulumi/pulumi-aws-infra'},
+    packages=find_packages(),
+    install_requires=[
+        'pulumi>=0.14.0,<0.15.0', 'pulumi_aws>=0.14'
```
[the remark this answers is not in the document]
swgillespie: The `<0.15.0` is closest to the npm-style carat dependency. If we stick to our current versioning scheme, 0.15.0 won't be compatible with 0.14.0 and we'd be wrong to accept 0.15.0 as a dependency for this library.

I think we could relax this if we don't make any breaking changes to this library in 0.15, but I'm not sure we should relax this constraint until we are sure that will be the case.

swgillespie:
I'm going to go ahead and merge this so we can start trying to hook up the publishing pipeline for this package - please let me know if there are any other changes anyone would like me to make.

swgillespie:
thanks everyone for the reviews!

PR #39, opened 2018-10-23
Fix network registration in Cluster.

CyrusNajmabadi:


CyrusNajmabadi:
Tagging @lukehoban .  I'm going to do a write-up in the PR explaining the root cause of hte problem (it's hairy...).  I'm thinking this is the right fix.  But i want to discuss it with you.

nodejs/aws-infra/cluster.ts
```diff
@@ -129,7 +133,6 @@ export class Cluster extends pulumi.ComponentResource {
         }
 
         super("aws-infra:cluster:Cluster", name, {
-            network: args.network,
```
CyrusNajmabadi: this is actually the critical line that is a problem. Here's what's going on:  'args.network' is effective a pojo, with keys that map to output-properties.  For example, it has a `{ ... vpcId: Output<string> ... }` property.  And, specifically, during preview, this is `vpcId: Output(value: undefined, isKnown: false)`

The way we handle a pojo is to serialize out all it's properties here: https://github.com/pulumi/pulumi/blob/72a6ed320c5982b1852c4d953feb912477b9fad7/sdk/nodejs/runtime/rpc.ts#L77-L110

For Cluster/Network, this means (during preview): we register a 'network' property with the shape `{ ... vpcId: unknownValue .. .}`.  That's all fine.  What's not fine is when we then deserialize this back when the engine returns the resultant output properties.  The issue here is that we have no way to indicate any sort of 'nested output values'.  Instead, the engine simply reports back that same shape `{ ... vpcId: unknownValue ... }`.  On our end, we then translate that to:

```Output<{ ... vpcId: undefined ... }>```, instead of the more accurate  `Output<{ ... vpcId: Output(value: undefined, isKnown: false) ... }>` that we started with.  Effectively, even though we're in preview, we "lose" that ths was an output, and we "lose" that the value should not be considered "known".  That means that anything downstream that attempts to use this vpcId will fail durin preview (if that value is required).  

The core issue here (in the lowest layer) is that we really don't have a concept of nested-outputs.  Objects are flattened with all their values transfered over, and when rehydrated, we don't know how to appropriate reconstitute things.  We only make top level 'Outputs' and we don't have any way to indicate that nested properties are themselves Outputs.

--

Making a low level change to try to represent this seems both scary, and a lot of work.  It would require appropriate info being marshalled to/from the engine, and it just makes a super complex area more complex.  A *much* simpler fix is to just do what i did above.  Do not try to register the 'network' object as a prop that is serialized/deserialied during created.  Instead, just assign the property to the type, and register it at the end.  This way the engine knows everything about the object, but the in-memory representation in the JS side is kept in a good state.

--

Note: this is a potential pitfall for anyone consuming the values of a *pojo* prop of any resource.  If the engine has returned a value for the outer prop, then we will consider the value 'known', but it will contain inner values that can be unknown depending on if they're computed.  However, instead of properly being Outputs with the isKnown bit set to false, they'll just be JS 'undefined values'.

CyrusNajmabadi: Tagging @lukehoban @joeduffy @pgavlin  about the above ^

CyrusNajmabadi:
Ok.  I'm going to move forward with this fix.  The idea of changing the low-level serialization logic is unpalatable for me.

pgavlin:
> this is actually the critical line that is a problem

Can you elaborate a bit on what the user-facing effects of the bug? It's not clear to me exactly what behavior we were seeing.

CyrusNajmabadi:
@pgavlin The main issue was that a customer was passing this 'network' value to a 'cluster'.  Then, later, they effectively did "cluster.network.vpcId" and passed that to some other resource.

However, because of how serialization/deserialization works, this vpcId was not an Output(unknown), but was instead just a straight up 'undefined' value.  This meant that 'undefined' got passed to a later resource, which then bottomed out in terraform, saying that it wasn't valid to pass no value for that property.

PR #51, opened 2018-11-25
Produce components at the aws-infra layer for Services/Tasks

CyrusNajmabadi:
Ready for review.

Overall summary:  This PR introduces several powerful and flexible component resources at the aws-infra layer.  It also introduces a few design patterns we can flesh out and document to help guide out we do future components.

The major components being added are:

1. Cluster.
2. Service (including EC2 and Fargate versions). 
3. TaskDefinition (including EC2 and Fargate versions)
4. FileSystem (for clusters)
5. AutoScalingGroups (for clusters)
6. Containers

This pr also introduces a new 'provider' pattern for allowing components to be created without taking hard dependencies on specific convenience patterns.  For example, TaskDefinitions do not have hard coded knowledge on how to make images.  That is extracted into an IImageProvider interface that allows us to build flexible solutions to make things easier, while not restricting users from doing things any way the may want this.  

This PR has providers for images, as well as for load balancers.


nodejs/aws-infra/aws.ts
```diff
@@ -16,13 +16,19 @@
...
+    return azs[index];
+}
+
+export function getAvailabilityZones(): Promise<string[]> {
+    if (!zones) {
+        zones = aws.getAvailabilityZones().then(r => r.names);
     }
-    return (await azs).names[index];
+
+    return zones;
```
CyrusNajmabadi: this logic is effectively unchanged.  i just cleaned it up a bit.

nodejs/aws-infra/aws_test.go
```diff
@@ -40,17 +53,34 @@ func Test_Examples(t *testing.T) {
...
+		// {
+		// 	Dir: path.Join(cwd, "./examples/cluster"),
+		// 	Config: map[string]string{
+		// 		"aws:region":     region,
+		// 		"cloud:provider": "aws",
+		// 	},
+		// 	Dependencies: []string{
+		// 		"@pulumi/aws-infra",
+		// 	},
+		// },
```
CyrusNajmabadi: temporary to speed up test results.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,132 @@
...
+import * as mod from ".";
+
+import { Network } from "./../network";
+
+import * as utils from "../utils";
+
+/**
+ * A Cluster is a general purpose ECS cluster configured to run in a provided Network.
+ */
+export class Cluster extends pulumi.ComponentResource {
```
CyrusNajmabadi: Decision: Even though there is a aws resource with thsi name, we use the same name.  This is in our 'components' package.  And it helps immediately make clear the general purpose for this type.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,132 @@
...
+    public readonly instance: aws.ecs.Cluster;
```
CyrusNajmabadi: Decision 1.  When wrapping a same-named aws resource, i've called the wrapped type 'instance'.  This is to prevent ugliness like `cluster.cluster...`.  i am not wedded to this decision.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,132 @@
...
+        const ALL = {
+            fromPort: 0,
+            toPort: 0,
+            protocol: "-1",  // all
+            cidrBlocks: [ "0.0.0.0/0" ],
+        };
+
+        // IDEA: Can we re-use the network's default security group instead of creating a specific
+        // new security group in the Cluster layer?  This may allow us to share a single Security Group
+        // across both instance and Lambda compute.
```
CyrusNajmabadi: i copied this comment, but it still applies.  The network does have 'securityGroups'.  Should we consider using that?

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,132 @@
...
+        const instanceSecurityGroup = args.instanceSecurityGroup || new aws.ec2.SecurityGroup(name, {
```
CyrusNajmabadi: General pattern: if we're going to create something on behalf of the user, also allow them to pass in a value to be used instead.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,132 @@
...
+                // Expose ephemeral container ports to Internet.
+                // TODO: Limit to load balancer(s).
+                {
+                    fromPort: 0,
+                    toPort: 65535,
+                    protocol: "TCP",
+                    cidrBlocks: [ "0.0.0.0/0" ],
+                },
+            ],
+            egress: [ ALL ],  // See TerraformEgressNote
```
CyrusNajmabadi: todo: both of these defaults (for ingress and egress) should be exposed through this type so that people can get access to it and provide a slightly augmented version if they want.  Otherwise, they need to greatly duplicate this structure.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,132 @@
...
+            tags: { Name: name },
+        }, parentOpts);
+
+        this.instance = instance;
+        this.network = network;
+        this.instanceSecurityGroup = instanceSecurityGroup;
```
CyrusNajmabadi: design: we expose all the values we created as well.  that way people can always access them.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,132 @@
...
+
+(<any>Cluster).doNotCapture = true;
+
+// The shape we want for ClusterArgs.  We don't export this as 'Overwrite' types are not pleasant to
+// work with. However, they internally allow us to succinctly express the shape we're trying to
+// provide. Code later on will ensure these types are compatible.
+type OverwriteShape = utils.Overwrite<aws.ecs.ClusterArgs, {
+    network: Network;
+    instanceSecurityGroup?: aws.ec2.SecurityGroup;
+}>;
```
CyrusNajmabadi: implementation design:  We use the 'overwrite' shape internally to succinctly define the we want in terms of a change to an underlying args type.  we then expose a real interface with all the arg-type properties copied into it.  We then ensure these types are compatible below.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+// limitations under the License.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as mod from ".";
+
+import { Overwrite, sha1hash } from "./../utils";
+
+export class ClusterAutoScalingLaunchConfiguration extends pulumi.ComponentResource {
```
CyrusNajmabadi: Open question on the naming of these guys.  Should we just call it LaunchConfiguration since that's the relevant thing it will create.  Or should the name guide you wrt to expected intent.  i.e. this is to make a launch config intended to help cluster autoscale services within it?

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+        this.instanceProfile = instanceProfile;
+        this.fileSystem = fileSystem;
+
+        this.registerOutputs({
+            instance,
+            cluster,
+            stackName,
+            instanceProfile,
+            fileSystem,
+        });
```
CyrusNajmabadi: general pattern i'm following is to do all initialization of resources first, then assign the resources to 'this', then 'registerOutputs'.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+            instanceType: pulumi.output(args.instanceType).apply(t => t || "t2.micro"),
+            keyName: args.keyName,
+            iamInstanceProfile: instanceProfile.id,
+            enableMonitoring: pulumi.output(args.enableMonitoring).apply(b => b !== undefined ? b : true),
+            placementTenancy: pulumi.output(args.placementTenancy).apply(t => t || "default"),
+            rootBlockDevice: pulumi.output(args.rootBlockDevice).apply(d => d || defaultRootBlockDevice),
+            ebsBlockDevices: pulumi.output(args.ebsBlockDevices).apply(d => d || defaultEbsBlockDevices),
+            securityGroups: pulumi.output(args.securityGroups).apply(g => g || [ cluster.instanceSecurityGroup.id ]),
+            userData: getInstanceUserData(cluster, args, stackName),
+        }, parentOpts);
```
CyrusNajmabadi: General pattern:

1. 'splat' the `args` object.  That way, users can provide all values they want, and we'll respect them.
2. provide jam in all values we directly control production of.
3. for values where we can provide sensible defaults, defer to the user if they provided any, otherwise, provide our default.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+        volumeSize: 5, // GiB
+        volumeType: "gp2", // default is "standard"
+        deleteOnTermination: true,
+    }, {
+        // Docker image and metadata volume
+        deviceName: "/dev/xvdcz",
+        volumeSize: 50, // GiB
+        volumeType: "gp2",
+        deleteOnTermination: true,
+    }];
```
CyrusNajmabadi: this should be exposed in some way.  that way if the user wants these, but wants to add another device, they can do so without having to copy this all over into their code.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+                # See http://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html
+                iptables --insert FORWARD 1 --in-interface docker+ --destination 169.254.169.254/32 --jump DROP
+                service iptables save
+
+                /opt/aws/bin/cfn-signal \
+                    --region "\${AWS_REGION}" \
+                    --stack "${cloudFormationStackName}" \
+                    --resource Instances
+        `;
+    });
```
CyrusNajmabadi: todo.  is there a sensible way we can provide all this code, wihle still allowing customers to extend this sensibly?

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+
+        let launchConfiguration: ClusterAutoScalingLaunchConfiguration;
+
+        // Use the autoscaling config provided, otherwise just create a default one for this cluster.
+        if (args.launchConfiguration) {
+            launchConfiguration = args.launchConfiguration;
+        }
+        else {
+            launchConfiguration = new ClusterAutoScalingLaunchConfiguration(
+                name, args.launchConfigurationArgs || { cluster }, parentOpts);
```
CyrusNajmabadi: another pattern.  We allow user to provide either `launchconfiguration' (the resource) or `launchconfigurationargs`.  That way, they can easily construct child objects, with a simple args object, without having to instantiate the value inline.  We may not want this as it's unclear how deep this should go, and if we shuld supply this for all  child resources.

A very common one this is good for though is "Service+TaskDefinition".  it would be somewhat annoying (IMO) when creating a Service to manually have to make the underlying TaskDefinitions.  Instead, you can just inline the TaskDefArgs into the ServiceArgs.  We should discuss how we feel about this.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+                    Value: ${instanceName}
+                    PropagateAtLaunch: true
+            CreationPolicy:
+                ResourceSignal:
+                    Count: ${minSize}
+                    Timeout: PT15M
+            UpdatePolicy:
+                AutoScalingRollingUpdate:
+                    MaxBatchSize: 1
+                    MinInstancesInService: ${minSize}
```
CyrusNajmabadi: note: i'm providing the same level of customization that cloud currently provides.  However, we should likely make every single thing in this template something you can parameterize.  i.e. MaxBatchSize, Timeout, PropagateAtLaunch, etc.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,566 @@
...
+    /**
+     * The config to use when creating the auto scaling group.
+     *
+     * [launchConfiguration] or [launchConfigurationArgs] can be provided.  And, if either are
+     * provided will be used as the launch configuration for the auto scaling group.
+     *
+     * If neither are provided, a default instance will be create by calling
+     * [cluster.createAutoScalingConfig()].
+     */
+    launchConfigurationArgs?: ClusterAutoScalingLaunchConfigurationArgs;
```
CyrusNajmabadi: see above for discussion on why we supply launchconfiguration and launchConfigurationArgs.

nodejs/aws-infra/experimental/clusterFileSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+// limitations under the License.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as mod from ".";
+
+import * as utils from "../utils";
+
+export class ClusterFileSystem extends pulumi.ComponentResource {
```
CyrusNajmabadi: What should we name this?  just FileSystem?  Or 'ClusterFileSystem' since it is particularly a FS that is for a cluster and which needs a cluster in order to be properly configured?

nodejs/aws-infra/experimental/clusterService.ts
```diff
@@ -0,0 +1,337 @@
...
+    let loadBalancers: mod.LoadBalancers;
+    for (const containerName of Object.keys(args.taskDefinition.containers)) {
+        const container = args.taskDefinition.containers[containerName];
+        if (container.loadBalancerProvider) {
+            loadBalancers = combineLoadBalancers(
+                loadBalancers, container.loadBalancerProvider.loadBalancers(containerName, name, service));
+        }
+    }
+
+    return loadBalancers;
```
CyrusNajmabadi: another pattern.  Instead of populating the loadBalancers value itself.  The Service calls out to the .loadBalancerProvider to compute this based on the internal data it has.  This way the loadBalancerProvider can be implemented any way it wants to be.  It just needs to know how to return the 'loadBalancer' info that the Service needs.

nodejs/aws-infra/experimental/clusterService.ts
```diff
@@ -0,0 +1,337 @@
...
+//     }
+
+//     getVolumeName() {
+//         return utils.sha1hash(`${this.kind}:${this.path}`);
+//     }
+
+//     getHostPath() {
+//         return this.path;
+//     }
+// }
```
CyrusNajmabadi: Currently not implemented.  It may not be necessary at all given that users can just supply these values when creating the Service. But i haven't looked into this more to see how this all works.

nodejs/aws-infra/experimental/clusterService.ts
```diff
@@ -0,0 +1,337 @@
...
+    iamRole?: pulumi.Input<string>;
+    /**
+     * A load balancer block. Load balancers documented below.
+     */
+    loadBalancers?: pulumi.Input<pulumi.Input<{
+        containerName: pulumi.Input<string>;
+        containerPort: pulumi.Input<number>;
+        elbName?: pulumi.Input<string>;
+        targetGroupArn?: pulumi.Input<string>;
+    }>[]>;
```
CyrusNajmabadi: User can still provide this.  However, we'll combine whatever they provide with the data given by the loadBalancerProvider

nodejs/aws-infra/experimental/containerDefinition.ts
```diff
@@ -0,0 +1,153 @@
...
+export function computeContainerDefinition(
+    parent: pulumi.Resource,
+    name: string,
+    containerName: string,
+    container: ContainerDefinition,
+    logGroup: aws.cloudwatch.LogGroup): pulumi.Output<aws.ecs.ContainerDefinition> {
+
+    if (container.image === undefined && container.imageProvider === undefined) {
+        throw new Error("container requires either [image] or [imageProvider] to be set.");
+    }
```
CyrusNajmabadi: we allow users to directly provide image strings, or they give a provider that will compute this on demand, but leave Service/Task out of hte job of figuring out how to do this.

nodejs/aws-infra/experimental/containerDefinition.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (imageProvider) {
+        image = imageProvider.image(name, parent);
+        environment = utils.combineArrays(
+            environment, imageProvider.environment(name, parent));
+    }
+
+    if (loadBalancerProvider) {
+        portMappings = utils.combineArrays(
+            portMappings, loadBalancerProvider.portMappings(containerName, name, parent));
+    }
```
CyrusNajmabadi: image, environment, and portMappings can all be user provided, but they all can be augmented by providers as well.  for single-data properties (like `image`), they provider can just supply the value.  For multi-valued properties (like `environment` or `portMappings`) the provider can add values.

nodejs/aws-infra/experimental/ec2.ts
```diff
@@ -0,0 +1,146 @@
...
+// limitations under the License.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as utils from "../utils";
+
+import * as mod from ".";
+
+export type EC2TaskDefinitionArgs = utils.Overwrite<mod.TaskDefinitionArgs, {
```
CyrusNajmabadi: still using an overwrite here.  but this should be converted to an interface.  annying that we'll have to copy over all the properties another time.

nodejs/aws-infra/experimental/ec2.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        if (args.taskDefinitionArgs) {
+            throw new Error("[args.taskDefinitionArgs] should not be provided.");
+        }
+
+        return new mod.EC2Service(name, {
+            ...args,
+            taskDefinition: this,
+        }, opts || { parent: this });
+    }
```
CyrusNajmabadi: i still kept around this instance method.  It seems like a nice way to bridge between resources that are very strongly related.  But we an remove if we don't think this is valuable.

nodejs/aws-infra/experimental/fargate.ts
```diff
@@ -0,0 +1,198 @@
...
+// limitations under the License.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as utils from "../utils";
+
+import * as mod from ".";
+
+export type FargateTaskDefinitionArgs = utils.Overwrite<mod.TaskDefinitionArgs, {
```
CyrusNajmabadi: EC2 comments apply to Fargate as well.

nodejs/aws-infra/experimental/image.ts
```diff
@@ -0,0 +1,206 @@
...
+import * as utils from "../utils";
+
+import * as mod from ".";
+
+export type ImageEnvironment = Record<string, pulumi.Input<string>>;
+
+export interface IImageProvider {
+    image(name: string, parent: pulumi.Resource): pulumi.Input<string>;
+    environment(name: string, parent: pulumi.Resource): pulumi.Input<aws.ecs.KeyValuePair[]>;
+}
```
CyrusNajmabadi: a 'provider' is a general purpose way of flipping the onus of data production around.  instead of something like service/task needing to know how to compute this, we simply hide it behind this interface.  

Technically, i could have implemented this witout even passing the ImageProvider to something like TaskDefinition.  *However*, in that case, the ImageProvider would ahve to make the resources without having a clear idea about who should be the parent.  Instead, we allow the parent to be passed in here, allowing the TaskDefinition to pass itself in here.  This gives good parenting.

If we decide this is not important, we can simply make image/environment into properties.

nodejs/aws-infra/experimental/image.ts
```diff
@@ -0,0 +1,206 @@
...
+
+export abstract class ImageProvider implements IImageProvider {
```
CyrusNajmabadi: providers are paired with a concrete factory class that can be used to easily create instances for patterns we want to support.  but users can always supply just an impl of hte interface.  we never depend on downcasting to the concrete types.

nodejs/aws-infra/experimental/image.ts
```diff
@@ -0,0 +1,206 @@
...
+    public static fromDockerBuild(build: docker.DockerBuild): IImageProvider {
+        return new AssetImageProvider(build);
+    }
+
+    /**
+     * Creates a [ContainerImage] given function code to use as the implementation of the container.
+     */
+    public static fromFunction(func: () => void): IImageProvider {
+        return new FunctionImageProvider(func);
+    }
```
CyrusNajmabadi: these are the forms Cloud.Service has traditionally supported.

nodejs/aws-infra/experimental/image.ts
```diff
@@ -0,0 +1,206 @@
...
+        return "lukehoban/nodejsrunner";
+    }
+
+    public environment(name: string, parent: pulumi.Resource): pulumi.Input<aws.ecs.KeyValuePair[]> {
+        const serialized = pulumi.runtime.serializeFunctionAsync(this.func);
+        return serialized.then(value => [{
+            name: "PULUMI_SRC",
+            value: value,
+        }]);
+    }
```
CyrusNajmabadi: things are much cleaner since tehre is no explicit dispatch logic anymore.  Each majorly separate form of image-providing is just in its own type.

nodejs/aws-infra/experimental/image.ts
CyrusNajmabadi: Note: providers should have methods that return 'Inputs'.  this gives impls the most flexibility in terms of computing and returning data.

nodejs/aws-infra/experimental/image.ts
```diff
@@ -0,0 +1,206 @@
...
+}
+
+class AssetImageProvider extends ImageProvider {
```
CyrusNajmabadi: all the logic here is copied from cloud-aws.Service

nodejs/aws-infra/experimental/loadBalancer.ts
```diff
@@ -0,0 +1,254 @@
...
+
+import * as aws from "@pulumi/aws";
+import * as docker from "@pulumi/docker";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as utils from "../utils";
+
+import * as mod from ".";
+
+export type LoadBalancers = aws.ecs.ServiceArgs["loadBalancers"];
```
CyrusNajmabadi: this is a pretty ugly type.  So i just used this form to avoid copying.  Might make sense to do the copy though.

nodejs/aws-infra/experimental/loadBalancer.ts
```diff
@@ -0,0 +1,254 @@
...
+
+export interface ILoadBalancerProvider {
+    portMappings(containerName: string, name: string, parent: pulumi.Resource): pulumi.Input<aws.ecs.PortMapping[]>;
+    loadBalancers(containerName: string, name: string, parent: pulumi.Resource): LoadBalancers;
```
CyrusNajmabadi: as above, Task/Service doesn't have to know/care about how these are implemented.  they simply return the data appropriate for them to add to their specific configs.

nodejs/aws-infra/experimental/loadBalancer.ts
CyrusNajmabadi: probably should be named ServiceLoadBalancers

nodejs/aws-infra/experimental/loadBalancer.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    // endpoints(): pulumi.Output<aws.apigateway.x.Endpoint[]>;
+    // defaultEndpoint(): pulumi.Output<aws.apigateway.x.Endpoint>;
+}
+
+export interface PortInfo {
```
CyrusNajmabadi: this is the equivalent aws-infra form of cloud.ContainerPort.  This enables easy porting down to this level.  However, *importantly*  you are not *required* to use this.  We simply provide a lot of heavy lifting if you want to use this form.  however, you can provider your own impl of ILoadBalancerProvider if you want.  And if there are other patterns that are super common, we can provide those as well.  This decoupling allows us to be opinionated, and to provide a lot of value, without locking someone into a particular approach to this sort of thing.

nodejs/aws-infra/experimental/loadBalancer.ts
```diff
@@ -0,0 +1,254 @@
...
+            portInfo: PortInfo,
+            args: aws.elasticloadbalancingv2.LoadBalancerArgs = {}) {
+        return new PortInfoLoadBalancerProvider(portInfo, args);
+    }
+}
+
+export class PortInfoLoadBalancerProvider extends LoadBalancerProvider {
+    portMappings: (containerName: string, name: string, parent: pulumi.Resource) => pulumi.Input<aws.ecs.PortMapping[]>;
+    loadBalancers: (containerName: string, name: string, parent: pulumi.Resource) => LoadBalancers;
+    defaultEndpoint: () => pulumi.Output<aws.apigateway.x.Endpoint>;
```
CyrusNajmabadi: interesting augmentation.  We don't require all ILoadBalancerProviders to expose this (i honestly don't know if they can).  However, this specific impl is able to expose this.  This keeps Task/Service from having any awareness of 'endpoitns'.  but it means if code uses 'fromPortInfo' they get back a provider that easily exposes to them this Endpoint value that they can use if they want for things like hooking up APIGateway proxies.

nodejs/aws-infra/experimental/taskDefinition.ts
```diff
@@ -0,0 +1,390 @@
...
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as utils from "../utils";
+
+import * as mod from ".";
+
+export declare type HostOperatingSystem = "linux" | "windows";
+
+export interface TaskRunOptions {
```
CyrusNajmabadi: this should probably include all the values that aws.ecs.runTask allows.

nodejs/aws-infra/experimental/taskDefinition.ts
```diff
@@ -0,0 +1,390 @@
...
+    /**
+     * The cluster to run this task in.
+     */
+    cluster: mod.Cluster;
```
CyrusNajmabadi: this allows Tasks to be cluster independent.  A task could be run using an EC2 or Fargate service.

nodejs/aws-infra/experimental/taskDefinition.ts
```diff
@@ -0,0 +1,390 @@
...
+}
+
+const defaultComputePolicies = [
+    aws.iam.AWSLambdaFullAccess,                 // Provides wide access to "serverless" services (Dynamo, S3, etc.)
+    aws.iam.AmazonEC2ContainerServiceFullAccess, // Required for lambda compute to be able to run Tasks
+];
+
+export function defaultTaskDefinitionTaskRolePolicies() {
+    return defaultComputePolicies.slice();
+}
```
CyrusNajmabadi: a way of exporting this stuff.  Should probably be a static method on TaskDefinition for better grouping.  Note that we always return copies so no one can muck with the defaults.

nodejs/tslint.json
```diff
@@ -32,7 +32,6 @@
         "member-access": false,
         "member-ordering": [
             true,
-            "static-before-instance",
```
CyrusNajmabadi: i hate this rule.

nodejs/tslint.json
```diff
@@ -57,7 +56,6 @@
         "no-internal-module": true,
         "no-parameter-properties": false,
         "no-require-imports": false,
-        "no-shadowed-variable": true,
```
CyrusNajmabadi: we've disabled this in other repos already.  it's really obnoxious with outputs/apply where it is actually super nice to just be able to reuse the same name for the output and the unwrapped value.

CyrusNajmabadi:
@lukehoban   Can you take a look?  

I highly recommend first skimming things and looking at the comments i wrote explaining design decisions/thoughts.  I'd like to hear your thoughts on things.

once we have some agreement on things, i'll make changes, and also do a writeup on the overall approach here.  Then, we can pull in others (like @joeduffy ) to go over this.  I don't want to do the writeup first though if we're going to change/tweak  a bunch of it :)   Thanks!

nodejs/aws-infra/experimental/loadBalancer.ts
```diff
@@ -0,0 +1,234 @@
...
+                certificateArn: certificateArn,
+                port: portInfo.port,
+                defaultAction: {
+                    type: "forward",
+                    targetGroupArn: targetGroup.arn,
+                },
+                // If SSL is used, we automatically insert the recommended ELB security policy from
+                // http://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html.
+                sslPolicy: certificateArn ? "ELBSecurityPolicy-2016-08" : undefined,
+            }, parentOpts);
```
CyrusNajmabadi: @lukehoban Open questions from talking to @kennyjwilli. Basically, this approach can work in the sense that he can use this helpful default system, but fallback to just fully providing a full impl ILoadBalancerProvider himself.  However, tha'ts somewhat unpalatable as that's another large cliff to fall off of (even though it is at least good to have that escape hatch).

So, for our default interface implementations, we should still consider ways we can provide flexibility extensibility.  Perhaps by having these guys have their own optional args to help control things.

nodejs/aws-infra/examples/ec2/index.ts
```diff
@@ -0,0 +1,335 @@
...
+
+import * as pulumi from "@pulumi/pulumi";
+
+import * as aws from "@pulumi/aws";
+import * as awsinfra from "@pulumi/aws-infra";
+
+import { Config, Output } from "@pulumi/pulumi";
+
+const network = awsinfra.Network.getDefault();
+const cluster = new awsinfra.x.Cluster("testing", { network });
```
joeduffy: If I omit the network argument, does it use the default?

nodejs/aws-infra/experimental/ec2.ts
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * All the containers to make a ClusterTaskDefinition from.  Useful when creating a
+     * ClusterService that will contain many containers within.
+     *
+     * Either [container] or [containers] must be provided.
+     */
+    containers?: Record<string, mod.ContainerDefinition>;
+}>;
+
+export class EC2TaskDefinition extends mod.TaskDefinition {
```
joeduffy: Is this standard casing for TypeScript, versus, say, `Ec2TaskDefinition`? (Really just asking.)

nodejs/aws-infra/experimental/cluster.ts
joeduffy: Did you consider `EcsCluster`, so that someone doesn't confuse this with an EKS cluster?

joeduffy: Related to this, would you envision supporting an `EksCluster` in here someday down the road (once it stabilizes a bit)?

nodejs/aws-infra/experimental/cluster.ts
joeduffy: :+1:

nodejs/aws-infra/examples/ec2/index.ts
joeduffy: I see later that, yes, it does.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,141 @@
...
+        return [
+            // Expose SSH
+            {
+                fromPort: 22,
+                toPort: 22,
+                protocol: "TCP",
+                cidrBlocks: [ "0.0.0.0/0" ],
+            },
+            // Expose ephemeral container ports to Internet.
+            // TODO: Limit to load balancer(s).
```
joeduffy: Should we file a work item so we don't forget to do this? If it's not in the initial version, fixing this later would be a breaking change. It's also something we probably want to consider exposing arguments to control.

nodejs/aws-infra/experimental/cluster.ts
joeduffy: Although I like avoiding `cluster.cluster` (which I agree is ugly), calling it `instance` does have the downside of potentially misleading people into thinking this returns the EC2 instance(s) underlying the cluster.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,141 @@
...
+        ];
+    }
+}
+
+(<any>Cluster).doNotCapture = true;
+
+// The shape we want for ClusterArgs.  We don't export this as 'Overwrite' types are not pleasant to
+// work with. However, they internally allow us to succinctly express the shape we're trying to
+// provide. Code later on will ensure these types are compatible.
+type OverwriteShape = utils.Overwrite<aws.ecs.ClusterArgs, {
```
joeduffy: I didn't follow most of this ... Does this have the effect of adding all properties on `aws.ecs.ClusterArgs` to the `ClusterArgs` type below or something? I notice we pass `args` above to the `aws.ecs.Cluster` constructor which led me to wonder if that means we expose the full superset of all arguments available on the ECS args?

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,572 @@
...
+        const instanceProfile = getInstanceProfile(name, args, parentOpts);
+        const fileSystem = args.fileSystem;
+
+        const securityGroups =
+            pulumi.output(args.securityGroups).apply(g => g || cluster.instanceSecurityGroups.map(g => g.id));
+        const instance = new aws.ec2.LaunchConfiguration(name, {
+            ...args,
+            securityGroups,
+            imageId: getEcsAmiId(args.ecsOptimizedAMIName),
+            instanceType: pulumi.output(args.instanceType).apply(t => t || "t2.micro"),
```
joeduffy: Why not rely on the default assigned by the provider, if an explicit one is not specified?

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    public static defaultInstanceProfilePolicyARNs() {
+        return [aws.iam.AmazonEC2ContainerServiceforEC2Role, aws.iam.AmazonEC2ReadOnlyAccess];
+    }
+}
+
+(<any>ClusterAutoScalingLaunchConfiguration).doNotCapture = true;
+
+const defaultRootBlockDevice = {
```
joeduffy: Any reason we need to define these, rather than just relying on the provider's default?

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,572 @@
...
+        }, parentOpts);
+
+        this.cluster = cluster;
+        this.launchConfiguration = launchConfiguration;
+    }
+}
+
+(<any>ClusterAutoScalingGroup).doNotCapture = true;
+
+// TODO[pulumi/pulumi-aws/issues#43]: We'd prefer not to use CloudFormation, but it's the best way to implement
```
joeduffy: Is this still required?

nodejs/aws-infra/experimental/fargate.ts
joeduffy: Seems a little strange to use a codename like "Fargate" in the class definitions.

nodejs/aws-infra/experimental/ec2.ts
CyrusNajmabadi: It's incredibly over the map, primarily because TS just wraps JS, and JS does wahtever it wants per library.

nodejs/aws-infra/experimental/cluster.ts
CyrusNajmabadi: we're discussing right now the best patterns we canhave for sensible defaults.  Right now this PR demosntrates an approach whereby:

1. we (pulumi) can have a default in case users do not want to supply values.
2. users can opt out entirely from our default and supply their own.

What is missing is the best way to allow users to *interact* with our defaults.  i.e. how does a user say: "i like your default, but i want to tweak this one part".  What this approach provide is a way for them to at least get at the data so they don't need to necessarily repeat it on their own side.

1. We need to decide if waht we have here is sufficient.
2. We need to make sure what what we have here does'nt block future work to make things more flexible if we decide it isn't.  

This is on the list for @lukehoban and i to discuss as part of the overall design/patterns here for this layer.

nodejs/aws-infra/experimental/cluster.ts
CyrusNajmabadi: First, this is an internal detail for bookkeeping purposes.  The idea here is to provide a way (internally) to simple express the shape of data we want to expose.  IN this example, we are saying "the args to the Pulumi-Cluster shuld be equivlent to the args for the ecs-Cluster, except with the following changes made".  It allows us to simply see how we're augmenting the shape of things.

However, the 'Overwrite' type is less pleasant to use from a 'docs'/'intellisense'/go-to-def/'errors/diagnostics' perspective.  So we don't expose that type for users to use themselves.  Instead, we hand-write the type they will be using.  We then ensure that these types are comptible by doing a no-op assignment each type to the other.  in this way, we will discover if changes happen to one that are not reflected in the other.

CyrusNajmabadi: Second:

> I notice we pass args above to the aws.ecs.Cluster constructor which led me to wonder if that means we expose the full superset of all arguments available on the ECS args?

Right.  the intent here is to ensure we do not lock users out from being able to pass any args properties along to the underlying ecs.Cluster that is instantiated.  This is an 'escape hatch' if you will, that means that if we miss something, or are behind, users can still be unblocked and can pass along the data they need.

nodejs/aws-infra/experimental/fargate.ts
CyrusNajmabadi: I actually think this is pretty reasonable.  This is the TaskDefintion you use when you know you're targetting fargate, and thus you don't need to supply a bunch of dummy defaults we can take care of for you.

CyrusNajmabadi: Do you have an alternate approach you think would be nicer?

nodejs/aws-infra/examples/ec2/index.ts
CyrusNajmabadi: yup.  this was more to allow the test to serve more as an explicit example of how you might do things.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
CyrusNajmabadi: We don't have to.  But then our awsinfra.Cluster story is a lot weaker than our cloud.Service story.  specifically, if you wanted to move down from cloud to awsinfra, you'd have to figure out all these defaults and add them in.  A high level goal of awsinfra that @lukehoban and i are exploring is:

> Can we provide sensible starting defaults that allow users to not have to configure so much stuff, while also being flexible enough to ensure we're not boxing users out from having control when tehy need it.

Note; if we did not provide these defaults, then honestly there's not that much value this PR is providing.  After all, the user could just manually make the launch-config themselves, as well as the cloud-formation-stack, as well as the cluster.  But that's not a very pleasant story (especially when you're moving downwards from 'cloud').

nodejs/aws-infra/experimental/clusterAutoScaling.ts
CyrusNajmabadi: I honestly don't know.  @lukehoban ?

Makefile
```diff
@@ -11,5 +11,5 @@ publish_packages:
 .PHONY: travis_cron travis_push travis_pull_request travis_api
 travis_cron: all
 travis_push: only_build only_test publish_packages
-travis_pull_request: all
+travis_pull_request: only_build only_test_fast
```
CyrusNajmabadi: this is bringing this repo in line with our other ones.  Specifically, it is enabling us to have cron 'full tests' and ci/cd 'short tests'

nodejs/aws-infra/aws_test.go
```diff
@@ -15,18 +15,30 @@
...
+	"github.com/pulumi/pulumi/pkg/resource"
+	"github.com/pulumi/pulumi/pkg/resource/config"
+	"github.com/pulumi/pulumi/pkg/resource/stack"
 	"github.com/pulumi/pulumi/pkg/testing/integration"
+	"github.com/pulumi/pulumi/pkg/util/contract"
 )
 
-// Fargate is only supported in `us-east-1`, so force Fargate-based tests to run there.
-const fargateRegion = "us-east-1"
+const fargateRegion = "us-west-1"
```
CyrusNajmabadi: have been running into limit problems in other regions.  So i'm using us-west-1 for now as it was having the least amount of problems.

nodejs/aws-infra/experimental/cluster.ts
CyrusNajmabadi: We could def do EcsCluster.  That's probably a good idea, unless we want this guy itself to live in an 'ecs' module.  will discuss with @lukehoban

nodejs/aws-infra/experimental/cluster.ts
CyrusNajmabadi: I'm hoping that the 'type' of the variable dissuades people from that idea.  do you have an alternate idea here?

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,594 @@
...
+                DIR_SRC=$AWS_AVAILABILITY_ZONE.$EFS_FILE_SYSTEM_ID.efs.$AWS_REGION.amazonaws.com
+                DIR_TGT=${mountPath}
+                # Update /etc/fstab with the new NFS mount
+                cp -p /etc/fstab /etc/fstab.back-$(date +%F)
+                echo -e \"$DIR_SRC:/ $DIR_TGT nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2 0 0\" | tee -a /etc/fstab
+                mount -a -t nfs4
+                # Restart Docker
+                docker ps
+                service docker stop
+                service docker start
```
CyrusNajmabadi: open questions: should any of this be configurable?

nodejs/aws-infra/experimental/clusterAutoScaling.ts
CyrusNajmabadi: Note; my feeling is leaning toward us not doing this.  The downside though is parenting/args.  With the pattern above if you make ClusterAutoScalingGroup it can automatically pass its cluster to the ClusterAutoScalingLaunchConfiguration.  It will also parent the launch config with itself. 

If we require the user to manually instantiate this child, then they have to repetitively pass that cluster along and have to figure out a reasonable parent.  Not sure if this makes it worth it to do things this way.

nodejs/aws-infra/experimental/clusterFileSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+            ingress: [
+                // Allow NFS traffic from the instance security group
+                {
+                    securityGroups: cluster.instanceSecurityGroups.map(g => g.id),
+                    protocol: "TCP",
+                    fromPort: 2049,
+                    toPort: 2049,
+                },
+            ],
+            tags: { Name: efsSecurityGroupName },
```
CyrusNajmabadi: not sure why we need this tag here since hte name here is the same as the name of the security group we make.

nodejs/aws-infra/experimental/clusterService.ts
```diff
@@ -0,0 +1,324 @@
...
+     * Wait for the service to reach a steady state (like [`aws ecs wait
+     * services-stable`](https://docs.aws.amazon.com/cli/latest/reference/ecs/wait/services-stable.html))
+     * before continuing. Defaults to `true`.
+     */
+    waitForSteadyState?: pulumi.Input<boolean>;
+
+    /**
+     * Optional auto-scaling group for the cluster.
+     */
+    autoScalingGroup?: mod.ClusterAutoScalingGroup;
```
CyrusNajmabadi: the autoscaling group is really only needed so that teh Service can take a 'dependson' relationship purely so it can wait for this to be ready.  it seems unfortunate.  One way we could get around this would be to have a ClusterAutoScalingGroup register itself with the Cluster.  Then, the Service could just take a dependsOn on that implicitly without people needing to know to pass this in.  Thoughts @lukehoban ?

nodejs/aws-infra/experimental/containerDefinition.ts
CyrusNajmabadi: note: we could make 'image' into a union type of `string | IImageProvider`, but it seemed a little hinky to me.

nodejs/aws-infra/experimental/containerDefinition.ts
```diff
@@ -0,0 +1,151 @@
...
+            //     containerPath: v.containerPath,
+            //     sourceVolume: (v.sourceVolume as Volume).getVolumeName(),
+            // })),
+            logConfiguration: container.logConfiguration || {
+                logDriver: "awslogs",
+                options: {
+                    "awslogs-group": logGroupId,
+                    "awslogs-region": aws.config.requireRegion(),
+                    "awslogs-stream-prefix": containerName,
+                },
```
CyrusNajmabadi: note: if the user provide a logConfiguration, then we don't need to generate a logGroup for the TaskDefinition.  We could consider making the loggroup lazily created.

nodejs/aws-infra/experimental/cluster.ts
CyrusNajmabadi: resolution: try to propose a module structure.

nodejs/aws-infra/experimental/cluster.ts
CyrusNajmabadi: todo: brainstorm naming on this.  ideas: "raw" "underlyingInstance".

nodejs/aws-infra/experimental/cluster.ts
CyrusNajmabadi: Decision: open issues for all these decisions.  We will want to ensure on a resource-by-resource basis that these values are what we want.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,141 @@
...
+                fromPort: 0,
+                toPort: 65535,
+                protocol: "TCP",
+                cidrBlocks: [ "0.0.0.0/0" ],
+            },
+        ];
+    }
+}
+
+(<any>Cluster).doNotCapture = true;
```
CyrusNajmabadi: try to remove the doNotCapture.  We shouldn't need that.

nodejs/aws-infra/experimental/cluster.ts
```diff
@@ -0,0 +1,141 @@
...
+     * created.
+     */
+    instanceSecurityGroups?: aws.ec2.SecurityGroup[];
+}
+
+// Make sure our exported args shape is compatible with the overwrite shape we're trying to provide.
+let overwriteShape: OverwriteShape = undefined!;
+let argsShape: ClusterArgs = undefined!;
+argsShape = overwriteShape;
+overwriteShape = argsShape;
```
CyrusNajmabadi: can we do this more compactly.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,579 @@
...
+
+        // Use the instance provided, or create a new one.
+        const instanceProfile = args.instanceProfile ||
+            ClusterAutoScalingLaunchConfiguration.createInstanceProfile(
+                name, /*assumeRolePolicy:*/ undefined, /*policyArns:*/ undefined, parentOpts);
+
+        const fileSystem = args.fileSystem;
+
+        const securityGroups =
+            pulumi.output(args.securityGroups).apply(g => g || cluster.instanceSecurityGroups.map(g => g.id));
```
CyrusNajmabadi: invert this.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        this.instance = new aws.cloudformation.Stack(name, {
+            ...args,
+            name: launchConfiguration.stackName,
+            templateBody: getCloudFormationTemplate(
+                name,
+                launchConfiguration.instance.id,
+                cluster.network.subnetIds,
+                args.templateParameters || {}),
```
CyrusNajmabadi: find terraform's autoscaling group args bag and use that.  aws.autoscaling.group.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,579 @@
...
+    /**
+     * The config to use when creating the auto scaling group.
+     *
+     * [launchConfiguration] or [launchConfigurationArgs] can be provided.  And, if either are
+     * provided will be used as the launch configuration for the auto scaling group.
+     *
+     * If neither are provided, a default instance will be create by calling
+     * [cluster.createAutoScalingConfig()].
+     */
+    launchConfiguration?: ClusterAutoScalingLaunchConfiguration;
```
CyrusNajmabadi: merge launchconfig args into autoscaling args.

nodejs/aws-infra/experimental/clusterFileSystem.ts
```diff
@@ -0,0 +1,183 @@
...
+// limitations under the License.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as mod from ".";
+
+import * as utils from "../utils";
+
+export class ClusterFileSystem extends pulumi.ComponentResource {
```
CyrusNajmabadi: drop this.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,579 @@
...
+                 .apply(([clusterId, cloudFormationStackName, fileSystemId, mountPath]) => {
+        let fileSystemRuncmdBlock = "";
+
+        if (fileSystemId) {
+            // This string must be indented exactly as much as the block of commands it's inserted
+            // into below!
+            mountPath = mountPath || "/mnt/efs";
+
+            // tslint:disable max-line-length
+            fileSystemRuncmdBlock = `
```
CyrusNajmabadi: find way to injet this instead of needing to pass in the file system.

nodejs/aws-infra/experimental/clusterAutoScaling.ts
```diff
@@ -0,0 +1,579 @@
...
+     * will be mounted at '/dev/xvdcz'.  Both devices will be deleted upon termination.
+     */
+    ebsBlockDevices?: aws.ec2.LaunchConfigurationArgs["ebsBlockDevices"];
+}
+
+// Make sure our exported args shape is compatible with the overwrite shape we're trying to provide.
+let overwriteShape: OverwriteShape = undefined!;
+let argsShape: ClusterAutoScalingLaunchConfigurationArgs = undefined!;
+argsShape = overwriteShape;
+overwriteShape = argsShape;
```
CyrusNajmabadi: need to write blog post about the story of these layers.

CyrusNajmabadi: need to move apigateway.x.over here.

nodejs/aws-infra/experimental/clusterService.ts
```diff
@@ -0,0 +1,307 @@
...
+                opts: pulumi.ResourceOptions = {}) {
+        super(type, name, args, opts);
+
+        // If the cluster has an autoscaling group, ensure the service depends on it being created.
+        // TODO(cyrusn): this isn't necessary if resource creation automatically makes 'deps' for
+        // the opts passed in. Investigate.
+        const dependsOn: pulumi.Resource[] = [];
+        if (args.autoScalingGroup) {
+            dependsOn.push(args.autoScalingGroup);
+        }
```
CyrusNajmabadi: fix this up.

nodejs/aws-infra/experimental/clusterService.ts
```diff
@@ -0,0 +1,307 @@
...
+    }];
+}
+
+function getLoadBalancers(service: mod.ClusterService, name: string, args: ClusterServiceArgs) {
+    let loadBalancers: mod.LoadBalancers;
+    for (const containerName of Object.keys(args.taskDefinition.containers)) {
+        const container = args.taskDefinition.containers[containerName];
+        if (container.loadBalancerProvider) {
+            loadBalancers = utils.combineArrays(
+                loadBalancers, container.loadBalancerProvider.loadBalancers(containerName, name, service));
```
CyrusNajmabadi: don't call this provider.

nodejs/aws-infra/experimental/cluster.ts
joeduffy: I had thought of `raw` also.

Just to check, are we applying this to _all_ resources? E.g., no more `bucket.bucket`?

nodejs/aws-infra/experimental/image.ts
```diff
@@ -0,0 +1,204 @@
...
+
+import * as aws from "@pulumi/aws";
+import * as docker from "@pulumi/docker";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as utils from "../utils";
+
+export type ImageEnvironment = Record<string, pulumi.Input<string>>;
+
+export interface IImageProvider {
```
joeduffy: Why the `I` prefix?

CyrusNajmabadi: .net strikes again!  i will remove :)

PR #63, opened 2018-12-06
Introduce a component abstraction for security groups.

CyrusNajmabadi:
This was pulled out from the "Load Balancer" work.  In that PR i ran into where an ApplicationListener wants to configure the ingress/egress rules for the security groups of its load balancer.  Having a first class way to do this makes it simple to add those rules after the fact, instead of needing to know them up front when making the ALB.


CyrusNajmabadi:
lightly easier to review as https://github.com/pulumi/pulumi-aws-infra/pull/63/files?w=1

PR #67, opened 2018-12-11
Allow caller to pass in a network config.

CyrusNajmabadi:


joeduffy:
Sweet, THANK YOU for the quick fix here! 🎉 👍

PR #69, opened 2018-12-11
Add in new abstractions for easily creating Application and Network load balancers.

CyrusNajmabadi:
This PR adds in more convenient helpers for creating:

1. LoadBalancers
2. Listeners
3. Target groups

All of the above are available for both Applications and Networks.  i.e. ALBs and NLBs.  The expected way to use these apis is like so:

```ts
const lb = network.createNetworkLoadBalancer("nlb");

// make a listener.  a targetgroup will automatically be created in this case if not specified.
const listener = lb.createListener("http", { port: 80 });

new x.ecs.EC2Service("custom", {
    cluster,
    taskDefinitionArgs: {
        containers: {
            webserver: {
                memory: 128,
                portMappings: listener,
...
});
```

In this case, the created listener can automatically be used to provide the 'portMappings' for the service being constructed.  

In the case where a target group is desired (for example, for mapping from one incoming port to a port on the container) that can be done like so:

```ts
const lb = network.createNetworkLoadBalancer("nlb");
const tg = network.createTargetGroup("group", { port: 8080 });
const listener = tg.createListener("http", { port: 80 });
...
```


CyrusNajmabadi:
Mrging in now to keep the size of things in check.  Will still need more work and an overall review.

PR #78, opened 2018-12-18
Introduce an abstraction for VPCs

CyrusNajmabadi:
Not ready for review.  Pushing this to get test coverage.

### DIFF
--- a/nodejs/aws-infra/examples/ec2/index.ts
+++ b/nodejs/aws-infra/examples/ec2/index.ts
@@ -21,6 +21,12 @@ const x = awsinfra.x;
 
 import { Config, Output } from "@pulumi/pulumi";
 
+
+const vpc = new awsinfra.x.ec2.Vpc("testvpc", {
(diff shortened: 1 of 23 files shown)


CyrusNajmabadi:
Merging in now.  Still have cleanup to do.

lukehoban:
One general question - we do use the current `Network` abstraction fairly heavily, and I'm not 100% clear on whether this new abstraction is a full replacement for that or not.  Is it possible to maintain compatibility with its core API so that this can be a drop-in replacement?  Or at least we should look at the use of this in EKS cluster and in examples and ensure that we are covering all the use cases well.  The `runTask` use case appears to have changed in a way that I think is not correct (not checking for privateSubnets when deciding whether to assign a public IP), which I think is because the interface to this new abstraction doesn't provide access to that information - which I expect will be critical for several use cases.

CyrusNajmabadi:
It's intended to be a drop-in-replacements.  but you're def right that i broke something.  i actually just misinterpreted the check you pointed out.  I'm going to make an explicit note about fixing this in the general tracking issue for awsinfra.

PR #81, opened 2018-12-19
Expose a more flexibe `runTask` on TaskDefinition

### LINKED ISSUE #80: `FargateTaskDefinition#run` should support full range of `runTask` usage
We still have a runtime method `run` on `FargateTaskDefinition` which has the same limited capabilities as the version that was in `cloud`.  We should ensure that this functionality is exposed in a rich way aligned with the overall goals of `aws-infra`.

I see two options:
1.  Keep a runtime method, probably naming it `runTask` for consistency, and have it's arg bag be derived from the full AWS-SDK arg bag with more inputs optional or not settable (in particular `launchType` and `networkConfiguration`).
2. Instead, have some way to get the `networkConfiguration` details off of a `cluster` so that the underlying `runTask` can be easily called directly.

The current API has several critical limitations:
1. No way to pass many parameters to `run` - like `command`.
2. No way to get the returned `TaskARN` to use for querying status of the task later.

lukehoban:
This removes some of the `cloud` leftovers of the `run` implementation, and instead aligns fully with the awssdk `runTask` function.

We use a similar trick to override the awssdk args interface with a slightly modified one that can infer more inputs from the `cluster` and `task`.

Fixes #80.

### DIFF
--- a/nodejs/aws-infra/examples/ec2/index.ts
+++ b/nodejs/aws-infra/examples/ec2/index.ts
@@ -288,11 +288,10 @@ const api = new aws.apigateway.x.API("examples-containers", {
             policies: [...x.ecs.TaskDefinition.defaultTaskRolePolicyARNs()],
             callback: async (req) => {
                 try {
-                    const c = cluster;
-                    await helloTask.run({ cluster: c });
+                    const result = await helloTask.run({ cluster }).promise();
                     return {
                         statusCode: 200,
-                        body: JSON.stringify({ success: true }),
+                        body: JSON.stringify({ success: true, tasks: result.tasks }),
                     };
                 } catch (err) {
                     return handleError(err);
--- a/nodejs/aws-infra/examples/ec2/index.ts
+++ b/nodejs/aws-infra/examples/ec2/index.ts
@@ -288,7 +288,7 @@ const api = new aws.apigateway.x.API("examples-containers", {
             policies: [...x.ecs.TaskDefinition.defaultTaskRolePolicyARNs()],
             callback: async (req) => {
                 try {
-                    const result = await helloTask.run({ cluster }).promise();
+                    const result = await helloTask.run({ cluster });
                     return {
                         statusCode: 200,
                         body: JSON.stringify({ success: true, tasks: result.tasks }),
--- a/nodejs/aws-infra/experimental/ecs/taskDefinition.ts
+++ b/nodejs/aws-infra/experimental/ecs/taskDefinition.ts
@@ -15,41 +15,12 @@
 import * as aws from "@pulumi/aws";
 import * as pulumi from "@pulumi/pulumi";
 
+import * as awssdk from "aws-sdk";
+
 import * as utils from "../../utils";
 
 import * as ecs from ".";
 import * as x from "..";
-
-export declare type HostOperatingSystem = "linux" | "windows";
-
-export interface TaskRunOptions {
-    /**
-     * The cluster to run this task in.
-     */
-    cluster: ecs.Cluster;
-
-    /**
-     * The name of the container to run as a task.  If not provided, the first container in the list
-     * of containers in the ClusterTaskDefinition will be the one that is run.
-     */
-    containerName?: string;
-
-    /**
-     * The OS to run.  Defaults to 'linux' if unspecified.
-     */
-    os?: HostOperatingSystem;
-
-    /**
-     * Optional environment variables to override those set in the container definition.
-     */
-    environment?: aws.ecs.KeyValuePair[];
-
-    // /**
-    //  * The subnets to run the task in.  Defaults to the public subnets of the cluster's VPC if
-    //  * unspecified.
-    //  */
-    // subnets?: string[];
-}
 
 export abstract class TaskDefinition extends pulumi.ComponentResource {
     public readonly instance: aws.ecs.TaskDefinition;
@@ -58,10 +29,10 @@ export abstract class TaskDefinition extends pulumi.ComponentResource {
     public readonly taskRole: aws.iam.Role;
     public readonly executionRole: aws.iam.Role;
 
-    /**
-     * Runs this task definition in this cluster once.
-     */
-    public readonly run: (options: TaskRunOptions) => Promise<void>;
+    public readonly runTask: (
+        params: RunTaskRequest,
+        callback?: (err: awssdk.AWSError, data: awssdk.ECS.Types.RunTaskResponse) => void,
+    ) => awssdk.Request<awssdk.ECS.Types.RunTaskResponse, awssdk.AWSError>;
 
     constructor(type: string, name: string,
                 isFargate: boolean, args: TaskDefinitionArgs,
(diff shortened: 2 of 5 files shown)


nodejs/aws-infra/experimental/ecs/taskDefinition.ts
```diff
@@ -195,38 +156,32 @@ export abstract class TaskDefinition extends pulumi.ComponentResource {
...
-            throw new Error("No valid container name found to run task for.");
-        }
-
-        // Extract the environment values from the options
-        const env1 = innerContainers[containerName] || [];
-        const env2 = options.environment || [];
-
-        const env = [...env1, ...env2];
-
         const assignPublicIp = isFargate; // && !usePrivateSubnets;
```
lukehoban: This change to remove `usPrivateSubnets` was made in the recent change from Network to Vpc.  I'm not sure this is correct now - Tasks in a public+private network should be running in the private subnet and should not be assigned public IPs.

nodejs/aws-infra/experimental/ecs/taskDefinition.ts
```diff
@@ -195,38 +156,32 @@ export abstract class TaskDefinition extends pulumi.ComponentResource {
...
-        taskDefArn: pulumi.Output<string>,
-        containerToEnvironment: pulumi.Output<Record<string, aws.ecs.KeyValuePair[]>>) {
+export type RunTaskRequest = utils.Overwrite<awssdk.ECS.RunTaskRequest, {
+    /**
+     * The Cluster to run the Task in.
+     */
+    cluster: ecs.Cluster;
+    taskDefinition: never;
+    launchType: never;
+}>;
```
CyrusNajmabadi: instead of this, can you follow my pattern of having this type, and then exporting the fully copied version over?

nodejs/aws-infra/experimental/ecs/taskDefinition.ts
```diff
@@ -195,38 +156,32 @@ export abstract class TaskDefinition extends pulumi.ComponentResource {
...
+
+function createRunFunction(isFargate: boolean, taskDefArn: pulumi.Output<string>) {
+    return function runTask(
+                params: RunTaskRequest,
+                callback?: (err: awssdk.AWSError, data: awssdk.ECS.Types.RunTaskResponse) => void,
+            ): awssdk.Request<awssdk.ECS.Types.RunTaskResponse, awssdk.AWSError> {
```
CyrusNajmabadi: aside: i like how you wrapped this to make the signature clear.  it's a constant problem with these super long sigs on how to do it in a readable manner.

nodejs/aws-infra/experimental/ecs/taskDefinition.ts
CyrusNajmabadi: yup.  this def isn't good.  i will address as part of the work to allow vpcplacement's to be provided.

nodejs/aws-infra/experimental/ecs/taskDefinition.ts
lukehoban: I've tried a couple times and can't get that to work - plus the error doesn't provide any clue how to fix.  Any ideas?

```
export interface RunTaskRequest {
    /**
     * The short name or full Amazon Resource Name (ARN) of the cluster on which to run your task. If you do not specify a cluster, the default cluster is assumed.
     */
    cluster: ecs.Cluster;
    /**
     * A list of container overrides in JSON format that specify the name of a container in the specified task definition and the overrides it should receive. You can override the default command for a container (that is specified in the task definition or Docker image) with a command override. You can also override existing environment variables (that are specified in the task definition or Docker image) on a container or add new environment variables to it with an environment override.  A total of 8192 characters are allowed for overrides. This limit includes the JSON formatting characters of the override structure. 
     */
    overrides?: awssdk.ECS.TaskOverride;
    /**
     * The number of instantiations of the specified task to place on your cluster. You can specify up to 10 tasks per call.
     */
    count?: awssdk.ECS.BoxedInteger;
    /**
     * An optional tag specified when a task is started. For example, if you automatically trigger a task to run a batch process job, you could apply a unique identifier for that job to your task with the startedBy parameter. You can then identify which tasks belong to that job by filtering the results of a ListTasks call with the startedBy value. Up to 36 letters (uppercase and lowercase), numbers, hyphens, and underscores are allowed. If a task is started by an Amazon ECS service, then the startedBy parameter contains the deployment ID of the service that starts it.
     */
    startedBy?: awssdk.ECS.String;
    /**
     * The name of the task group to associate with the task. The default value is the family name of the task definition (for example, family:my-family-name).
     */
    group?: awssdk.ECS.String;
    /**
     * An array of placement constraint objects to use for the task. You can specify up to 10 constraints per task (including constraints in the task definition and those specified at runtime).
     */
    placementConstraints?: awssdk.ECS.PlacementConstraints;
    /**
     * The placement strategy objects to use for the task. You can specify a maximum of five strategy rules per task.
     */
    placementStrategy?: awssdk.ECS.PlacementStrategies;
    /**
     * The platform version the task should run. A platform version is only specified for tasks using the Fargate launch type. If one is not specified, the LATEST platform version is used by default. For more information, see AWS Fargate Platform Versions in the Amazon Elastic Container Service Developer Guide.
     */
    platformVersion?: String;
    /**
     * The network configuration for the task. This parameter is required for task definitions that use the awsvpc network mode to receive their own elastic network interface, and it is not supported for other network modes. For more information, see Task Networking in the Amazon Elastic Container Service Developer Guide.
     */
    networkConfiguration?: awssdk.ECS.NetworkConfiguration;
    /**
     * The metadata that you apply to the task to help you categorize and organize them. Each tag consists of a key and an optional value, both of which you define. Tag keys can have a maximum character length of 128 characters, and tag values can have a maximum length of 256 characters.
     */
    tags?: awssdk.ECS.Tags;
    /**
     * Specifies whether to enable Amazon ECS managed tags for the task. For more information, see Tagging Your Amazon ECS Resources in the Amazon Elastic Container Service Developer Guide.
     */
    enableECSManagedTags?: awssdk.ECS.Boolean;
    /**
     * Specifies whether to propagate the tags from the task definition or the service to the task. If no value is specified, the tags are not propagated.
     */
    propagateTags?: awssdk.ECS.PropagateTags;
}

type RunTaskRequestOverrideShape = utils.Overwrite<awssdk.ECS.RunTaskRequest, {
    /**
     * The Cluster to run the Task in.
     */
    cluster: ecs.Cluster;
    taskDefinition?: never;
    launchType?: never;
}>;
const test: string = utils.checkCompat<RunTaskRequestOverrideShape, RunTaskRequest>();
```

lukehoban: Ahh - indeed - fixed.

nodejs/aws-infra/examples/ec2/index.ts
```diff
@@ -288,11 +288,10 @@ const api = new aws.apigateway.x.API("examples-containers", {
             policies: [...x.ecs.TaskDefinition.defaultTaskRolePolicyARNs()],
             callback: async (req) => {
                 try {
-                    const c = cluster;
-                    await helloTask.run({ cluster: c });
+                    const result = await helloTask.run({ cluster }).promise();
```
lukehoban: @CyrusNajmabadi Curious what you think about this `.promise()`?  I tried to keep this faithful to the API design of `aws-sdk`, which I think is the right call, but it does mean a very specific and unusual API design which may not be obvious given this is exposed apart from the `aws-sdk` itself.

CyrusNajmabadi: I'm pretty biased and i think that async/await has 'won the war'.  So my preference would be to just have the promisified form by default.  I don't think anyone is going to really be harmed by the *pulumi* layer being opinionated in that fashion.  

But i also don't feel super strongly about this.  i get the attractiveness of having the signatures be very close.  But it just feels like Promises are the better solution all the time.  

1. you can use async/await if you want.
2. if callback-style is preferable for you, you can always manually `.then` things. 

`.promise()` always feels like the way to deal with moving an older api forward.  I'd prefer to avoid it on the new apis.

CyrusNajmabadi:
Thanks for the `promise` change. :)

PR #90, opened 2019-01-03
Make it simpler to createa cluster by passing security-group IDs directly.

CyrusNajmabadi:
Followup to https://github.com/pulumi/pulumi-aws-infra/pull/89

### DIFF
--- a/nodejs/aws-infra/experimental/ecs/cluster.ts
+++ b/nodejs/aws-infra/experimental/ecs/cluster.ts
@@ -54,8 +54,7 @@ export class Cluster
         // IDEA: Can we re-use the network's default security group instead of creating a specific
         // new security group in the Cluster layer?  This may allow us to share a single Security Group
         // across both instance and Lambda compute.
-        this.securityGroups = args.securityGroups ||
-            [Cluster.createDefaultSecurityGroup(name, this.vpc, parentOpts)];
+        this.securityGroups = getSecurityGroups(this, name, args, parentOpts);
 
         this.extraBootcmdLines = () => instance.id.apply(clusterId =>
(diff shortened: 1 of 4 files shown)


nodejs/aws-infra/experimental/ecs/cluster.ts
```diff
@@ -160,10 +179,16 @@ export interface ClusterArgs {
...
+     * The security group to place new instances into.  If neither [securityGroups] nor
+     *  [securityGroupIds] are provided, a default will be created for the Cluster.
      */
     securityGroups?: x.ec2.SecurityGroup[];
+
+    /**
+     * The security group ids to place new instances into.  If neither [securityGroups] nor
+     *  [securityGroupIds] are provided, a default will be created for the Cluster.
+     */
+    securityGroupIds?: pulumi.Input<string>[];
```
lukehoban: Any reason not to just have one field and allow it to be `securityGroups?: Array<x.ec2.SecurityGroup | pulumi.Input<string>>;`?  That would make the API a little simpler, and would even seem to make the implementation here a bit simpler.  Lastly it's closer to how our strongly typed properties look on raw AWS APIs when we add `AltType` annotations (although I don't think we have cases of arrays of these alternatives).

CyrusNajmabadi: @lukehoban I'll try that.  i personally have a bias against those sorts of complex signatures as i find them harder to read.  Also, impl-wise it is a bit of a PITA as you have to dynamically check for 'string/promise/output'.  But if you honestly prefer it, i'm ok with that :)  let me create an alternate pr and we can compare!

CyrusNajmabadi:
Going with https://github.com/pulumi/pulumi-aws-infra/pull/92 instead.

PR #92, opened 2019-01-03
Make it simpler to createa cluster by passing security-group IDs directly.

CyrusNajmabadi:
Alternate approach over https://github.com/pulumi/pulumi-aws-infra/pull/90/

### DIFF
--- a/nodejs/aws-infra/experimental/ec2/securityGroup.ts
+++ b/nodejs/aws-infra/experimental/ec2/securityGroup.ts
@@ -25,17 +25,26 @@ export class SecurityGroup extends pulumi.ComponentResource {
     public readonly egressRules: x.ec2.IngressSecurityGroupRule[] = [];
     public readonly ingressRules: x.ec2.IngressSecurityGroupRule[] = [];
 
-    constructor(name: string, args: SecurityGroupArgs, opts?: pulumi.ComponentResourceOptions) {
+    constructor(name: string, args: SecurityGroupArgs = {}, opts: pulumi.ComponentResourceOptions = {}) {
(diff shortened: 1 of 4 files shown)


nodejs/aws-infra/experimental/ecs/cluster.ts
```diff
@@ -130,14 +129,37 @@ export class Cluster
...
+            }
+            else {
+                result.push(x.ec2.SecurityGroup.fromExistingId(`${name}-${i}`, obj, {
+                    vpc: cluster.vpc,
+                }, opts));
+            }
+        }
+
+        return result;
+    }
```
CyrusNajmabadi: I was wrong.  it wasn't actually hard to make this work @lukehoban

nodejs/aws-infra/experimental/ec2/securityGroup.ts
```diff
@@ -37,6 +40,11 @@ export class SecurityGroup extends pulumi.ComponentResource {
         this.registerOutputs();
     }
 
+    /** @internal */
+    public static isSecurityGroupInstance(obj: any): obj is SecurityGroup {
+        return !!(<SecurityGroup>obj).__isSecurityGroupInstance;
+    }
```
lukehoban: It's a shame these are needed.  I would have hoped we could just check for "is Input<string>" more easily - but I can see that probably isn't easy to implement.  The public API here feels simpler though - so I still do generally like this approach.

CyrusNajmabadi: 👍

PR #94, opened 2019-01-04
Use a consistent approach to security groups.

CyrusNajmabadi:
This approach was introduced in https://github.com/pulumi/pulumi-aws-infra/commit/f45138ffc6a8a920aecb6195836c09bcc3864ceb.  However, it was not applied uniformly to all of aws-infra.  This PR does that.

### DIFF
--- a/nodejs/aws-infra/experimental/autoscaling/autoscaling.ts
+++ b/nodejs/aws-infra/experimental/autoscaling/autoscaling.ts
@@ -20,6 +20,7 @@ import * as utils from "./../../utils";
 
 export class AutoScalingLaunchConfiguration extends pulumi.ComponentResource {
     public readonly instance: aws.ec2.LaunchConfiguration;
+    public readonly securityGroups: x.ec2.SecurityGroup[];
 
     /**
      * Optional file system to mount.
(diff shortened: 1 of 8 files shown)


nodejs/aws-infra/experimental/ec2/securityGroup.ts
```diff
@@ -79,6 +79,32 @@ export class SecurityGroup extends pulumi.ComponentResource {
     }
 }
 
+export type SecurityGroupOrId = SecurityGroup | pulumi.Input<string>;
```
CyrusNajmabadi: this seems like a generally reasonable pattern for resources we want to allow passing either as rich objects, or as IDs to be rehydrated.  VPCs, subnets, etc. would likely all benefit from a similar approach.

CyrusNajmabadi: The name `XxxOrId` also seems better than ```XxxLike```

nodejs/aws-infra/experimental/autoscaling/autoscaling.ts
```diff
@@ -20,6 +20,8 @@ import * as utils from "./../../utils";
 
 export class AutoScalingLaunchConfiguration extends pulumi.ComponentResource {
     public readonly launchConfiguration: aws.ec2.LaunchConfiguration;
+    public readonly id: pulumi.Output<string>;
```
CyrusNajmabadi: it's common enough to go to the underlying resource just to get the ID.  So to make that simpler, i've moved to a pattern of exposing the 'id' directly on the rich wrapper as well.

CyrusNajmabadi:
Prremptively merging in.

PR #101, opened 2019-01-08
Allow a default cluster to be created on behalf of the user.

CyrusNajmabadi:


CyrusNajmabadi:
@lukehoban I'm curious what you think about this.  Unlike with a 'vpc', we're not 'discovering' an existing 'default' thing created by Amazon.  Instead, we're just creating an optional ecs.Cluster for the user on their behalf to allow them to avoid having to create and pass this in themselves.

Upsides are that this can simplify and shorten the amount of code you'd need to write in an actual pulumi app.

Downside is that by not being explicit, you might forget to pass along an actual Cluster to something (say a Service), and then you'd get this default picked up.

Overall, i think it's a net positive.  However, i could see us going with the approach where you need to be more explicit about this sort of thing.  Thoughts?

nodejs/aws-infra/examples/ec2/index.ts
```diff
@@ -24,7 +24,7 @@ const autoScalingGroup = cluster.createAutoScalingGroup("testing", {
         minSize: 20,
     },
     launchConfigurationArgs: {
-        instanceType: "t2.medium",
+        instanceType: "t2.xlarge",
```
CyrusNajmabadi: will undo this.

CyrusNajmabadi:
Going to go ahead with this.  If you don't like this idea, let me know and i'll roll back.

PR #112, opened 2019-01-23
Fix issue that was preventing creating a vpc at a cidr address above a certain point.

CyrusNajmabadi:


nodejs/aws-infra/ec2/cidr.ts
```diff
@@ -126,15 +126,15 @@ function getIPv4AddressValue(address: string): number {
...
         if (val < 0 || val >= 256) {
-            throw new Error(`Each part of the ip address must be between 0 and 255. Got "${val}"`);
+            throw new Error(`Each part of the ip address must be between 0 and 255. Parsed "${val}" from "${piece} in ${address}"`);
         }
 
         return val;
     }
 }
 
 export function getIPv4Address(value: number): string {
```
swgillespie: Not sure if you've got a Node test harness set up, but it would be awesome if we could turn this into a normal Node unit test 😄

CyrusNajmabadi: Yeah.  we could def make these into unit tests as the cidr object is just pure js.  let me file issue to track that.

PR #114, opened 2019-01-23
Remove docker caching from awsinfra layer.

CyrusNajmabadi:
We really do not want to have caching in our infrastructure layers.  It makes ownership of resources much more complex, and produces all sorts of other problems (i.e. what to name things, when to reuse them).

It's cleaner to just produce resources on demand, clearly parented by an obvious root, and allow users to create caching themselves at their own layer (i.e. by making their own singleton object they could pass around as needed).


CyrusNajmabadi:
Merging in.  let me know if you want any changes @lukehoban

PR #120, opened 2019-01-24
expose public routetable id

houqp:
this PR adds back the public attribute added in #10

### DIFF
--- a/nodejs/aws-infra/network.ts
+++ b/nodejs/aws-infra/network.ts
@@ -112,6 +112,10 @@ export class Network extends pulumi.ComponentResource implements ClusterNetworkA
     public readonly publicSubnetIds: pulumi.Output<string>[];
 
     /**
+     * The public subnet route table for the VPC.
+     */
+    public readonly publicRouteTableId: pulumi.Output<string>;
+    /**
      * Gets the default VPC for the AWS account as a Network.  This first time this is called, the
      * default network will be lazily created, using whatever options are provided in opts. All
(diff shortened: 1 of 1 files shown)


houqp:
I have a quick question regarding exposing children resources as public attributes. What's the reason for choosing `publicRouteTable.id` instead of just `publicRouteTable`? Wouldn't the latter result in a more flexible API?

CyrusNajmabadi:
> I have a quick question regarding exposing children resources as public attributes. What's the reason for choosing publicRouteTable.id instead of just publicRouteTable? Wouldn't the latter result in a more flexible API?

We don't have a hard/fast rule here.  You're definitely correct that this would make things more flexible.  However, it would also lock us into a fixed implementation.  Consider, for example, if we wanted to change things so that this information could be provided by configuration, and no actual resource was created.  Just exposing the id gives us that flexibility.

That said, i would not have a problem if you actually wanted to expose the route table directly.  So feel free to make that change if that would help you out :)

CyrusNajmabadi:
I'll wait for your decision on if you want to change things.  If you're ok as-is, i'll happily merge.  Otherwise, if you'd prefer to just expose the table, then feel free to make the change, and then ping me once things pass.  thanks!

houqp:
Thanks for the explanation, I will keep it as is for now since the original PR also exposes it as id string. For our own use-case, an id is good enough :)

CyrusNajmabadi:
Thanks much!

PR #129, opened 2019-01-28
WIP: IAM User support

hausdorff:


nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
+import * as aws from "@pulumi/aws";
```
CyrusNajmabadi: copyright.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
...
+        const groupMembership = args.groupMembership;
+        const accessKey = args.accessKey;
+
+        // Explicitly delete these props so we do *not* pass them into the User created
+        // below.
+        delete args.groupMembership;
+        delete args.accessKey;
+        delete args.user;
+
+        this.user = args.user || new aws.iam.User(name, args && args, { parent: this });
```
CyrusNajmabadi: `args && args`?

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
...
+        args: UserArgs = {},
+        opts: pulumi.ComponentResourceOptions = {}
+    ) {
+        return new User(name, { ...args, user: aws.iam.User.get(name, id, {}, opts) }, opts);
+    }
+}
+
+type OverwriteUserArgs = utils.Overwrite<
+    aws.iam.UserArgs,
+    {
```
CyrusNajmabadi: can you follow hte syntax used elsewhere? consistency is preferred here :)

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
...
+     *
+     * @param name The _unique_ name of the resource. This name will be transformed into the form
+     * `employee.${name}`.
+     * @param args The arguments to use to populate this resource's properties.
+     * @param opts A bag of options that control this resource's behavior.
+     */
+    constructor(name: string, args?: UserArgs, opts?: pulumi.CustomResourceOptions) {
+        super(`employee.${name}`, args, opts);
+    }
+}
```
CyrusNajmabadi: is this for testing purposes?

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
...
+        super(`contractor.${name}`, args, opts);
+    }
+}
+
+/**
+ * BotUser represents an AWS IAM User that is a bot, such as a CI system. Because it is useful to
+ * quickly distinguish between various types of users, all BotUsers are allocated names of the form
+ * `bot.${name}`.
+ */
+export class BotUser extends User {
```
CyrusNajmabadi: i'm not sure i get a lot of value from these.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
...
+     * created.
+     */
+    user?: aws.iam.User;
+}
+
+// Make sure our exported args shape is compatible with the overwrite shape we're trying to provide.
+const test1: string = utils.checkCompat<OverwriteUserArgs, UserArgs>();
+
+type OverwriteUserGroupMembershipArgs = utils.Overwrite<
+    aws.iam.UserGroupMembershipArgs,
```
CyrusNajmabadi: same complaint.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        if (groupMembership !== undefined) {
+            this.groupMembership = new aws.iam.UserGroupMembership(
+                name,
+                { ...groupMembership, user: this.user.name },
+                { parent: this }
+            );
+        }
+
+        if (accessKey == "create") {
```
CyrusNajmabadi: intead of this string we can define things as such:

1. if args.accessKey is undefined, do nothing.
2. if it eixsts, we create an accesskey from whatever it is.

I'm not sure i see the need for an actual 'create'.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,213 @@
...
+                { parent: this }
+            );
+        } else if (args && accessKey !== undefined && accessKey != "create") {
+            this.accessKey = new aws.iam.AccessKey(
+                name,
+                { ...accessKey, user: this.user.name },
+                { parent: this }
+            );
+        }
+    }
```
CyrusNajmabadi: i'm on the fence about hte accesskey/group stuff being in the constructor.  perhaps it would be nicer to model as instance methods like:

`addToGroup(...)` and `createAccessKey(...)`?

nodejs/aws-infra/iam/index.ts
hausdorff: I think this makes things much less clear, actually, because without `user`, none of the fields are required. So passing `accessKey: {}` would result in a creation of an access key, which seems super weird.

That said, I think it's fine to pull this out of the constructor, and put it into a method instead.

nodejs/aws-infra/iam/index.ts
hausdorff: As mentioned offline:

1. this is is a broadly-accepted best practice ([_cf._](https://github.com/gruntwork-io/module-security-public/tree/master/modules/iam-policies#iam-users)), which
2. we will be pushing up into the "prod" packages, and pulling it back into x when we're sure it's the right thing to do.

nodejs/aws-infra/iam/index.ts
hausdorff: I believe:

* The vast majority of human `User`s will be added to groups, and therefore I think it makes sense to include that in the constructor.
* It will also be useful to allow people to add group memberships after instantiation, so an `addToGroups` function makes sense to me.

`AccessKey` is not something most users will want, so I'm ok moving that to a method.

Happy to hear alternative views.

hausdorff: A question I do have is whether `AccessKey` and `UserGroupMemberships` should be stored inside the `User` when using these methods. I see that `SecurityGroup` does not keep these resources, and instead returns them. I think that is confusing, but I'm guessing we had good reasons for doing that.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,195 @@
...
+     * @param args The properties of the resulting `AccessKey` resource.
+     * @param opts A bag of options that control the resulting `AccessKey` resource's behavior.
+     * __NOTE__: the `parent` field is overridden and set to `this`, rather than the value (if any)
+     * in `opts`.
+     */
+    public createAccessKey(
+        name: string,
+        args?: AccessKeyArgs,
+        opts?: pulumi.CustomResourceOptions,
+    ): aws.iam.AccessKey {
```
CyrusNajmabadi: nit: we don't wrap the final paren in the style normally.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,195 @@
...
+        const key = new aws.iam.AccessKey(
+            name,
+            { ...args, user: this.user.name },
+            { ...opts, parent: this },
+        );
```
CyrusNajmabadi: same comment about paren.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public createAccessKey(
+        name: string,
+        args?: AccessKeyArgs,
+        opts?: pulumi.CustomResourceOptions,
+    ): aws.iam.AccessKey {
+        const key = new aws.iam.AccessKey(
+            name,
+            { ...args, user: this.user.name },
+            { ...opts, parent: this },
```
CyrusNajmabadi: our pattern elsewhere has been to do:  `opts || { parent: this }`   But yours is arguably more correct.  most correct would likely be: `{ parent: this, ...opts}`.  i.e. we try to fill in useful default values, but the caller always has the final say.

That said, can you use `opts || { parent: this }` for now (to be consistent with the rest of aws-infra).  I'll make a change to update these all to be better.

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,195 @@
...
+     * Add the current `User` to an arbitrary set of IAM Groups, returning the resulting
+     * `UserGroupMembership` resources.
+     *
+     * @param name The _unique_ name of the resulting `UserGroupMembership` resource.
+     * @param args The properties of the resulting `UserGroupMembership` resource.
+     * @param opts A bag of options that control the resulting `UserGroupMembership` resource's
+     * behavior. __NOTE__: the `parent` field is overridden and set to `this`, rather than the value
+     * (if any) in `opts`.
+     */
+    public addToGroups(
```
CyrusNajmabadi: if we have this, do we need to support passing in group-membership in the constructor?

nodejs/aws-infra/iam/index.ts
```diff
@@ -0,0 +1,195 @@
...
+
+type OverwriteAccessKeyArgs = utils.Overwrite<aws.iam.AccessKeyArgs, { user?: never }>;
+
+/**
+ * The set of arguments for constructing a AccessKey resource.
+ */
+export interface AccessKeyArgs {
+    /**
+     * Either a base-64 encoded PGP public key, or a
+     * keybase username in the form `keybase:some_person_that_exists`.
```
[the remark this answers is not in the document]
CyrusNajmabadi: also, instead of this, should we consider having two exclusive properties, and just passing them along accordingly.  this is our abstraction, so we don't necessarliy have to overload concepts like this.

leezen:
Closing old PR

PR #138, opened 2019-02-03
Don't manually provide an AutoScalingGroup name.  

CyrusNajmabadi:
This can cause problems with updates because it can make stacks conflict across updates.

### DIFF
--- a/nodejs/awsx/aws_test.go
+++ b/nodejs/awsx/aws_test.go
@@ -70,23 +70,23 @@ func Test_Examples(t *testing.T) {
 				"@pulumi/awsx",
 			},
 		},
-		{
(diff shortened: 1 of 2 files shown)


nodejs/awsx/aws_test.go
```diff
@@ -70,23 +70,23 @@ func Test_Examples(t *testing.T) {
...
-				"@pulumi/awsx",
-			},
-			Quick:       true,
-			SkipRefresh: true,
-			PreviewCommandlineFlags: []string{
-				"--diff",
-			},
-			ExtraRuntimeValidation: containersRuntimeValidator(region, true /*isFargate*/),
-		},
+		// {
```
CyrusNajmabadi: temporary, so we can have travis just do the ec2 tests.

Issue #139, opened 2019-02-03
Expose subnets and NAT creation on Vpc


kennyjwilli:
The `Vpc` object makes working with newly created VPCs easy -- you get to specify the number of public/private/isolated subnets and automatically have NAT and/or Internet gateways automatically created. My use case requires us to work within an **existing** VPC so we do not have access to all those niceties provided by the `Vpc` object -- they are only available when creating a **new** `Vpc`.

My suggestion is to expose the subnet, NAT gateway, and Internet gateway creation as methods on the `Vpc` object. For example: 

```typescript
const vpc = infra.ec2.Vpc.fromExistingIds("my-vpc", {
    vpcId: myExistingVpcId,
    publicSubnetIds: mySubnets,
});

vpc.addSubnet()
vpc.subnets[0].withNatGateway()
vpc.addNatGatewayToPrivateSubnets
vpc.addInternetGateway
```

CyrusNajmabadi:
Great idea!

CyrusNajmabadi:
Note: `vpc.addSubnet()` would be harder to do as we compute the cidr breakdown up front.  All the rest would likely be things we could do hopefully simply.

PR #148, opened 2019-02-09
Make it possible to easily add an Internet Gateway to a vpc

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/139

### DIFF
--- a/nodejs/awsx/ec2/natGateway.ts
+++ b/nodejs/awsx/ec2/natGateway.ts
@@ -29,7 +29,7 @@ export class NatGateway
     constructor(name: string, vpc: x.ec2.Vpc, args: NatGatewayArgs, opts?: pulumi.ComponentResourceOptions);
     constructor(name: string, vpc: x.ec2.Vpc, args: ExistingNatGatewayArgs, opts?: pulumi.ComponentResourceOptions);
     constructor(name: string, vpc: x.ec2.Vpc, args: NatGatewayArgs | ExistingNatGatewayArgs, opts?: pulumi.ComponentResourceOptions) {
-        super("awsx.x.ec2.NatGateway", name, {}, opts || { parent: vpc });
+        super("awsx:x:ec2:NatGateway", name, {}, opts || { parent: vpc });
 
         const parentOpts = { parent: this };
 
--- a/nodejs/awsx/ec2/natGateway.ts
+++ b/nodejs/awsx/ec2/natGateway.ts
@@ -38,10 +38,6 @@ export class NatGateway
             this.natGateway = args.natGateway;
         }
         else {
-            if (!args.subnet) {
-                throw new Error("[args.subnet] must be provided.");
-            }
-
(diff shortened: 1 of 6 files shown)


CyrusNajmabadi:
tagging Sean/Luke to see what you guys think here. This fits into our goal of making it easy to stand up resources with reasonable defaults.  For example, if you have a vpc and you want to give it an IG that routes to all public subnets, that can be done easily now. It wasn't particular hard before.  But it couldbe annoying (especially determining what values to set up in hte routes for the IG).

nodejs/awsx/ec2/natGateway.ts
```diff
@@ -0,0 +1,96 @@
...
+        implements x.ec2.SubnetRouteProvider {
+
+    public readonly vpc: x.ec2.Vpc;
+    public readonly elasticIP: aws.ec2.Eip;
+    public readonly natGateway: aws.ec2.NatGateway;
+
+    constructor(name: string, vpc: x.ec2.Vpc, args: NatGatewayArgs, opts?: pulumi.ComponentResourceOptions);
+    constructor(name: string, vpc: x.ec2.Vpc, args: ExistingNatGatewayArgs, opts?: pulumi.ComponentResourceOptions);
+    constructor(name: string, vpc: x.ec2.Vpc, args: NatGatewayArgs | ExistingNatGatewayArgs, opts?: pulumi.ComponentResourceOptions) {
+        super("awsx.x.ec2.NatGateway", name, {}, opts || { parent: vpc });
```
swgillespie: Should these be colons?

nodejs/awsx/ec2/vpc.ts
```diff
@@ -145,32 +187,22 @@ export class Vpc extends pulumi.ComponentResource {
...
+            });
+        }
 
-function createInternetGateway(vpc: Vpc, name: string) {
-    if (vpc.publicSubnets.length === 0) {
-        return undefined;
-    }
+        if (idArgs.natGatewayIds) {
+            for (const natGatewayId of idArgs.natGatewayIds) {
+                vpc.natGateways.push(new x.ec2.NatGateway(name, vpc, {
```
swgillespie: I think you need to suffix `name` here with something unique - multiple nat gateways will result in duplicate `NatGateway` instances with the same name.

swgillespie: I think `name` needs a disambiguate suffix here - supplying two nat gateway IDs looks like it would create two NatGateway resources with the same URN.

nodejs/awsx/ec2/natGateway.ts
swgillespie: should these be colons?

CyrusNajmabadi: yes :)

nodejs/awsx/ec2/vpc.ts
CyrusNajmabadi: good point!

CyrusNajmabadi: definitely!

PR #150, opened 2019-02-11
Map our Api component more closely to how AWS handles integrations

CyrusNajmabadi:


nodejs/awsx/apigateway/api.ts
```diff
@@ -118,17 +118,51 @@ function isStaticRoute(route: Route): route is StaticRoute {
...
+     *
+     * mock: for integrating the API method request with API Gateway as a "loop-back" endpoint
+     * without invoking any backend.
+     */
+    type: pulumi.Input<IntegrationType>;
+
+    /**
+     * Specifies the integration's HTTP method type.  Currently, the only supported type is 'ANY'.
+     */
+    httpMethod?: "ANY";
```
CyrusNajmabadi: note: this is a limitation to how we currently set up swagger paths (and compute sha hashes and whatnot).  we can relax this in the future, but it will require a bit more effort.

CyrusNajmabadi:
Improvements that bring us more in line with how AWS manages integrations.

nodejs/awsx/apigateway/api.ts
```diff
@@ -118,17 +118,51 @@ function isStaticRoute(route: Route): route is StaticRoute {
 }
 
 /**
- * An apigateway route that maps to some target uri, or some elastic-load-balancer host/port.
+ * An apigateway route for an integration.
+ * https://docs.aws.amazon.com/apigateway/api-reference/resource/integration/ for more details.
  */
-export interface ProxyRoute {
+export interface IntegrationRoute {
```
CyrusNajmabadi: note: this is technically a breaking change. however, awsx-apigateway is so new i think it can be acceptable for us to do this.  Alternativly, we could keep the 'proxy' name, though 'integration' really is better given what this represents.

What do you think @lukehoban

CyrusNajmabadi:
@lukehoban Can you take a look here when you get a chance?

CyrusNajmabadi:
@lukehoban Can you take a look here when you get a chance?

nodejs/awsx/elasticloadbalancingv2/network.ts
```diff
@@ -133,16 +133,16 @@ export class NetworkListener
...
         // create a VpcLink to the load balancer in the VPC
         const vpcLink = new aws.apigateway.VpcLink(name, {
             targetArn: this.loadBalancer.loadBalancer.arn,
         }, { parent });
 
         return this.endpoint.apply(ep => ({
             uri: `http://${ep.hostname}:${ep.port}/`,
-            type: <x.apigateway.ApigatewayIntegrationType>"http_proxy",
-            connectionType: <x.apigateway.ApigatewayConnectionType>"VPC_LINK",
+            type: <x.apigateway.IntegrationType>"http_proxy",
```
swgillespie: Are these casts necessary? Seems to me that `IntegrationType` is a union that includes `"http_proxy"`.

CyrusNajmabadi: unfortunately yes.   the problem here is that TS infers "string" as the type, not the explicit string-type 'http_proxy'.  This is a known issue with TS that they're considerign improving on for the future.

PR #176, opened 2019-03-09
Add support for custom providers passed in ComponentResourceOptions

igorshapiro:


igorshapiro:
@lukehoban That's what I thought, but then I've seen that all the resources are created using the default provider (even though an explicit provider was passed).
Maybe I missed something, but try passing the component a custom provider...

nodejs/awsx/ec2/natGateway.ts
```diff
@@ -31,7 +31,7 @@ export class NatGateway
     constructor(name: string, vpc: x.ec2.Vpc, args: NatGatewayArgs | ExistingNatGatewayArgs, opts: pulumi.ComponentResourceOptions = {}) {
         super("awsx:x:ec2:NatGateway", name, {}, { parent: vpc, ...opts });
 
-        const parentOpts = { parent: this };
+        const parentOpts = { parent: vpc };
```
CyrusNajmabadi: this doesn't seem right.  we'd want the child resources to see this NatGateway as their parent.  is there a reason this needed ot happen?

nodejs/awsx/ec2/vpc.ts
```diff
@@ -71,12 +71,12 @@ export class Vpc extends pulumi.ComponentResource {
                 enableDnsHostnames: utils.ifUndefined(args.enableDnsHostnames, true),
                 enableDnsSupport: utils.ifUndefined(args.enableDnsSupport, true),
                 instanceTenancy: utils.ifUndefined(args.instanceTenancy, "default"),
-            });
+            }, { parent: this });
```
CyrusNajmabadi: this is good.

nodejs/awsx/ec2/vpc.ts
```diff
@@ -71,12 +71,12 @@ export class Vpc extends pulumi.ComponentResource {
...
             this.id = this.vpc.id;
 
             // Create the appropriate subnets.  Default to a single public and private subnet for each
             // availability zone if none were specified.
-            const topology = new VpcTopology(this, name, cidrBlock, numberOfAvailabilityZones, opts);
+            const topology = new VpcTopology(this, name, cidrBlock, numberOfAvailabilityZones, { parent: this, ...opts });
```
CyrusNajmabadi: `...opts` should be necessary.  we should just need to pass along us as the parent.

nodejs/awsx/ec2/natGateway.ts
igorshapiro: Fixed

nodejs/awsx/ec2/vpc.ts
igorshapiro: Fixed. Only `opts` left

CyrusNajmabadi:
Closing out as the primary parts of this PR have been merged in.  Thanks for the contribution @igorshapiro !

PR #177, opened 2019-03-13
Fix building of application security group description.

CyrusNajmabadi:


lukehoban:
@swgillespie Open an issue to track?  Definitely seems like a good idea which could help catch this kind of thing early.

PR #189, opened 2019-03-25
Define a reasonable parent for component resources.

CyrusNajmabadi:


CyrusNajmabadi:
Note: this should have little impact in that most callers just pass nothing for 'opts' so you get the same results as bfore.

One thing that does change though is that NatGateways are now properly parents by their owning VPC (if they were created as part of creating the VPC).  THis wasn't happening before because one codepath was causing a default `{}` `opts` to be passed in, and that meant we didn't set a parent at all.  Now, we always try to set a parent, but we allow that to be overridden by the caller if they want a different parent.

PR #193, opened 2019-03-26
Flow v0.18.0 changes downstream

jen20:


nodejs/awsx/package.json
```diff
@@ -12,7 +12,7 @@
     "repository": "https://github.com/pulumi/pulumi-awsx",
     "dependencies": {
         "@pulumi/pulumi": "^0.17.1",
-        "@pulumi/aws": "^0.17.0",
+        "@pulumi/aws": "0.17.2-dev.1553126336",
```
CyrusNajmabadi: is this appropriate?  why can't we just reference the latest aws?  or will that come once you publishthe next aws version?

jen20: This (for now) was to reference a known version, but I'll change it to a proper version constraint once we start releasing these.

PR #195, opened 2019-03-28
Add a README.md for apigateway & Request Parameter Validation

CyrusNajmabadi:


nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
+# Pulumi API Gateway Components
+
+Pulumi's API for simplifying working with [API Gateway](https://aws.amazon.com/api-gateway/). The API currently provides ways to define a route that accepts and forwards traffic to a specified destination.
+
```
CyrusNajmabadi: maybe "define routes that accept".  "route"-singular here makes it sound like you can only create one route.

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+## Defining an Endpoint
+
+To define an endpoint you will either need to specify the routes or a swagger string. You can also define the stage name (else it will default to "stage"). A `stage` is an addressable instance of the Rest API.
+
```
CyrusNajmabadi: i would not even cover the swagger part TBH.  It's not really an expected way we think people will use.  I would also lead with the event handler stuff.

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+### Routes
+
+The destination is determined by the route, which can be an Event Handler Route, a Static Route, an Integration Route or a Raw Data Route.
+
```
CyrusNajmabadi: consider using links here to take the user to those sections.

nodejs/awsx/apigateway/README.md
CyrusNajmabadi: It might make sense to explain what a route (or routing) is.  i.e. that it's goign to be an accesssible http address (or URI) that supports some http method (i.e. GET/POST/etc.)  to access it, and which will respond in some specialized way when the http address is hit.

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+#### Event Handler Route
+
+An Event Handler Route is a route that will map to a [Lambda](https://aws.amazon.com/lambda/). You will need to specify the path, method and the lambda. Pulumi allows you to define the lambda inline with your Pulumi code & provisions that appropriate permissions.
+
```
CyrusNajmabadi: 1. "with your app(lication) code"
2. "and" is generally preferred over "&"

CyrusNajmabadi: "provisions that appropriate permissions." 

Maybe "provisions the appropriate permissions on your behalf so that APIGateway can communicate with the Lambda."

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+            return {
+                statusCode: 200,
+                body: "hello",
+            };
+        },
+    }],
+})
+```
+
+A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blob/master/nodejs/awsx/examples/api/index.ts).
```
CyrusNajmabadi: Consider mentioning:
1. that you could link to an existing lambda, and show how you could use `aws.lambda.Function.get` to get that and pass it in here.
2. that you can control the lambda that is created by calling `new aws.lambda.CallbackFunction`

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Static Route
+
+A Static Route is a route that will map to static content in files/directories. You will need to define the local path & then the files (and subdirectories) will be uploaded into S3 objects. If the local path points to a file, you can specify the content-type. Else, the content types for all files in a directory are inferred.
+
+By default, any request on directory will serve index.html. This behavior can be disabled, by setting the index field to false.
```
CyrusNajmabadi: `disabled, by`.  No need for comma.

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+import * as awsx from "@pulumi/awsx";
+
+let endpoint = new awsx.apigateway.API("example", {
+    routes: [{
+        path: "/www",
+        localPath: "www",
+        index: false,
+        requestValidator: "PARAMS_ONLY",
+    }],
+    requestValidator: "ALL",
```
CyrusNajmabadi: i'm not sure i understand the example since you have the requestValidator both at the route level and the API level.  Does this mean that the route option overrides the API option?  Or vice versa?

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+        path: "/www",
+        localPath: "www",
+        index: false,
+        requestValidator: "PARAMS_ONLY",
+        requiredParams: [{
+            name: "key",
+            in: "query",
+        }]
+    }],
+})
```
CyrusNajmabadi: It might be good to indicate an example of a request that this would pass for, and one where it would fail.

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+#### Request Body
+
+[TODO]
+
+### Swagger String
+
+You can use a OpenAPI specification that is in string form to initialize the API Gateway. This in a way is an escape hatch for implementing featured not yet supported by Pulumi. You must manually provide permission for any route targets to be invoked by API Gateway when using this option.
+
```
CyrusNajmabadi: this part is fine.  And i like that it would be at the end.  I wouldn't even mention anything about swagger earlier on.

nodejs/awsx/apigateway/api.ts
```diff
@@ -82,6 +83,18 @@ export type EventHandlerRoute = {
...
+
+    /**
+     * Required Parameters to validate. If the request validator is set to ALL or PARAMS_ONLY, api
+     * gateway will validate these before sending traffic to the event handler.
+     */
+    requiredParameters?: reqvalidation.Parameter[];
+
+    /**
+    * Request Validator specifies the validator to use at the method leve. This will override anything
+    * defined at the API level.
```
CyrusNajmabadi: "leve"

nodejs/awsx/apigateway/api.ts
```diff
@@ -437,8 +481,13 @@ interface ApigatewayIntegration {
     credentials?: pulumi.Output<string>;
 }
 
-function createSwaggerSpec(api: API, name: string, routes: Route[]) {
+function addValidators(swagger: SwaggerSpec, requestValidator: reqvalidation.RequestValidator) {
+    swagger["x-amazon-apigateway-request-validators"] = reqvalidation.allValidators;
+    swagger["x-amazon-apigateway-request-validator"] = requestValidator;
+}
```
CyrusNajmabadi: feel free to inline this (unless it's used in several locations).

nodejs/awsx/apigateway/api.ts
```diff
@@ -376,6 +410,8 @@ interface SwaggerSpec {
     paths: { [path: string]: { [method: string]: SwaggerOperation; }; };
     "x-amazon-apigateway-binary-media-types"?: string[];
     "x-amazon-apigateway-gateway-responses": Record<string, SwaggerGatewayResponse>;
+    "x-amazon-apigateway-request-validators"?: reqvalidation.Validators;
+    "x-amazon-apigateway-request-validator"?: reqvalidation.RequestValidator;
```
CyrusNajmabadi: what's the diff between these two?

nodejs/awsx/apigateway/api.ts
```diff
@@ -544,6 +604,22 @@ function addEventHandlerRouteToSwaggerSpec(
     }
 }
 
+function addRequiredParametersToSwaggerOperation(swaggerOperation: SwaggerOperation, requiredParameters: reqvalidation.Parameter[]) {
+    requiredParameters.forEach(requiredParam => {
```
CyrusNajmabadi: we generally don't use forEach in our codebase.  just `for (const requiredParam of requiredParameters)` is sufficient.

nodejs/awsx/apigateway/api.ts
```diff
@@ -544,6 +604,22 @@ function addEventHandlerRouteToSwaggerSpec(
...
+        let param = {
```
CyrusNajmabadi: const.

nodejs/awsx/apigateway/api.ts
```diff
@@ -593,13 +669,20 @@ function addStaticRouteToSwaggerSpec(
         }, parentOpts);
     }
 
-    function processFile(route: StaticRoute) {
+    function processFile(route: StaticRoute, requiredParams?: reqvalidation.Parameter[]) {
```
CyrusNajmabadi: in general, for internal APIs, it's good to not make the params optional (since people might forget them).  instead, do this: `requiredParam: Parameter[] | undefined`.  this makes hte parameter something the caller has to pass, but it's acceptable for them to pass an undefined value.

nodejs/awsx/apigateway/api.ts
```diff
@@ -662,13 +752,20 @@ function addStaticRouteToSwaggerSpec(
         // Take whatever path the client wants to host this folder at, and add the
         // greedy matching predicate to the end.
         const proxyPath = directoryServerPath + "{proxy+}";
-        addSwaggerOperation(swagger, proxyPath, swaggerMethod("ANY"), createSwaggerOperationForObjectKey(directoryKey, role, "proxy"));
+        let swaggerOperation = createSwaggerOperationForObjectKey(directoryKey, role, "proxy");
```
CyrusNajmabadi: const.  (i'm actually surpirsed the linter isn't complaining).

nodejs/awsx/apigateway/api.ts
```diff
@@ -649,8 +732,15 @@ function addStaticRouteToSwaggerSpec(
                     if (childPath === indexPath) {
                         // We hit the file that we also want to serve as the index file. Create
                         // a specific swagger path from the server root path to it.
+                        let swaggerOperation = createSwaggerOperationForObjectKey(childUrn, role);
+                        if (directory.requiredParameters) {
+                            addRequiredParametersToSwaggerOperation(swaggerOperation, directory.requiredParameters);
```
CyrusNajmabadi: it might be worth documenting that for a static route, the required parameters will apply to all of them. 

Note: it's also not clear to me what it means for a static route to even have validation.  It's not like the query variables would ever be involved in anything right?

nodejs/awsx/apigateway/requestValidator.ts
```diff
@@ -0,0 +1,45 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// These APIs are currently experimental and may change.
+
+export type RequestValidator = "ALL" | "PARAMS_ONLY" | "BODY_ONLY";
+
+export interface Validators {
+    [name: string]: {
```
CyrusNajmabadi: what the key (i.e. 'name') here supposed to represent?

nodejs/awsx/apigateway/requestValidator.ts
```diff
@@ -0,0 +1,45 @@
...
+        validateRequestBody: boolean;
+        validateRequestParameters: boolean;
+    };
+}
+
+export interface Parameter {
+    name: string;
+    in: Location;
+}
```
CyrusNajmabadi: these coudl all use documentation.

nodejs/awsx/apigateway/requestValidator.ts
```diff
@@ -0,0 +1,45 @@
...
+export type Location = "path" | "query" | "header";
+
+export const allValidators = {
```
CyrusNajmabadi: this isn't something a user would ever use right?  Consider inlining into the usage point so it doesn't become part of our public API.

nodejs/awsx/apigateway/requestValidator.ts
CyrusNajmabadi: for example, if i say `name: "foo", in: "path"`  does that mean a substring check that the path contains "foo"?  Same with "header". Does it mean the header just contains that string, or that there's a specific key in the header with that exact name.

CyrusNajmabadi:
Looking great.  Very minor concerns!

nodejs/awsx/apigateway/requestValidator.ts
ekrengel: You give the validators any names you want and then define the validator equal to one of those names. Since this isnt something users need to be bothered w/ I have moved it inline w the SwaggerSpec definition & changed name to validatorName for clarity.

nodejs/awsx/apigateway/api.ts
ekrengel: Will do 

and yeah -- I was a bit unsure if I should to include them or not for static routes for this reason.

nodejs/awsx/apigateway/api.ts
ekrengel: It actually is complaining -- and I'm confused about how this works since we edit it on like 756. Would love to chat w/ you about this.

nodejs/awsx/apigateway/api.ts
ekrengel: Looks like that was left over / I did not it anymore. But thats a great tip thanks.

nodejs/awsx/apigateway/api.ts
ekrengel: For my understanding: Is this a style thing? Or is there a particular reason?

nodejs/awsx/apigateway/api.ts
ekrengel: You have to define the validators (i.e. validate all things, just the body, just the params) & name them. Then "x-amazon-apigateway-request-validator" sets the validator to used at the API level. You can also set a different validator at each method level which will override the API level one.

nodejs/awsx/apigateway/README.md
ekrengel: Route validators override the API validator. Maybe I can add comments to the examples.
`{
        path: "/www",
        localPath: "www",
        requiredParameters: [{
            name: "key",
            in: "query",
        }],
        /**
         * These parameters will not actually be required since the method level validator is
         * set to only validate the body.
         */
        requestValidator: "BODY_ONLY", // This will override the API level requestValidator.
    }`

nodejs/awsx/apigateway/api.ts
CyrusNajmabadi: Totally a style thing.  Since it doesn't seem to add much value, I'd prefer we just keep the code consistent with these sorts of patterns.  If you have a strong desire to use it though, go ahead :)

nodejs/awsx/apigateway/README.md
CyrusNajmabadi: Comments would be great :)

CyrusNajmabadi:
Looking good.  Only skimmed things.  Will need to do a full pass once ready.  Having a test that at least validates one scenario would be good :)

nodejs/awsx/apigateway/README.md
CyrusNajmabadi: nit: maybe "supports a specified HTTP method ('GET', 'POST', etc.)".

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,169 @@
...
+```ts
+import * as aws from "@pulumi/aws";
+import * as awsx from "@pulumi/awsx";
+
+let endpoint = new awsx.apigateway.API("example", {
+    routes: [{
+        path: "/",
+        method: "GET",
+        eventHandler: async (event) => {
+            return {
```
CyrusNajmabadi: consider adding a ``// This code will run in AWS Lambda and will be invoked any time `/` is hit``

nodejs/awsx/apigateway/README.md
```diff
@@ -0,0 +1,169 @@
...
+        }),
+    }],
+})
+```
+
+#### Static Route
+
+A Static Route is a route that will map to static content in files/directories. You will need to define the local path and then the files (and subdirectories) will be uploaded into S3 objects. If the local path points to a file, you can specify the content-type. Else, the content types for all files in a directory are inferred.
+
+By default, any request on directory will serve index.html. This behavior can be disabled by setting the index field to false.
```
CyrusNajmabadi: "by setting the `index` field"

nodejs/awsx/apigateway/api.ts
```diff
@@ -74,6 +87,19 @@ export type StaticRoute = {
...
+
+    /**
+     * Required Parameters to validate. If the request validator is set to ALL or PARAMS_ONLY, api
+     * gateway will validate these before sending traffic to the event handler. Parameter validation
+     * will get applied to all static resources defined by the localPath.
+     */
+    requiredParameters?: reqvalidation.Parameter[];
+
+    /**
+    * Request Validator specifies the validator to use at the method leve. This will override anything
```
CyrusNajmabadi: "leve"

nodejs/awsx/apigateway/api.ts
```diff
@@ -507,6 +581,22 @@ function addEventHandlerRouteToSwaggerSpec(
...
+            name: requiredParam.name,
+            in: requiredParam.in,
+            required: true,
+        };
+
+        if (swaggerOperation["parameters"]) {
+            swaggerOperation["parameters"].push(param);
+        } else {
+            swaggerOperation["parameters"] = [param];
+        }
```
CyrusNajmabadi: another pattern we also do is:

```
swaggerOperation["parameters"] = swaggerOperation["parameters"] || [];
swaggerOperation["parameters"].push(param);
```

CyrusNajmabadi: (that's a nit, if you prefer your form, that's fine).

nodejs/awsx/apigateway/api.ts
```diff
@@ -625,13 +729,20 @@ function addStaticRouteToSwaggerSpec(
...
+        addSwaggerOperation(swagger, proxyPath, swaggerMethod("ANY"), swaggerOperation);
     }
 
     function createSwaggerOperationForObjectKey(
-            objectKey: string,
-            role: aws.iam.Role,
-            pathParameter?: string): SwaggerOperation {
+        objectKey: string,
+        role: aws.iam.Role,
+        pathParameter?: string): SwaggerOperation {
```
CyrusNajmabadi: note: the indentation above was intentional.  the problem with single-indent on parameters is it can make it more difficult to distinguish the parameters from the body statements.  

Not pushing back hard on htis.  if you prefer this form, that's ok.  if so though, please leave the blank line between the parameters and first statements.

nodejs/awsx/apigateway/requestValidator.ts
```diff
@@ -0,0 +1,33 @@
...
+    // name is the specific key of the parameter that is required.
+    name: string;
+
+    /**
+     * in is where the parameter is expected to appear. API Gateway can validate
+     * parameters expected in the path, query or header.
+     */
+    in: Location;
+}
+export type Location = "path" | "query" | "header";
```
CyrusNajmabadi: i think we should either name this as `ParameterLocation`.  Or just remove this entirely and inline this value directly in `Parameter.in: "path" | "query" | "header"` if that's the only place it is used.

CyrusNajmabadi:
LGTM.  Gave minor feedback.  Feel free to change if you want!

nodejs/awsx/apigateway/api.ts
ekrengel: my editor automatically formatted this. i can def change it back

nodejs/awsx/apigateway/api.ts
ekrengel: thats much nicer! thanks for the suggestion

ekrengel: ALSO -- i love nits like these :)

ekrengel:
@CyrusNajmabadi I addressed the remaining comments. How should we go about approving this since you're the one that opened it?

PR #196, opened 2019-03-28
Add files to tsconfig to make sure we build docs for them.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/194

### DIFF
--- a/nodejs/awsx/tsconfig.json
+++ b/nodejs/awsx/tsconfig.json
@@ -20,6 +20,8 @@
         "cluster.ts",
         "network.ts",
         "utils.ts",
+        "apigateway/api.ts",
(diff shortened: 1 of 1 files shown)


lukehoban:
Do you know of any way we could catch this sort of thing earlier?  Or maybe we just need to do something in the doc generator side to avoid having a dependency on the `tsconfig.json` files listing?

CyrusNajmabadi:
> Do you know of any way we could catch this sort of thing earlier?

We could def write some sort of tool for this.  I'll see if one might exist.  

> Or maybe we just need to do something in the doc generator side to avoid having a dependency on the tsconfig.json files listing?

I will look into this as well.

PR #201, opened 2019-04-05
API gateway lambda authorizers

ekrengel:
@CyrusNajmabadi


nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,82 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
 
 API Gateway can perform basic validations against request parameters, a request payload or both. When a validation fails, a 400 error is returned immediately.
 
+### Lambda Authorizers
+
+[Lambda Authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html) are AWS Lambda functions that provide control access to an API. You can define a Lambda Authorizer for an Event Handler Route or a Static Route.
+
+To define an Authorizer, you provide a Lambda that fulfills `aws.lambda.EventHandler<AuthorizerEvent, AuthorizerResponse>` or you provide information on a pre-existing Lambda authorizer. The exmaple below shows defining the Authorizer Lambda directly inline. See the [Event Handler Route](#Event-Handler-Route) section on other ways you can define a Lambda for the Authorizer.
+
```
CyrusNajmabadi: exmaple !

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,82 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+                        policyDocument: {
+                            Version: "2012-10-17",
+                            Statement: [{
+                                Action: "execute-api:Invoke",
+                                Effect: "Allow",
+                                Resource: event.methodArn,
+                            }],
+                        },
+                    };
+                    callback(null, policy);
```
CyrusNajmabadi: can you see if this will work with the 'async' form?

ekrengel: works!

nodejs/awsx/apigateway/README.md
ekrengel: okay -- i was worried about "over exampling" but i can add these :)

ekrengel: lol ohh i have a typ youre not telling me to write more examples haha

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,82 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+                        policyDocument: {
+                            Version: "2012-10-17",
+                            Statement: [{
+                                Action: "execute-api:Invoke",
+                                Effect: "Allow",
+                                Resource: event.methodArn,
+                            }],
+                        },
+                    };
+                    return policy;
```
CyrusNajmabadi: you can even make this simpler by just returning the POJO object.

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,82 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+        authorizers: [{
+            authorizerName: "prettyAuthorizer",
+            parameterName: "auth",
+            parameterLocation: "query",
+            authType: "custom",
+            authorizer: {
+                type: "request",
+                 authorizer: async (event) => {
+                    console.log("Received event:", JSON.stringify(event, null, 2));
+                    const policy: awsx.apigateway.AuthorizerResponse = {
```
CyrusNajmabadi: perhaps add a comment like `// add your own custom authorization logic here.`

nodejs/awsx/apigateway/api.ts
```diff
@@ -58,6 +59,12 @@ export type EventHandlerRoute = {
     * defined at the API level.
     */
     requestValidator?: reqvalidation.RequestValidator;
+
+    /**
+     * Authorizers allows you to define Lambda authorizers be applied for authorization when the
+     * the route is called.
+     */
+    authorizers?: authorizer.SecurityDefinition[];
```
CyrusNajmabadi: should this possibly be called `securityDefinitions?:`?  Or should `SecurityDefinition` possibly be called `Authorizer`?  it's a rather strange naming disparity that the property and the type seem unrelated.

nodejs/awsx/apigateway/api.ts
```diff
@@ -404,6 +419,28 @@ interface SwaggerOperation {
     responses?: { [code: string]: SwaggerResponse };
     "x-amazon-apigateway-integration": ApigatewayIntegration;
     "x-amazon-apigateway-request-validator"?: reqvalidation.RequestValidator;
+
+    /**
+     * security maps to the authorizerName defined in the top level securityDefinitions
+     */
```
CyrusNajmabadi: i'm not sure i understand this comment.  by saying "maps to" it sounds lke the authorizerName will be in the `string[]` value of the map.  But it looks more like the authorizerName is the key.  So i would think it would be "maps from".  (also, i don't know what the `string[]` means here.

nodejs/awsx/apigateway/api.ts
```diff
@@ -581,6 +622,97 @@ function addEventHandlerRouteToSwaggerSpec(
     }
 }
 
+function addAuthorizersToSwagger(swagger: SwaggerSpec, authorizers: authorizer.SecurityDefinition[]): string[] {
+    const authNames: string[] = [];
+    for (const auth of authorizers) {
+        swagger["securityDefinitions"] = swagger["securityDefinitions"] || {};
```
CyrusNajmabadi: looks like this doesn't depend on the loop and can be lifted out.

nodejs/awsx/apigateway/api.ts
```diff
@@ -581,6 +622,97 @@ function addEventHandlerRouteToSwaggerSpec(
...
+function getLambdaAuthorizer(authorizerName: string, lambdaAuthorizer: authorizer.LambdaAuthorizer): LambdaAuthorizer {
+    if (authorizer.isLambdaInfo(lambdaAuthorizer.authorizer)) {
+        return {
+            type: lambdaAuthorizer.type,
+            authorizerUri: lambdaAuthorizer.authorizer.authorizerUri,
+            authorizerCredentials: lambdaAuthorizer.authorizer.authorizerCredentials,
+            identitySource: lambdaAuthorizer.identitySource,
+            identityValidationExpression: lambdaAuthorizer.identityValidationExpression,
+            authorizerResultTtlInSeconds: lambdaAuthorizer.authorizerResultTtlInSeconds,
+        };
```
CyrusNajmabadi: looks like this could be entirely replaced with `return { type: lambdaAuthorizer.type, ...lambdaAuthorizer.authorizer }`

nodejs/awsx/apigateway/api.ts
```diff
@@ -581,6 +622,97 @@ function addEventHandlerRouteToSwaggerSpec(
...
+            throw new Error("Different authorizers with same name " + authName);
+        }
+    }
+    return authNames;
+}
+
+function getLambdaAuthorizer(authorizerName: string, lambdaAuthorizer: authorizer.LambdaAuthorizer): LambdaAuthorizer {
+    if (authorizer.isLambdaInfo(lambdaAuthorizer.authorizer)) {
+        return {
+            type: lambdaAuthorizer.type,
```
CyrusNajmabadi: any reason that 'type' is on lambdaAuthorizer and not on lambdaAuthorizer.authorizer?

nodejs/awsx/apigateway/api.ts
```diff
@@ -581,6 +622,97 @@ function addEventHandlerRouteToSwaggerSpec(
...
+                "Version": "2012-10-17",
+                "Statement": [
+                    {
+                        "Action": "lambda:InvokeFunction",
+                        "Effect": "Allow",
+                        "Resource": "${arn}"
+                    }
+                ]
+            }
+                `),
```
CyrusNajmabadi: consider using pulumi.interpolate here.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,114 @@
...
+import * as pulumi from "@pulumi/pulumi";
+
+import * as awslambda from "aws-lambda";
+
+export type AuthorizerEvent = awslambda.CustomAuthorizerEvent;
+export type AuthorizerResponse = awslambda.CustomAuthorizerResult;
+
+/**
+ * SecurityDefinition provides the definition for a custom Authorizer for API Gateway.
+ */
```
CyrusNajmabadi: why not AuthorizerDefinition or CustomAuthorizerDefinition?

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,114 @@
...
+export interface SecurityDefinition {
+    /**
+     * Pretty name for the security definition to be referenced as. This is only used within the
+     * swagger and must be unique for each unique authorizer within the API. If you want
+     * to share an authorizer across routes, provide the same name for them. If not defined, a unique
+     * name for you.
```
CyrusNajmabadi: "a unique name for you"

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * parameterName is the name of the header or query parameter containing the authorization token. Must be
+     * "Unused" for multiple identity sources or non header or query type of request parameters.
+     * */
+    parameterName: string;
+
+    /**
+     * Required and the value must be "header" or "query" for an API Gateway API. Must be "header" for multiple
+     * identity sources or non header or query type of request parameters.
```
CyrusNajmabadi: i'm finding the sentences here a bit hard to parse.  note: stating that it's required isn't really necessary given that hte property isn't optional.  also, stating it has to eb 'header' or 'query' is shown in the type signature.  so it might be better to explain what those two values mean instead.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,114 @@
...
+    authorizer: LambdaAuthorizer;
+}
+
+/**
+ * Defines a Lambda authorizer (the x-amazon-apigateway-authorizer object) to be applied for authorization of
+ * method invocations in API Gateway.
+ */
+export interface LambdaAuthorizer {
+    /**
+     * The type of the authorizer. This is a required property and the value must be one of the following:
```
CyrusNajmabadi: don't have to mention 'required'.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,114 @@
...
+     *      - "token", for an authorizer with the caller identity embedded in an authorization token
+     *      - "request", for an authorizer with the caller identity contained in request parameters
+     */
+    type: "token" | "request";
+
+    /**
+     * The authorizer specifies information about the authorizing Lambda. You can either set up the Lambda
+     * separately and just provide the required information or you can define the Lambda inline.
```
CyrusNajmabadi: "or you can define the Lambda inline using a JavaScript function".

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,114 @@
...
+     */
+    authorizer: LambdaInfo | aws.lambda.EventHandler<AuthorizerEvent, AuthorizerResponse>;
+
+    /**
+     * Comma-separated list of mapping expressions of the request parameters as the identity source. Applicable
+     * for the authorizer of the "request" type only.
+     * Example: "method.request.header.HeaderAuth1, method.request.querystring.QueryString1"
+     */
+    identitySource?: string;
```
CyrusNajmabadi: consider taking this as a `string[]` and producing the comma-separated string ourself.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,114 @@
...
+    authorizerUri: pulumi.Input<string>;
+
+    /**
+     * Credentials required for invoking the authorizer, if any, in the form of an ARN of an IAM execution role.
+     * For example, "arn:aws:iam::account-id:IAM_role".
+     */
+    authorizerCredentials: pulumi.Input<string>;
+}
+
+export function isLambdaInfo(info: LambdaInfo | aws.lambda.EventHandler<AuthorizerEvent, AuthorizerResponse>): info is LambdaInfo {
```
CyrusNajmabadi: add `/** @internal */` if this is just a helper function for us.

nodejs/awsx/examples/api/authfunction/index.js
```diff
@@ -0,0 +1,44 @@
+// Example modified from https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html
```
CyrusNajmabadi: why 'js' instead of 'ts' here?

nodejs/awsx/examples/api/index.ts
```diff
@@ -67,23 +55,103 @@ const api = new awsx.apigateway.API("myapi", {
...
+            authorizerName: "prettyAuthorizer",
+            parameterName: "auth",
+            parameterLocation: "query",
+            authType: "custom",
+            authorizer: {
+                type: "request",
+                authorizer: async (event) => {
+                    console.log("Received event:", JSON.stringify(event, null, 2));
+                    const policy: awsx.apigateway.AuthorizerResponse = {
+                        principalId: "me",
```
CyrusNajmabadi: is "me" actually something supposed to be used?

nodejs/awsx/examples/api/authfunction/index.js
ekrengel: The example was copied & then modified from an AWS example. I can change it to TS!

nodejs/awsx/apigateway/api.ts
ekrengel: From a user perspective these are Authorizer definitions. But the nitty-gritty is that the swagger integration requires them to be called Security Definitions. I think I will keep public things referring to this as authorizers and then the private things will call it Security Definitions.

nodejs/awsx/apigateway/api.ts
ekrengel: The `lambdaAuthorizer.authorizer` is an object I made to allow your easily supporting Cognito Pools in the future. I think some of the naming here maybe confusing, so I will try to fix that to clarify whats going on.

nodejs/awsx/apigateway/authorizer.ts
ekrengel: I will re-read / re-write these

nodejs/awsx/examples/api/authfunction/index.js
ekrengel: Already talk to you about this: but this is getting packaged up using `pulumi.asset.FileArchive("./authfunction")`

nodejs/awsx/examples/api/index.ts
ekrengel: I'll add a comment here. in theory if you were getting a user from a DB or something that would be passed back as the principalId for the Lambda /backend to then use.

nodejs/awsx/apigateway/api.ts
ekrengel: i will try to clarify here.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
   LoadBalancer if none is provided.
 - Added easier convenience methods overloads on an awsx.ec2.SecurityGroup to make ingress/egress
   rules.
+- Add support for Lambda Authorizers for API Gateway in TypeScript
```
CyrusNajmabadi: consider writing `[Lambda Authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html)`

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+            };
+        },
+        authorizers: [{
+            authorizerName: "prettyAuthorizer",
+            parameterName: "auth",
+            parameterLocation: "query",
+            authType: "custom",
+            type: "request",
+            handler: async (event: awsx.apigateway.AuthorizerEvent) => {
+                console.log("Received event:", JSON.stringify(event, null, 2));
```
CyrusNajmabadi: so, from this example, it doesn't look like we're actually validating anything about the incoming data.  Should we potentially remove this console.log line and add something like:

```ts
// Perform domain-specific validation of the event
```

?

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+                return {
+                    principalId: "user",
+                    policyDocument: {
+                        Version: "2012-10-17",
+                        Statement: [{
+                            Action: "execute-api:Invoke",
+                            Effect: "Allow",
+                            Resource: event.methodArn,
+                        }],
+                    },
```
CyrusNajmabadi: while i don't see a problem with this per-se.  Should we provide simple ways for people to generate common forms of responses here to make their life easier?  i.e. an easy way to generate an "Allow" response for a given method (or a deny response, etc. etc.)?

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
 
 API Gateway can perform basic validations against request parameters, a request payload or both. When a validation fails, a 400 error is returned immediately.
 
+### Lambda Authorizers
+
+[Lambda Authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html) are AWS Lambda functions that provide control access to an API. You can define a Lambda Authorizer for an Event Handler Route or a Static Route. API Gateway supports "request" or "token" type Lambda authorizers.
+
+To define an Authorizer, you provide a Lambda that fulfills `aws.lambda.EventHandler<AuthorizerEvent, AuthorizerResponse>` or you provide information on a pre-existing Lambda authorizer. The example below shows defining the Authorizer Lambda directly inline. See the [Event Handler Route](#Event-Handler-Route) section for other ways you can define a Lambda for the Authorizer.
+
+Below is an example of a custom "request" Lambda Authorizer.
```
CyrusNajmabadi: on the fence if we should give a 1-liner stating the difference between what a "request" authorizer is versus a "token" authorizer.

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+                body: "<h1>Hello world!</h1>",
+            };
+        },
+        authorizers: [{
+            parameterName: "Authorization",
+            parameterLocation: "header",
+            authType: "oauth2",
+            type: "request",
+            handler: async (event: awsx.apigateway.AuthorizerEvent) => {
+                // Add your own custom authorization logic here.
```
CyrusNajmabadi: i like this comment.  we shoudl likely have something similar above.

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+                        Version: "2012-10-17",
+                        Statement: [{
+                            Action: "execute-api:Invoke",
+                            Effect: "Allow",
+                            Resource: event.methodArn,
+                        }],
+                    },
+                };
+            },
+            identitySource: ["method.request.querystring.auth"],
```
CyrusNajmabadi: is this identitySource correct for token authorization?  making sure this isn't a copy/paste mistake.

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+        },
+        authorizers: [{
+            parameterName: "Authorization",
+            parameterLocation: "header",
+            authType: "oauth2",
+            type: "request",
+            handler: async (event: awsx.apigateway.AuthorizerEvent) => {
+                // Add your own custom authorization logic here.
+                // Access the token using event.authorizationToken
+                console.log("Received event:", JSON.stringify(event, null, 2));
```
CyrusNajmabadi: i'd remove the console.log

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+            return {
+                statusCode: 200,
+                body: "<h1>Hello world!</h1>",
+            };
+        },
+        authorizers: [{
+            parameterName: "Authorization",
+            parameterLocation: "header",
+            authType: "oauth2",
+            type: "request",
```
CyrusNajmabadi: similar to above, this is a lot of "goop" to get right.  Perhaps we can make thins easier for the user?  As an example, in awsx we have the concept of `SecurityGroupRuleLocation`s and `SecurityGroupRulePorts`.  These are just raw data, but we also make it easy to get preconfigured instances of those by saying things like `new AnyIPv4Location()` or `new TcpPort(80)`.

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+        ...
+        authorizers: [{
+            authorizerName: "testing",
+            parameterName: "auth",
+            parameterLocation: "query",
+            authType: "custom",
+            type: "request",
+            handler: {
+                uri: pulumi.interpolate`arn:aws:apigateway:${region}:lambda:path/2015-03-31/functions/${authorizerLambda.arn}/invocations`,
+                credentials: gatewayRole.arn,
```
CyrusNajmabadi: this should probably also allow, as an overload, just accepting the authorizer lambda.  We would generate this uri under the covers for the user if they provider that.

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+            },
+            identitySource: ["method.request.querystring.auth"],
```
CyrusNajmabadi: is this correct here?

nodejs/awsx/apigateway/README.md
```diff
@@ -98,6 +98,121 @@ A complete example can be found [here](https://github.com/pulumi/pulumi-awsx/blo
...
+        }],
+    }],
+});
+```
+
+A complete example of defining the Lambda Authorizer elsewhere can be found [here](https://github.com/pulumi/pulumi-awsx/blob/master/nodejs/awsx/examples/api/index.ts).
+
```
CyrusNajmabadi: in general, use a permalink in github so that we always reference something at a known good point in time.  that way, for example, if we remove the code from this example and move it elsewhere, the permalink is still valid.

nodejs/awsx/apigateway/api.ts
CyrusNajmabadi: > The authorizerNames' values are empty arrays.

Wacky.  So nothing ever goes in there?

nodejs/awsx/apigateway/api.ts
```diff
@@ -581,6 +624,74 @@ function addEventHandlerRouteToSwaggerSpec(
     }
 }
 
+function addAuthorizersToSwagger(swagger: SwaggerSpec, authorizers: authorizer.LambdaAuthorizerDefinition[]): string[] {
+    const authNames: string[] = [];
+    swagger["securityDefinitions"] = swagger["securityDefinitions"] || {};
+
+    for (const auth of authorizers) {
+        const suffix = Object.keys(swagger["securityDefinitions"]).length;
```
CyrusNajmabadi: nit: i honestly think it would be easier to understand if this was just:

```ts
for (let i = 0; i < authorizers.length; i++) {
    const auth = authorizers[i];
    // and then just use 'i' as the suffix
```

nodejs/awsx/apigateway/api.ts
```diff
@@ -581,6 +624,74 @@ function addEventHandlerRouteToSwaggerSpec(
...
+            in: auth.parameterLocation,
+            "x-amazon-apigateway-authtype": auth.authType,
+            "x-amazon-apigateway-authorizer": getLambdaAuthorizer(authName, auth),
+        };
+
+        // Add security definition if it's a new authorizer
+        if (!swagger["securityDefinitions"][authName]) {
+            authNames.push(authName);
+            swagger["securityDefinitions"][authName] = securityDef;
+        } else if (swagger["securityDefinitions"][authName] !== securityDef) {
```
CyrusNajmabadi: this doesn't seem necessary.  either we have a collision on authName or not.  If we don't, we're good.  Otehrwise, we always throw.  

In other words, since you just defined securityDef above, if `!swagger["securityDefinitions"][authName]` is false, then `swagger["securityDefinitions"][authName] !== securityDef` will always be true, since hte existing sec-def will never equal `securityDef`

nodejs/awsx/apigateway/api.ts
```diff
@@ -606,17 +717,23 @@ function addStaticRouteToSwaggerSpec(
     // Create a bucket to place all the static data under.
     bucket = bucket || new aws.s3.Bucket(safeS3BucketName(name), undefined, parentOpts);
 
+    // Lazily initialize auth names
```
CyrusNajmabadi: i'm not sure what this comment means.  doesn't seem to add much value if it's just restating that we create auth-names if route.authorizers is present.

CyrusNajmabadi:
Overall this is looking great.  Structure makes sense and code is clean and fits into the existing flow nicely.

I don't have any concerns.  But i do have a suggestion that we look into how we can make it simpler for people to do common auth tasks.  I think it's *great* that there is the total flexibility of the full api at a user's disposal, and we'll def want to keep that.  

For example, to create a security group rule, you can supply the following:

```ts
    cidrBlocks?: pulumi.Input<pulumi.Input<string>[]>;
    description?: pulumi.Input<string>;
    fromPort: pulumi.Input<number>;
    ipv6CidrBlocks?: pulumi.Input<pulumi.Input<string>[]>;
    prefixListIds?: pulumi.Input<pulumi.Input<string>[]>;
    protocol: pulumi.Input<string>;
    self?: pulumi.Input<boolean>;
    sourceSecurityGroupId?: pulumi.Input<string>;
    toPort: pulumi.Input<number>;
```

but we also make it easy to just say: `secGroup.createIngressRule(name, { location: new AnyIPv4Location(), ports: new AllTcpPorts() })` which ends up setting all the necessayr values in a simpler manner.

I think we should examine if we can do something similar here.

nodejs/awsx/apigateway/README.md
ekrengel: it is indeed a C+P mistake

nodejs/awsx/apigateway/README.md
ekrengel: yes this is a request example

nodejs/awsx/apigateway/api.ts
ekrengel: I had seen comments somewhere else where we called out lazily initializing a variable, so I was following that. Probably not needed in this case

nodejs/awsx/apigateway/api.ts
ekrengel: The problem here is that multiple routes can add authorizers.So if you choose to not name the authorizers in 2 different routes you would end up w/ two different "authorizer-0". I am going to change the name to have a random 5 character suffix.

nodejs/awsx/apigateway/api.ts
CyrusNajmabadi: ah gotcha

nodejs/awsx/apigateway/api.ts
CyrusNajmabadi: ahhhhhhhh.  i missed that.  actually in that case your technique works well.  might be worth a comment... but it's probably ok without one.

nodejs/awsx/apigateway/README.md
```diff
@@ -202,6 +202,242 @@ let endpoint = new awsx.apigateway.API("example", {
...
 
+### Lambda Authorizers
+
+[Lambda Authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html) are AWS Lambda functions that provide control access to an API. You can define a Lambda Authorizer for an Event Handler Route or a Static Route. API Gateway supports `request` or `token` type Lambda authorizers. A `token` Lambda Authorizer uses an authorization token (i.e. a header in the form `Authorization: Token <token>`) to authorize the user, whereas a `request` Lambda Authorizer uses the request parameters (i.e. headers, path parameter or query parameters).
+
+To define an Authorizer, you provide a Lambda that fulfills `aws.lambda.EventHandler<AuthorizerEvent, AuthorizerResponse>` or you provide information on a pre-existing Lambda authorizer. The example below shows defining the Authorizer Lambda directly inline. See the [Event Handler Route](#Event-Handler-Route) section for other ways you can define a Lambda for the Authorizer.
+
+#### Token Authorizer
+
+Below is an example of a custom `request` Lambda Authorizer that uses `awsx.apigateway.getRequestLambdaAuthorizerDefinition` to simplify defining the authorizer.
```
CyrusNajmabadi: consider naming to `getRequestLambdaAuthorizer` as that seems sufficiently clear as to waht it does. This also fits the shape better as you have: `authorizers: [awsx.apigateway.getRequestLambdaAuthorizerDefinition({`.  If it were `authorizerDefinitions: ...` then `getRequestLambdaAuthorizerDefinition` would fit better.  But since you're calling these `authorizers` it makes sense to have the naming on the other side match as well.

nodejs/awsx/apigateway/README.md
```diff
@@ -202,6 +202,242 @@ let endpoint = new awsx.apigateway.API("example", {
...
+                statusCode: 200,
+                body: "<h1>Hello world!</h1>",
+            };
+        },
+        authorizers: [awsx.apigateway.getRequestLambdaAuthorizerDefinition({
+            queryParameters: ["auth"],
+            handler: async (event: awsx.apigateway.AuthorizerEvent) => {
+                // Add your own custom authorization logic here.
+                // Access the headers using event.headers, the query parameters using
+                // event.queryStringParameters or path parameters using event.pathParameters
```
CyrusNajmabadi: very helpful!

nodejs/awsx/apigateway/README.md
```diff
@@ -202,6 +202,242 @@ let endpoint = new awsx.apigateway.API("example", {
...
+        eventHandler: async () => {
+            return {
+                statusCode: 200,
+                body: "<h1>Hello world!</h1>",
+            };
+        },
+        authorizers: [awsx.apigateway.getTokenLambdaAuthorizerDefinition({
+            handler: async (event: awsx.apigateway.AuthorizerEvent) => {
+                // Add your own custom authorization logic here.
+                // Access the token using event.authorizationToken
```
CyrusNajmabadi: this comment doesn't seem necessary given the below :)

nodejs/awsx/apigateway/README.md
```diff
@@ -202,6 +202,242 @@ let endpoint = new awsx.apigateway.API("example", {
...
+        authorizers: [awsx.apigateway.getRequestLambdaAuthorizerDefinition({
+            queryParameters: ["auth"],
+            handler: async (event: awsx.apigateway.AuthorizerEvent) => {
+                // Add your own custom authorization logic here.
+                // Access the headers using event.headers, the query parameters using
+                // event.queryStringParameters or path parameters using event.pathParameters
+                return awsx.apigateway.AuthorizerResponse(
+                    "user",
+                    "Allow",
+                    event.methodArn);
```
CyrusNajmabadi: nit, unwrap these lines as they make what could be a simple single-liner (or 2-liner) into 4. (apples below as well).

nodejs/awsx/apigateway/README.md
```diff
@@ -202,6 +202,242 @@ let endpoint = new awsx.apigateway.API("example", {
...
+            parameterLocation: "header",
+            authType: "oauth2",
+            type: "request",
+            handler: callbackFxn,
+        }],
+    }],
+});
+```
+
+#### Using a Lambda Predefined Lambda
```
CyrusNajmabadi: Using a pre-existing AWS Lambda.

CyrusNajmabadi: i'm also not ure we need this section.

nodejs/awsx/apigateway/README.md
```diff
@@ -202,6 +202,242 @@ let endpoint = new awsx.apigateway.API("example", {
...
+        }],
+    }],
+});
+```
+
+A complete example of defining the Lambda Authorizer elsewhere can be found [here](https://github.com/pulumi/pulumi-awsx/blob/61d2996b8bdb20ea625e66e17ebbaa7b62f9c163/nodejs/awsx/examples/api/index.ts#L94-L152).
+
+#### Authorizer Response
+
+A helper function `awsx.apigateway.AuthorizerResponse` has been created to simplify generating the authorizer response. This can be used when defining the authorizer handler as follows:
```
CyrusNajmabadi: if it's a function, we should camelCase it.

CyrusNajmabadi: also, i'm not sure we need tihs.  you already use this helper above, so the readme naturally points people toward using this helper.

nodejs/awsx/apigateway/api.ts
```diff
@@ -58,6 +59,12 @@ export type EventHandlerRoute = {
     * defined at the API level.
     */
     requestValidator?: reqvalidation.RequestValidator;
+
+    /**
+     * Authorizers allows you to define Lambda authorizers be applied for authorization when the
+     * the route is called.
+     */
+    authorizers?: authorizer.LambdaAuthorizerDefinition[];
```
CyrusNajmabadi: as per above, let's just call tihs LambdaAuthorizer

nodejs/awsx/apigateway/api.ts
```diff
@@ -404,6 +419,30 @@ interface SwaggerOperation {
...
+    "x-amazon-apigateway-authorizer": LambdaAuthorizer;
+}
+
+interface LambdaAuthorizer {
+    type: "token" | "request";
+    authorizerUri: pulumi.Input<string>;
+    authorizerCredentials: pulumi.Input<string>;
+    identitySource?: string;
+    identityValidationExpression?: string;
+    authorizerResultTtlInSeconds?: number;
```
CyrusNajmabadi: note: if you're worried about colliding with this, we can just call this SwaggerLambdaAuthorizer.

nodejs/awsx/apigateway/api.ts
```diff
@@ -581,6 +626,87 @@ function addEventHandlerRouteToSwaggerSpec(
...
+
+function getLambdaAuthorizer(authorizerName: string, lambdaAuthorizer: authorizer.LambdaAuthorizerDefinition): LambdaAuthorizer {
+    if (authorizer.isLambdaAuthorizerInfo(lambdaAuthorizer.handler)) {
+        const identitySource = authorizer.getIdentitySource(lambdaAuthorizer.identitySource);
+        let uri: pulumi.Input<string>;
+        if (authorizer.isLambdaFunction(lambdaAuthorizer.handler.uri)) {
+            uri = lambdaAuthorizer.handler.uri.invokeArn;
+        } else {
+            uri = pulumi.output(lambdaAuthorizer.handler.uri);
+        }
```
CyrusNajmabadi: nit.  consider `?:` here.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,355 @@
...
+// These APIs are currently experimental and may change.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as awslambda from "aws-lambda";
+
+export type AuthorizerEvent = awslambda.CustomAuthorizerEvent;
+export type AuthorizerResponse = awslambda.CustomAuthorizerResult;
+export type AuthResponseContext = awslambda.AuthResponseContext;
```
CyrusNajmabadi: great. i'm glad we could use the authoritative types here.

nodejs/awsx/apigateway/authorizer.ts
CyrusNajmabadi: unresolving :)

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,355 @@
...
+     * The Uniform Resource Identifier (URI) of the authorizer Lambda function. The Lambda may also
+     * be passed directly, in which cases the URI will be obtained for you.
+     */
+    uri: pulumi.Input<string> | aws.lambda.Function;
+
+    /**
+     * Credentials required for invoking the authorizer in the form of an ARN of an IAM execution role.
+     * For example, "arn:aws:iam::account-id:IAM_role".
+     */
+    credentials: pulumi.Input<string>;
```
CyrusNajmabadi: consider taking `| aws.iamRole` as well.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,355 @@
...
+
+/**
+ * AuthorizerResponse simplifies creating an AuthorizerResponse.
+ * @param principalId - unique identifier for the user
+ * @param effect - whether to "Allow" or "Deny" the request
+ * @param resource - the API method to be invoked (typically event.methodArn)
+ * @param context - key-value pairs that are passed from the authorizer to the backend Lambda
+ * @param apiKey - if the API uses a usage plan, this must be set to one of the usage plan's API keys
+ */
+export function AuthorizerResponse(principalId: string, effect: Effect, resource: string, context?: AuthResponseContext, apiKey?: string): AuthorizerResponse {
```
CyrusNajmabadi: camelCase.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,355 @@
...
+        },
+    };
+
+    if (context) {
+        response.context = context;
+    }
+
+    if (apiKey) {
+        response.usageIdentifierKey = apiKey;
+    }
```
CyrusNajmabadi: is it necessary to do the actual `if` checks here? if you don't have a context or key, what's the harm in assigning that to the response?

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,355 @@
...
+ */
+export function getRequestLambdaAuthorizerDefinition(args: RequestAuthorizerArgs): LambdaAuthorizerDefinition {
+    let parameterName: string;
+    let location: "header" | "query";
+
+    const numQueryParams = getLength(args.queryParameters);
+    const numHeaders = getLength(args.headers);
+
+    if (numQueryParams === 0 && numHeaders === 0) {
+        throw new Error("Must specify at least one parameter");
```
CyrusNajmabadi: error message is a bit vague.  it can be useful to be more specific.  like `"[args.queryParameters] and [args.headers] were both empty.  At least one query parameter or header must be specified"`.

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,355 @@
...
+/** @internal */
+function getParameterName(args: RequestAuthorizerArgs, numHeaders: number, numQueryParameters: number): string {
+    if (numQueryParameters + numHeaders === 1) {
+        if (args.queryParameters) {
+            return args.queryParameters[0];
+        } else if (args.headers) {
+            return args.headers[0];
+        }
+    }
+    return "Unused";
```
CyrusNajmabadi: intententional?  what are the semantics if both are provided?

nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -0,0 +1,355 @@
...
+}
+
+/** @internal */
+function getLocation(numHeaders: number, numQueryParameters: number): "header" | "query" {
+    if (numHeaders > 0) {
+        return "header";
+    } else if (numQueryParameters > 0) {
+        return "query";
```
CyrusNajmabadi: it it an error to provide multiple headers and multiple queryparameters?  should we throw if they do that?

nodejs/awsx/apigateway/authorizer.ts
ekrengel: this is intentional -- if more than 1 is provided this must be "Unused"

nodejs/awsx/apigateway/authorizer.ts
ekrengel: not an error

nodejs/awsx/apigateway/api.ts
ekrengel: LambdaAuthorizer is not exposed here so this shouldnt be an issue for now.

PR #204, opened 2019-04-09
Add apikey support

ekrengel:
Should we add support to create the usage plan & apikeys during within the `awsx.apigateway.api` definition?


nodejs/awsx/examples/api/index.ts
```diff
@@ -86,4 +87,32 @@ const api = new awsx.apigateway.API("myapi", {
...
+
+/**
+ * Set up the Usage Plan for the API.
+ */
+const usagePlan = new aws.apigateway.UsagePlan("my-usage-plan", {
+    apiStages: [{
+        apiId: api.restAPI.id,
+        stage: api.stage.stageName,
+    }],
+});
```
CyrusNajmabadi: this feels like something that would be easy and convenient to create off of our awsx.apigateway.API.   Basically `const usagePlan = myApi.createUsagePlan("name")`.

nodejs/awsx/examples/api/index.ts
```diff
@@ -86,4 +87,32 @@ const api = new awsx.apigateway.API("myapi", {
...
+
+/**
+ * Create an API Key for testing purposes.
+ */
+const apikey = new aws.apigateway.ApiKey("my-key", {});
+const usagePlanKey = new aws.apigateway.UsagePlanKey("my-key", {
+    keyId: apikey.id,
+    keyType: "API_KEY",
+    usagePlanId: usagePlan.id,
+});
```
CyrusNajmabadi: what's the general relation between apikeys, plankeys, and plans?   are there 1:1 relations for any of them?  1:n?  n:m?

nodejs/awsx/examples/api/index.ts
ekrengel: Cool I agree! Just wanted to get your 👍 before I went down this path.

nodejs/awsx/examples/api/index.ts
ekrengel: - many usage plans per many APIs (in a usage plan you define things like rate & throttling limits & this can be applied across many APIs & stages)
- many apikeys per usage plan (this is typically 1 apikey per client)
- 1 plankey per apikey (this is what associates and apikey to a usage plan)

CyrusNajmabadi: ok great.  that will help inform how we can structure helpers here.  for example, since there are "many apikeys per usage plan",  we'd likely want `usagePlan.createApiKey`

ekrengel:
A bit unrelated to this... but wondering if there is a reason we are defaulting to 404 for access denied? I came across this when an authorizer returns a "Deny" effect the client receives a 404.

```ts
        "x-amazon-apigateway-gateway-responses": {
            "MISSING_AUTHENTICATION_TOKEN": {
                "statusCode": 404,
                "responseTemplates": {
                    "application/json": "{\"message\": \"404 Not found\" }",
                },
            },
            "ACCESS_DENIED": {
                "statusCode": 404,
                "responseTemplates": {
                    "application/json": "{\"message\": \"404 Not found\" }",
                },
            },
        },
```

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,142 @@
...
+            for (const a of args.apis) {
+                stages.push({
+                    apiId: a.restAPI.id,
+                    stage: a.stage.stageName,
+                });
+            }
+
+            if (args.usagePlan) {
+                if (args.usagePlan.apiStages) {
+                    throw new Error("[args.api] and [args.usagePlan.apiStages] cannot both be defined");
```
ekrengel: I wanted to avoid having to redefine `aws.apigateway.UsagePlanArgs` here, but I would like users to be able to define the apiStages via these args or by passing in `awsx.apigateway.API`. Not sure if theres a way to add an apiStage to the `apiStages` since its the following type:

```ts
apiStages?: pulumi.Input<pulumi.Input<{
        apiId: pulumi.Input<string>;
        stage: pulumi.Input<string>;
    }>[]>;
```

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,142 @@
...
+    usagePlanKey: aws.apigateway.UsagePlanKey;
+}
+
+/**
+ * Helper function that allows you to quickly create API Keys and associate them with an API.
+ *
+ * @param name The _unique_ name of the resource.
+ * @param args The arguments to use to populate this resource's properties.
+ */
+export function createAssociatedAPIKeys(name: string, args: APIKeyArgs): AssociatedAPIKeys {
```
ekrengel: I'd love to be able to add some unit tests here vs the integration tests for every possible input!

ekrengel:
@CyrusNajmabadi this is ready for review now

nodejs/awsx/apigateway/api.ts
```diff
@@ -626,6 +664,22 @@ function addEventHandlerRouteToSwaggerSpec(
     }
 }
 
+function addAPIkeyToSecurityDefinitions(swagger: SwaggerSpec) {
+    swagger.securityDefinitions = swagger.securityDefinitions || {};
+
+    if (swagger.securityDefinitions["api_key"] && swagger.securityDefinitions["api_key"] !== apiKeySecurityDefinition) {
+        throw new Error("Defined a none apikey security definition with the name api_key");
```
CyrusNajmabadi: 'none'?  or 'non-'?

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,147 @@
...
+    const usagePlan = getUsagePlan(name, args);
+    const keys = getKeys(name, args, usagePlan);
+
+    return {
+        usagePlan: usagePlan,
+        keys: keys,
+    };
```
CyrusNajmabadi: fyi, can simplify this as: `return { usagePlan, keys };`

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/** @internal */
+function getUsagePlan(name: string, args: APIKeyArgs): aws.apigateway.UsagePlan {
+    if (args.usagePlan && args.usagePlan instanceof aws.apigateway.UsagePlan) {
```
CyrusNajmabadi: sounds like you were having trouble with instanceof?  Shoudl this be CustomeResult.IsInstance?

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,147 @@
...
+     * Define the apis you would like to associate the usage plan with. This can be used in place of
+     * defining the apiStages defined in the [usagePlan]. You cannot define [apis] and [usagePlan.apiStages].
+     */
+    apis?: api.API[];
+
+    /**
+     * Define the usage plan to create. If an existing Usage Plan is passed in, then the API Keys
+     * will be associated with the usage plan and no [apis] can be added.
+     */
+    usagePlan?: aws.apigateway.UsagePlanArgs | aws.apigateway.UsagePlan;
```
CyrusNajmabadi: it feels like this is the most important thing, and should potentially be first.  i.e. if there is an [aws.apigwateway.UsagePlan] provdied, then we'll just use that.  Then, if we don't have that, we'll combine the UsagePlanArgs with [apis] to produce the usage plan.

Note: i had to read through a few time to figure out the set of combinations of all this stuff which are valid.  it might be worth enumerating.  i.e.:

1. You can provided [apis] and nothing else, in which case: ...
2. You can provider [apis] and [UsagePlanArgs], in which case...
3. You can provide a pre-existing UsagePlan.  In this case, [apis] cannot be provided.
etc. etc.

CyrusNajmabadi: Also proably worth mentioning what hapens if both are not provided.

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    if (args.apiKeys) {
+        for (let i = 0; i < args.apiKeys.length; i++) {
+            const currKey = args.apiKeys[i];
+            let apikey: aws.apigateway.ApiKey;
+
+            if (isAPIKey(currKey)) {
+                apikey = currKey;
+            } else {
+                apikey = new aws.apigateway.ApiKey(name + "-apikey-" + i, currKey);
```
CyrusNajmabadi: the `-apikey-` bit seems unnecessary.  is `${name}-${i}` not suitable here?

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            const usagePlanKey = new aws.apigateway.UsagePlanKey(name + "-usage-plan-key-" + i, {
```
CyrusNajmabadi: same comemnt about this name.

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,147 @@
...
+                keyId: apikey.id,
+                keyType: "API_KEY",
+                usagePlanId: usagePlan.id,
+            });
+
+            keys.push({
+                apikey: apikey,
+                usagePlanKey: usagePlanKey,
+            });
```
CyrusNajmabadi: nit, can be: `push({ apikey, usagePlanKey })`

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+    return keys;
+}
+
+/** @internal */
+export function isAPIKey(apikey: aws.apigateway.ApiKey | aws.apigateway.ApiKeyArgs): apikey is aws.apigateway.ApiKey {
+    return (<aws.apigateway.ApiKey>apikey).value !== undefined;
```
CyrusNajmabadi: sounds ike this should be CustomResult.isInstance?

ekrengel: I think most of the problems arose when we had a Pulumi.Input<T> and a CustomResource.. but I will change these all to be safe :)

nodejs/awsx/apigateway/apikey.ts
ekrengel: I think I've addressed this but please let me know if its still not clear!

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export interface AssociatedAPIKeys {
+    readonly usagePlan: aws.apigateway.UsagePlan;
+    readonly keys: Key[];
+}
+
+export interface Key {
+    apikey: aws.apigateway.ApiKey;
+    usagePlanKey: aws.apigateway.UsagePlanKey;
+}
```
CyrusNajmabadi: on the fence if these two exported interfaces should be doc'ed.  They're both nearly clear enough to be ok.  If you feel like they would benefit from an example in the docs, then doc'ing would be good.  However, if the docs are merely going to be boilerplate like "the usage place this api key is associated with", then we don't really need anything here.

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,149 @@
...
+                    quotaSettings: args.usagePlan.quotaSettings,
+                    throttleSettings: args.usagePlan.throttleSettings,
+                };
+            } else {
+                usagePlanArgs = {
+                    apiStages: stages,
+                };
+            }
+        }
+        return new aws.apigateway.UsagePlan(name, usagePlanArgs);
```
CyrusNajmabadi: i'd like something in the docs to mention that you can omit both args.usagePlan  and args.apis and you'll still get a UsagePLan created on your behalf with all default values.

nodejs/awsx/apigateway/apikey.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (args.apiKeys) {
+        for (let i = 0; i < args.apiKeys.length; i++) {
+            const currKey = args.apiKeys[i];
+            let apikey: aws.apigateway.ApiKey;
+
+            if (pulumi.CustomResource.isInstance(currKey)) {
+                apikey = currKey;
+            } else {
+                apikey = new aws.apigateway.ApiKey("${name}-${i}", currKey);
+            }
```
CyrusNajmabadi: tiny nit, and if you prefer this form feel free to keep it.  However, i find it easier to reason about 'const' variables.  And seeing blocks like this that are jsut used to initialize a single variable seems more complex than just having:

```ts
const apiKey = CustomResource.isInstance(currKey)
    ? currKey
    : new ...
```

Also note: our `?:` style is to wrap on the punctuation and align them indented on the next line.

--

If you prefer the if/else style though, that's fine.

PR #206, opened 2019-04-09
Add integration and raw data to README and examples

ekrengel:


nodejs/awsx/apigateway/README.md
```diff
@@ -94,6 +94,57 @@ let endpoint = new awsx.apigateway.API("example", {
...
+
+```ts
+import * as awsx from "@pulumi/awsx";
+
+let endpoint = new awsx.apigateway.API("example", {
+    routes: [{
+        path: "/integration",
+        target: {
+            uri: "https://www.google.com",
+            type: "http_proxy",
```
CyrusNajmabadi: if it's http_proxy, do you need to create a vpc_link: https://docs.aws.amazon.com/apigateway/latest/developerguide/getting-started-with-private-integration.html  ?  or is that optional for an `http_proxy`?

nodejs/awsx/apigateway/README.md
```diff
@@ -94,6 +94,57 @@ let endpoint = new awsx.apigateway.API("example", {
...
+        },
+    }],
+})
+```
+
+#### Raw Data Route
+
+A Raw Data Route is a fallback route for when raw swagger control is desired.  The `data` field should be an object that will be then included in the final swagger specification.
+
```
CyrusNajmabadi: do we want to perhaps link to https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration.html or another useful reference doc for the "x-amazon-apigateway-integration" swagger point?

nodejs/awsx/examples/api/index.ts
```diff
@@ -82,6 +82,30 @@ const api = new awsx.apigateway.API("myapi", {
...
+                "passthroughBehavior": "when_no_match",
+                "httpMethod": "GET",
+                "type": "http_proxy",
+            },
+        },
+    }, {
+        path: "/integration",
+        target: {
+            uri: "https://www.google.com",
+            type: "http_proxy",
```
CyrusNajmabadi: does this actually work?

ekrengel: it does -- it just returns a blob from google

nodejs/awsx/apigateway/README.md
ekrengel: i think its optional since it works as is

CyrusNajmabadi:
LGTM.  this might be another area where we mght benefit from helpers *in the future* to allow people to easily make integrations to services.

PR #210, opened 2019-04-13
Properly parent VPC resources.

CyrusNajmabadi:


nodejs/awsx/ec2/vpc.ts
```diff
@@ -124,8 +126,6 @@ export class Vpc extends pulumi.ComponentResource {
             throw new Error("Cannot add InternetGateway to Vpc that already has one.");
         }
 
-        opts = { parent: this, ...opts };
-
         // See https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Internet_Gateway.html#Add_IGW_Attach_Gateway
         // for more details.
         this.internetGateway = new x.ec2.InternetGateway(name, this, args, opts);
```
lukehoban: Why is this forwarding `opts` instead of just parenting the resource?

CyrusNajmabadi: yeah, good point.  will fix.

CyrusNajmabadi: oh, nevermind. this is the instance method on vpc to make a natgateway.  We basically will parent with 'this vpc' by default, but we allow the caller to pass in their own opts in case they want to control this.  so ths is by design.  

The *direct* resources a component creates (i.e. in the constructor), are parented by teh component.  however, indirect resources (i.e. resources that can be created by a external party *from* the resource), will parent themselves by default, but will allow overriding by the external party.

CyrusNajmabadi: Note: this is also why we use the pattern `{ parent: x, ...opts }`.  That way, if `opts` includes a parent, it is respected. However, if it doesn't include one, we can pick a reasonable default.

nodejs/awsx/ec2/internetGateway.ts
```diff
@@ -38,7 +39,7 @@ export class InternetGateway
             this.internetGateway = new aws.ec2.InternetGateway(name, {
                 ...args,
                 vpcId: vpc.vpc.id,
-            }, { parent: this });
+            }, parentOpts);
```
CyrusNajmabadi: this is stylistic.  But it matches what we do elsewhere, so i'd like it here for consistency.

nodejs/awsx/ec2/vpc.ts
```diff
@@ -71,12 +73,12 @@ export class Vpc extends pulumi.ComponentResource {
                 enableDnsHostnames: utils.ifUndefined(args.enableDnsHostnames, true),
                 enableDnsSupport: utils.ifUndefined(args.enableDnsSupport, true),
                 instanceTenancy: utils.ifUndefined(args.instanceTenancy, "default"),
-            });
+            }, parentOpts);
```
CyrusNajmabadi: this was the root issue.  we were creating the underlying resource without properly parenting it to the awsx.ec2.Vpc.

nodejs/awsx/ec2/vpc.ts
```diff
@@ -71,12 +73,12 @@ export class Vpc extends pulumi.ComponentResource {
...
             this.id = this.vpc.id;
 
             // Create the appropriate subnets.  Default to a single public and private subnet for each
             // availability zone if none were specified.
-            const topology = new VpcTopology(this, name, cidrBlock, numberOfAvailabilityZones, opts);
+            const topology = new VpcTopology(this, name, cidrBlock, numberOfAvailabilityZones);
```
CyrusNajmabadi: opts was not desirable here.  We don't want to pass along opts passed into us to child resource.  We just want to pass 'this' along which can be used as hte parent of all children.

nodejs/awsx/ec2/vpc.ts
CyrusNajmabadi: this is inside an instance helper method on top of awsx.ec2.Vpc.  With the instance methods, we pass along 'this' and 'opts' untouched.  'opts' allows the caller to choose what they want as options for this auxilliary resource.  But we also pass 'this' in case they do not specify a parent.  In that case, we will parent these auxilliary resources with 'this' resource as it can sensibly be a parent here.

lukehoban:
@CyrusNajmabadi is this a breaking change?  By introducing a new parent, will that cause all VPCs to be replaced on updates?  If so, we should include in change-log and probably make part of a major version?

PR #215, opened 2019-04-17
fix typechecking

ekrengel:


nodejs/awsx/apigateway/authorizer.ts
```diff
@@ -90,33 +90,23 @@ export interface LambdaAuthorizer {
...
 
-/** @internal */
-export function isLambdaFunction(uri: pulumi.Input<string> | aws.lambda.Function): uri is aws.lambda.Function {
-    return (<aws.lambda.Function>uri).invokeArn !== undefined;
-}
-
-/** @internal */
-export function isIAMRole(creds: pulumi.Input<string> | aws.iam.Role): creds is aws.iam.Role {
-    return (<aws.iam.Role>creds).assumeRolePolicy !== undefined;
-}
```
CyrusNajmabadi: yeah, this approach def won't work to distinguish a Role from an Output.  However, we do need to figure out why `instanceof aws.iam.Role` isn't working here.  That should work properly.

ekrengel: I talk to Sean about instanceof not working properly (A string gets identified properly by a lambda gets identified as not a lambda). He thinks it may be a result of a side by side dependency load. We tried to resolve this but has no luck For now, I have used something list this:

```ts
        let uri: pulumi.Input<string>;
        if (pulumi.CustomResource.isInstance(lambdaAuthorizer.handler.uri)) {
            uri = lambdaAuthorizer.handler.uri.invokeArn;
        } else {
            uri = lambdaAuthorizer.handler.uri;
        }
```

CyrusNajmabadi: yup, that's another way things could be done.  i'm really curious how you're having a sxs issues here.  but teh .isInstance approach will work well for your needs here.

PR #216, opened 2019-04-17
add cognito user pool authorizers

ekrengel:


nodejs/awsx/apigateway/README.md
```diff
@@ -416,6 +416,28 @@ const api = new awsx.apigateway.API("myapi", {
...
+const cognitoUserPool = new aws.cognito.UserPool("pool", {});
+
+// Create an API that requires authorizes against the User Pool
+const apiWithCognitoAuthorizer = new awsx.apigateway.API("cognito-protected-api", {
+    routes: [{
+        path: "/www_old",
+        localPath: "www",
+        authorizers: [awsx.apigateway.getCognitoAuthorizer({
+            providerARNs: [cognitoUserPool],
+        })],
```
CyrusNajmabadi: note: in cases where we allow arrays, it's often nice to allow a single item to be provided as well.  it looks a little strange having:

```ts
authorizers: awsx.apigateway.getCognitoAuthorizer(...),
```
But it's nice to avoid just the extra cruft of `[...]`

nodejs/awsx/apigateway/api.ts
```diff
@@ -445,6 +452,13 @@ interface LambdaAuthorizer {
     authorizerResultTtlInSeconds?: number;
 }
 
+interface CognitoAuthorizer {
+    type: "cognito_user_pools";
+    identitySource: string;
```
CyrusNajmabadi: intentional that identitySource is required here, but optional with LambdaAuthorizer?

nodejs/awsx/apigateway/api.ts
```diff
@@ -20,12 +20,14 @@ import * as fspath from "path";
 
 import * as aws from "@pulumi/aws";
 import * as pulumi from "@pulumi/pulumi";
-import * as authorizer from "./authorizer";
+
+import * as cognitoauthorizer from "./cognitoauthorizer";
+import * as lambdaauthorizer from "./lambdaauthorizer";
```
CyrusNajmabadi: capitalize Authorizer for these (including the files).

nodejs/awsx/apigateway/api.ts
```diff
@@ -628,48 +642,80 @@ function addEventHandlerRouteToSwaggerSpec(
...
+                    type: "apiKey",
+                    name: auth.parameterName,
+                    in: auth.parameterLocation,
+                    "x-amazon-apigateway-authtype": auth.authType,
+                    "x-amazon-apigateway-authorizer": {
+                        type: auth.authType,
+                        identitySource: lambdaauthorizer.getIdentitySource(auth.identitySource),
+                        providerARNs: getCognitoPoolARNs(auth.providerARNs),
+                    },
+                };
```
CyrusNajmabadi: securityDefs in both branches seem the same (except for the actual authorizer portion).  Consider unifying more of the code and making that the only place where they need to differ.

nodejs/awsx/apigateway/cognitoauthorizer.ts
```diff
@@ -0,0 +1,146 @@
...
+export interface CognitoAuthorizer {
+    /**
+     * The name for the Authorizer to be referenced as. This must be unique for each unique
+     * authorizer within the API. If no name if specified, a name will be generated for you.
+     */
+    authorizerName?: string;
+
+    /**
+     * parameterName is the name of the header containing the authorization token.
+     * */
```
CyrusNajmabadi: nit: wrapping of `*/`

nodejs/awsx/apigateway/cognitoauthorizer.ts
```diff
@@ -0,0 +1,146 @@
...
+    parameterName: string;
+
+    /**
+     * Defines where in the request API Gateway should look for identity information. The value must
+     * be "header" for a Cognito Authorizer.
+     */
+    parameterLocation: "header";
```
CyrusNajmabadi: as this is a public API, this is just something onerous for the user.  We should remove and do this on their behalf.

nodejs/awsx/apigateway/cognitoauthorizer.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Specifies the authorization mechanism for the client.
+     */
+    authType: "cognito_user_pools";
```
CyrusNajmabadi: same.

nodejs/awsx/apigateway/cognitoauthorizer.ts
CyrusNajmabadi: oh.  i think i see.  this is *not* what the user wuld create.  instead, they woudl use the 'args' below to get one of these?  in that case, this is likely fine.

nodejs/awsx/apigateway/cognitoauthorizer.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * The name for the Authorizer to be referenced as. This must be unique for each unique
+     * authorizer within the API. If no name if specified, a name will be generated for you.
+     */
+    authorizerName?: string;
+
+    /**
+     * The request header for the authorization token. If not set, this defaults to
+     * Authorization.
```
CyrusNajmabadi: consider "Authorization" to make it clear that is a literal value.

nodejs/awsx/apigateway/api.ts
ekrengel: Yes per [docs](https://docs.aws.amazon.com/apigateway/api-reference/resource/authorizer/#identitySource)

The strange thing is that in some [examples](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-enable-cognito-user-pool.html) I dont see this set.

nodejs/awsx/apigateway/cognitoauthorizer.ts
ekrengel: Yes the hope is they use the helper function but could also choose to define a `CognitoAuthorizer` directly. I will take these out and just add them where I create the SwaggerCognitoAuthorizer.

nodejs/awsx/apigateway/README.md
ekrengel: I will do this in a separate PR after this gets merged since it will affect both cognito and lambda auhtorizers.

nodejs/awsx/apigateway/api.ts
ekrengel: Either way we do this for the user so they wont have to deal w it.

ekrengel:
@CyrusNajmabadi this ready for another pass

nodejs/awsx/apigateway/README.md
CyrusNajmabadi: sounds good.

nodejs/awsx/apigateway/api.ts
```diff
@@ -605,37 +607,67 @@ function addAPIKeyToSwaggerOperation(swaggerOperation: SwaggerOperation) {
...
-            swagger["securityDefinitions"][auth.authorizerName] = securityDef;
         }
-        authNames.push(auth.authorizerName);
+
+        const methods = lambdaAuthorizer.isLambdaAuthorizer(auth) || !auth.methodsToAuthorize
+            ? []
+            : auth.methodsToAuthorize;
+
+        authRecords.push({ [authName]: methods });
+
```
CyrusNajmabadi: nit: extra line.

nodejs/awsx/apigateway/api.ts
```diff
@@ -605,37 +607,67 @@ function addAPIKeyToSwaggerOperation(swaggerOperation: SwaggerOperation) {
...
     }
-    return authNames;
+    return authRecords;
+}
+
+function getCognitoPoolARNs(pools: Array<pulumi.Input<string> | aws.cognito.UserPool>): pulumi.Input<string>[] {
```
CyrusNajmabadi: nit: up to you  but return value is optional her.e  if you like it for documentation purposes, that's fine.

nodejs/awsx/apigateway/cognitoAuthorizer.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * parameterName is the name of the header containing the authorization token.
+     */
+    parameterName: string;
+
+    /**
+     * The ARNs of the Cognito User Pools to use.
+     */
+    providerARNs: Array<pulumi.Input<string> | aws.cognito.UserPool>;
```
CyrusNajmabadi: fyi: `(pulumi.Input<string> | aws.cognito.UserPool)[]` is also a way to write this.  Up to you which you prefer.

nodejs/awsx/apigateway/cognitoAuthorizer.ts
```diff
@@ -0,0 +1,133 @@
...
+    const parameterName = args.header || "Authorization";
+
+    return {
+        authorizerName: args.authorizerName,
+        parameterName: parameterName,
+        providerARNs: args.providerARNs,
+        identitySource: ["method.request.header." + parameterName],
+        identityValidationExpression: args.identityValidationExpression,
+        authorizerResultTtlInSeconds: args.authorizerResultTtlInSeconds,
+        methodsToAuthorize: args.methodsToAuthorize,
```
CyrusNajmabadi: technically can be simplified to:

```ts
return {
    ...args,
    identitySource: ["method.request.header." + parameterName],
};
```

nodejs/awsx/apigateway/api.ts
ekrengel: If I dont assign it here then the compiler complains that:

Type 'string' is not assignable to type '"cognito_user_pools"'.

nodejs/awsx/apigateway/cognitoAuthorizer.ts
ekrengel: parameterName: parameterName cause a problem here... but this is a cool tip!

CyrusNajmabadi: you're right, you'd need to write:

```ts
return {
    ...args,
    parameterName,
    identitySource: ["method.request.header." + parameterName],
};
```

PR #223, opened 2019-04-18
Expose API lambdas that are created for the user.

CyrusNajmabadi:


nodejs/awsx/apigateway/api.ts
```diff
@@ -360,16 +362,29 @@ export class API extends pulumi.ComponentResource {
...
         }, { parent: this, dependsOn: permissions });
 
+        this.swaggerLambdas = swaggerLambdas || {};
+
         this.registerOutputs({});
     }
+
+    /**
+     * Returns the [aws.lambda.Function] an [EventHandlerRoute] points to.  This will either be for
+     * the aws.lambda.Function created on your behalff if the route was passed a normal
```
ekrengel: behalf

nodejs/awsx/apigateway/api.ts
```diff
@@ -299,11 +299,13 @@ export interface APIArgs {
...
-    public deployment: aws.apigateway.Deployment;
-    public stage: aws.apigateway.Stage;
+    public readonly restAPI: aws.apigateway.RestApi;
+    public readonly deployment: aws.apigateway.Deployment;
+    public readonly stage: aws.apigateway.Stage;
 
-    public url: pulumi.Output<string>;
+    public readonly url: pulumi.Output<string>;
+
+    private readonly swaggerLambdas: SwaggerLambdas;
```
ekrengel: Do want to do the same for authorizers? Maybe thats an add on down the line.

CyrusNajmabadi: in general, yes :) Basically any resource we create on their behalf should be reachable somehow.  But i'm willing to wait o it :)

PR #232, opened 2019-04-19
Add the ability to scale an autoscaling group based on a metric.

CyrusNajmabadi:
Note: there will be more PRs to come after this.  Specifically, we need to cover simple scaling and step scaling.  

There also needs to be full readmes on this.


nodejs/awsx/ecs/service.ts
```diff
@@ -233,7 +253,7 @@ export interface ServiceArgs {
      * that use the `awsvpc` network mode to receive their own Elastic Network Interface, and it is
      * not supported for other network modes.
      */
-    networkConfiguration?: aws.ecs.ServiceArgs["networkConfiguration"];
+    networkConfiguration?: pulumi.Input<NetworkConfiguration>;
```
CyrusNajmabadi: No change.  But this gives us docs on things we didn't have docs for before.

nodejs/awsx/elasticloadbalancingv2/targetGroup.ts
```diff
@@ -203,7 +203,7 @@ export interface TargetGroupArgs {
     /**
      * The type of target that you must specify when registering targets with this target group. The
      * possible values are `instance` (targets are specified by instance ID) or `ip` (targets are
-     * specified by IP address). The default is `instance`. Note that you can't specify targets for
+     * specified by IP address). The default is `ip`. Note that you can't specify targets for
```
CyrusNajmabadi: this comment was incorrect.  this now matches our actual default behavior.

nodejs/awsx/autoscaling/policy.ts
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * The target value for the metric.
+     */
+    targetValue: pulumi.Input<number>;
+}
+
+export interface ApplicationTargetGroupTrackingPolicyArgs extends TargetTrackingPolicyArgs {
+    /**
+     * The target group to scale [AutoScalingGroup] in response to number of requests to.
+     * If not provided, the first [TargetGroup] in [AutoScalingGroup.targetGroups] will be used.
```
ekrengel: This seems like a risky default. For example, what if it's working as expected and then something gets added to the `AutoScalingGroup.targetGroups` and now the thing doesn't scale as expected anymore. I'm guessing the hope is they catch this in a diff but still seems like something we should require the user specify.

CyrusNajmabadi: Yes.  i see your point.  I'll make this required.

PR #235, opened 2019-04-19
Make it easier to specify cron schedules in JS/TS.

CyrusNajmabadi:


nodejs/awsx/autoscaling/schedule.ts
```diff
@@ -53,11 +106,11 @@ export interface ScheduleRecurrenceArgs {
     /** 1 to 31.  Leave undefined to indicate no specific value. */
     dayOfMonth?: number;
 
-    /** 1 to 12.  Leave undefined to indicate no specific value. */
-    month?: number;
+    /** Month of the year to perform the scheduled action on.  Leave undefined to indicate no specific value. */
+    month?: Month;
```
ekrengel: Should this be able to take both a number or a Month? (same w DayOfWeek)

CyrusNajmabadi: it does.  'Month' is simply defined as 'number | "..."`

PR #249, opened 2019-04-22
Break out short/long tests for fargate.

CyrusNajmabadi:


CyrusNajmabadi:
This also only costs a single load balancer, making it easier to run multiple tests simultaneously.

PR #250, opened 2019-04-22
Ekrengel/add new features to integration routes

ekrengel:


nodejs/awsx/apigateway/api.ts
```diff
@@ -155,6 +155,32 @@ function isStaticRoute(route: Route): route is StaticRoute {
...
+     * the API level and by default, the source will be the HEADER.
+     */
+    apiKeyRequired?: boolean;
+
+    /**
+     * Authorizers allows you to define Lambda authorizers be applied for authorization when the
+     * the route is called. The authorizer will get applied to all static resources defined by the
+     * localPath.
+     */
+    authorizers?: Authorizer[] | Authorizer;
```
CyrusNajmabadi: these four now seem common to most routes (except 'raw json') right?  If so, can we extract out to `interface BaseRoute` and then have the others extend from that so we don't have to copy this into all of them?

nodejs/awsx/apigateway/api.ts
```diff
@@ -986,11 +1012,31 @@ function addIntegrationRouteToSwaggerSpec(
...
+    }
+    if (route.requestValidator) {
+        swaggerOpWithoutProxyPathParam["x-amazon-apigateway-request-validator"] = route.requestValidator;
+        swaggerOpWithProxyPathParam["x-amazon-apigateway-request-validator"] = route.requestValidator;
+    }
+    if (route.apiKeyRequired) {
+        addAPIkeyToSecurityDefinitions(swagger);
+        addAPIKeyToSwaggerOperation(swaggerOpWithoutProxyPathParam);
+        addAPIKeyToSwaggerOperation(swaggerOpWithProxyPathParam);
+    }
```
CyrusNajmabadi: not possible to extract a function that will do the 4 steps, but then call it for both swaggerOpWithoutProxyPathParam and swaggerOpWithProxyPathParam?

CyrusNajmabadi: feels like we've triplicated this with addEventHandlerRouteToSwaggerSpec and addStaticRouteToSwaggerSpec (Esp for the file/dir handling).  This feels like this must be extractable.

nodejs/awsx/aws_test.go
```diff
@@ -489,23 +507,23 @@ func validateAPITests(apiTests []apiTest) func(t *testing.T, stack integration.R
 
 			if tt.requiredToken != nil {
 				resp := GetHTTP(t, req, 401)
-				assertRequestBody(t, `{"message":"Unauthorized"}`, resp)
+				assertRequestBody(t, `{"message":"Unauthorized"}`, false, resp)
```
CyrusNajmabadi: can you do `false /*skipBodyValidation*/`?  Passing constants like this makes it hard to tell at his callsite what's going on.

nodejs/awsx/apigateway/api.ts
ekrengel: Its not in the Raw Data Route. I'll spend some time thinking about how to add it to that..

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 0.18.3 (Unreleased)
 
+- Add Typescript support for Authorizers, API Keys and Request Validation to Integration Routes in API Gateway
```
CyrusNajmabadi: remove 'TypeScript'.

PR #256, opened 2019-04-25
Expose the 'disableRollback' option when creating an AutoScalingGroup.

CyrusNajmabadi:


nodejs/awsx/autoscaling/autoscaling.ts
```diff
@@ -341,6 +341,18 @@ export interface AutoScalingGroupArgs {
      * have the "instance" [targetType].
      */
     targetGroups?: x.elasticloadbalancingv2.TargetGroup[];
+
+    /**
+     * Set to true to disable rollback of the underlying aws.cloudformation.Stack if that Stack
+     * creation failed.  Defaults to 'false'.  Conflicts with `onFailure`.
+     */
+    disableRollback?: pulumi.Input<boolean>;
```
CyrusNajmabadi: i wanted to default this to 'true', however changing that doesn't work for an existing stack because you get:

```
Plan apply failed: Creating CloudFormation stack failed: AlreadyExistsException: Stack [testing-1-2819e76] already exists
```

But this at least exposes the signatures on the TS side.  Note: this was always something you coudl provide on the JS side since we pass all the args over to the underlying stack.

PR #262, opened 2019-05-01
Update lambda permissions sourceArn to reflect correct path

mnbbrown:
See AWS docs: https://docs.aws.amazon.com/lambda/latest/dg/with-on-demand-https-example.html#with-on-demand-https-add-permission

### DIFF
--- a/nodejs/awsx/apigateway/api.ts
+++ b/nodejs/awsx/apigateway/api.ts
@@ -399,7 +399,7 @@ function createLambdaPermissions(api: API, name: string, swaggerLambdas: Swagger
     const permissions: aws.lambda.Permission[] = [];
     for (const [path, lambdas] of swaggerLambdas) {
         for (const [method, lambda] of lambdas) {
-            const methodAndPath = `${method === "ANY" ? "*" : method}${path}`;
+            const methodAndPath = `${method === "ANY" ? "*" : method}/${path}`;
 
             permissions.push(new aws.lambda.Permission(name + "-" + sha1hash(methodAndPath), {
                 action: "lambda:invokeFunction",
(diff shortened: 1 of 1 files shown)


nodejs/awsx/apigateway/api.ts
```diff
@@ -399,7 +399,7 @@ function createLambdaPermissions(api: API, name: string, swaggerLambdas: Swagger
     const permissions: aws.lambda.Permission[] = [];
     for (const [path, lambdas] of swaggerLambdas) {
         for (const [method, lambda] of lambdas) {
-            const methodAndPath = `${method === "ANY" ? "*" : method}${path}`;
+            const methodAndPath = `${method === "ANY" ? "*" : method}/${path}`;
```
CyrusNajmabadi: this doesnt seem right.  the path will normally have the / in it already.  so this would likely break that, right?

CyrusNajmabadi:
> @CyrusNajmabadi @ekrengel if we do intend to support paths without a leading / should we do the normalization earlier?

Yes, that's probably the best thing to do.  It's what we did over back in cloud-land.  Looks like we didn't take that approach down here, nor did we explicitly validate that the paths must start with a `/`:

https://github.com/pulumi/pulumi-cloud/blob/a87477e07f8255b49cffe9766f9d5d1ddf4794b2/aws/api.ts#L77-L96

Note: i'm also fine just not supporting these paths and throwing.  I'm fine with either approach.  Thoughts @lukehoban ?  One is friendlier but less clear in intent (IMO).  THe other is less friendly, but makes it extremely clear at the callsite what's going on.

CyrusNajmabadi:
Hey @mnbbrown Thanks for the contribution!  We're going to fix this issue, but through https://github.com/pulumi/pulumi-awsx/pull/263 instead.  Effectively, we'll document that you can pass an unrooted path to API, but we'll automatically add the / to it on your behalf.

mnbbrown:
@CyrusNajmabadi 👍 thanks for considering it!

Issue #276, opened 2019-05-15
Make it easier for VPCs to use all availability zones


joeduffy:
It would be nice if `numberOfAvailabilityZones` automatically used the number of AZs in my current region, or at least offered the option to do so.

Unfortunately, because this is a `number`, doing so manually requires a bit of async tedium:

```typescript
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";

module.exports = (async () => {
    const azs = await aws.getAvailabilityZones();
    const vpc = new awsx.ec2.Vpc("my-vpc", {
        numberOfAvailabilityZones: azs.names.length,
    });
    return {
        vpcId: vpc.id,
    };
})();
```

This is certainly not impossible, but any of the following would have been much nicer:

_*Option 1.* Automatically use all AZs by default.

```typescript
import * as awsx from "@pulumi/awsx";
const vpc = new awsx.ec2.Vpc("my-vpc");
export vpcId = vpc.id;
```

_*Option 2.* Accept an `auto`  parameter for `numberOfAvailabilityZones` (opt in).

```typescript
import * as awsx from "@pulumi/awsx";
const vpc = new awsx.ec2.Vpc("my-vpc", { numberOfAvailabilityZones: "auto" });
export vpcId = vpc.id;
```

_*Option 3.* Accept a `Promise<number>` for `numberOfAvailabilityZones`.

```typescript
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";
const vpc = new awsx.ec2.Vpc("my-vpc", {
    numberOfAvailabilityZones: aws.getAvailabilityZones().then(az => az.names.length),
});
export vpcId = vpc.id;
```

CyrusNajmabadi:
This is hard.  THe main issue is that we need the az info in order to create resources (i.e. it affects the subnets we create).  Thus, we need to be able to call into our provider functions to do this.  Which makes VPC constructoin async, which makes thigns *hard* and has a lot of downstream effects.  

If we can make getAvailabilityZones sync (and i think we can), then this becomes much nicer.

PR #278, opened 2019-05-16
 Dynamically determine the number of availability zones for a VPC to use.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/276

### DIFF
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -15,6 +15,8 @@
 import * as aws from "@pulumi/aws";
 import * as pulumi from "@pulumi/pulumi";
 
+import * as deasync from "deasync";
+
 import * as x from "..";
 import { getAvailabilityZone } from "./../aws";
 import { VpcTopology } from "./vpcTopology";
@@ -60,7 +62,8 @@ export class Vpc extends pulumi.ComponentResource {
         }
         else {
             const cidrBlock = args.cidrBlock === undefined ? "10.0.0.0/16" : args.cidrBlock;
-            const numberOfAvailabilityZones = args.numberOfAvailabilityZones === undefined ? 2 : args.numberOfAvailabilityZones;
+            const numberOfAvailabilityZones = getNumberOfAvailabilityZones(args.numberOfAvailabilityZones);
+
             const numberOfNatGateways = args.numberOfNatGateways === undefined ? numberOfAvailabilityZones : args.numberOfNatGateways;
             if (numberOfNatGateways > numberOfAvailabilityZones) {
                 throw new Error(`[numberOfNatGateways] cannot be greater than [numberOfAvailabilityZones]: ${numberOfNatGateways} > ${numberOfAvailabilityZones}`);
@@ -225,6 +228,19 @@ export class Vpc extends pulumi.ComponentResource {
 
 (<any>Vpc.prototype.addInternetGateway).doNotCapture = true;
 (<any>Vpc.prototype.addNatGateway).doNotCapture = true;
+
+function getNumberOfAvailabilityZones(requestedCount: number | undefined) {
+    if (requestedCount !== undefined) {
+        return requestedCount;
+    }
+
+    const availabilityZones = utils.promiseResult(aws.getAvailabilityZones());
+    if (availabilityZones && availabilityZones.names && availabilityZones.names.length > 0) {
+        return availabilityZones.names.length;
+    }
+
+    return 2;
+}
 
 function createNatGateways(vpcName: string, vpc: Vpc, numberOfAvailabilityZones: number, numberOfNatGateways: number) {
     // Create nat gateways if we have private subnets and we have public subnets to place them in.
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -229,14 +229,16 @@ export class Vpc extends pulumi.ComponentResource {
 (<any>Vpc.prototype.addInternetGateway).doNotCapture = true;
 (<any>Vpc.prototype.addNatGateway).doNotCapture = true;
 
-function getNumberOfAvailabilityZones(requestedCount: number | undefined) {
-    if (requestedCount !== undefined) {
+function getNumberOfAvailabilityZones(requestedCount: "all" | number | undefined) {
+    if (typeof requestedCount === "number") {
         return requestedCount;
     }
 
-    const availabilityZones = utils.promiseResult(aws.getAvailabilityZones());
-    if (availabilityZones && availabilityZones.names && availabilityZones.names.length > 0) {
-        return availabilityZones.names.length;
+    if (requestedCount === "all") {
+        const availabilityZones = utils.promiseResult(aws.getAvailabilityZones());
+        if (availabilityZones && availabilityZones.names && availabilityZones.names.length > 0) {
+            return availabilityZones.names.length;
+        }
     }
 
     return 2;
@@ -389,10 +391,10 @@ export interface VpcArgs {
     subnets?: VpcSubnetArgs[];
 
     /**
-     * The maximum number of availability zones to use in the current region.  Defaults to '2' if
-     * unspecified.
+     * The maximum number of availability zones to use in the current region.  Defaults to `2` if
+     * unspecified.  Use `"all"` to use all the availability zones in the current region.
      */
-    numberOfAvailabilityZones?: number;
+    numberOfAvailabilityZones?: number | "all";
 
     /**
      * The number of NAT gateways to create if there are any private subnets created.  A NAT gateway
--- a/nodejs/awsx/utils.ts
+++ b/nodejs/awsx/utils.ts
@@ -15,6 +15,8 @@
 import * as aws from "@pulumi/aws";
 import * as pulumi from "@pulumi/pulumi";
 
+import * as deasync from "deasync";
+
 import * as crypto from "crypto";
 
 type Diff<T extends string | number | symbol, U extends string | number | symbol> =
@@ -102,4 +104,34 @@ export function hasTrueBooleanMember(obj: any, memberName: string | number | sym
     }
 
     return val === true;
+}
+
+/** @internal */
+export function promiseResult<T>(promise: Promise<T>): T {
+    enum State {
+        running,
+        finishedSuccessfully,
+        finishedWithError,
+    }
+
+    let result: T;
+    let error = undefined;
+    let state = <State>State.running;
+
+    promise.then(
+        val => {
+            result = val;
+            state = State.finishedSuccessfully;
+        },
+        err => {
+            error = err;
+            state = State.finishedWithError;
+        });
+
+    deasync.loopWhile(() => state === State.running);
+    if (state === State.finishedWithError) {
+        throw error;
+    }
+
+    return result!;
 }
--- a/nodejs/awsx/package.json
+++ b/nodejs/awsx/package.json
@@ -15,12 +15,14 @@
         "@pulumi/aws": "^0.18.1",
         "@pulumi/docker": "^0.17.0",
         "@types/aws-lambda": "^8.10.23",
-        "mime": "^2.0.0"
+        "mime": "^2.0.0",
+        "deasync": "^0.1.15"
     },
     "devDependencies": {
         "@types/aws-sdk": "^2.7.0",
         "@types/mime": "^2.0.0",
         "@types/node": "^10.0.0",
+        "@types/deasync": "^0.1.0",
         "tslint": "^5.11.0",
         "typedoc": "^0.13.0",
         "typescript": "^3.4.1"


CyrusNajmabadi:
Ping @swgillespie @joeduffy @lukehoban.  I'm sending this PR out to start a discussion around how we could use provider functions in a simpler fashion.

This PR is interesting in that it allows us to call a provider function at runtime, but in a synchronous fashion.  To do this, it uses the [deasync](https://github.com/abbr/deasync) library to allow us to simulate blocking on an async operation.  For the purposes of this PR, this allows us to call the GetAvailabilityZones API, block on getting the result, and then perform flow control based on the result.

I personally think this works out quite nicely.  If we like this, we can update our generated provider packages to expose sync versions of all these functions, vastly simplifying the consumption experience for customers (by allowing them to avoid async/await).

nodejs/awsx/ec2/vpc.ts
```diff
@@ -373,10 +391,10 @@ export interface VpcArgs {
     subnets?: VpcSubnetArgs[];
 
     /**
-     * The maximum number of availability zones to use in the current region.  Defaults to '2' if
-     * unspecified.
+     * The maximum number of availability zones to use in the current region.  Defaults to `2` if
+     * unspecified.  Use `"all"` to use all the availability zones in the current region.
```
CyrusNajmabadi: while i'd like to have 'undefined' mean 'use all', that could be deeply impactful to customers who would have their VPC change from 2 AZs to some other value.  So i've made this opt-in on the part of the user.  We can consider if that's the right default.

nodejs/awsx/utils.ts
```diff
@@ -103,3 +105,33 @@ export function hasTrueBooleanMember(obj: any, memberName: string | number | sym
...
+        },
+        err => {
+            error = err;
+            state = State.finishedWithError;
+        });
+
+    deasync.loopWhile(() => state === State.running);
+    if (state === State.finishedWithError) {
+        throw error;
+    }
```
CyrusNajmabadi: extremely simple async-blocking.  The promise above runs, and when complete sets result/error/state.  `loopWhile` then just pumps the node main loop as long as the condition inside keeps returning true.  Code for deasync is extremely trivial:

`loopWhile` is:

https://github.com/abbr/deasync/blob/90e24d8206b6d8d8ada9179ff00be5acf245c68a/index.js#L68-L73

And the actual `Run` function it calls into is:

https://github.com/abbr/deasync/blob/90e24d8206b6d8d8ada9179ff00be5acf245c68a/src/deasync.cc#L8-L13

CyrusNajmabadi: While this is risky to take this dependency on this native code module, it feels like this is so minimal and so targeted, that it should be an acceptable risk given the benefit we get.

Ideally, nodejs grpc could just expose sync calls.  But, in the absence of that, this gets us to that state at a safe enough level IMO.

CyrusNajmabadi:
Also tagging @ellismg .  If we take a native dep this could impact us moving forward to future nodejs versions.  As above, i think this risk is extremely tiny.  But we should consider and discuss if it's acceptable to us.

nodejs/awsx/utils.ts
```diff
@@ -103,3 +105,33 @@ export function hasTrueBooleanMember(obj: any, memberName: string | number | sym
...
+export function promiseResult<T>(promise: Promise<T>): T {
+    enum State {
+        running,
+        finishedSuccessfully,
+        finishedWithError,
+    }
+
+    let result: T;
+    let error = undefined;
+    let state = <State>State.running;
```
lukehoban: Curious - why is a cast needed?

nodejs/awsx/ec2/vpc.ts
lukehoban: FWIW - I actually think "2" is the right default here.  There is increased cost, and significantly diminishing returns, to using more than 2 AZs.

nodejs/awsx/utils.ts
CyrusNajmabadi: Note: the intent would be to move this down to the provider level.  So `pulumi/aws` would expose two methods `getAvailabilityZones(): Promise<GetAvailabilityZonesResult>` and `getAvailabilityZonesSync(): GetAvailabilityZonesResult`.  We'd keep the async versions around for compat (and because you should prefer them in cases where you can avoid the sync).  

I also thought about just making the existing calls return the non-promise result.  THat would be non-breaking for people who are calling `await` today.  But it would break people who explicit do things like `.then(...)`.

nodejs/awsx/utils.ts
CyrusNajmabadi: i'm baffled by this as well.  if i remove the cast, TS is now adamant that the local `state` variable has the `.running` value.  it then complains with the `if` check below stating that it's not ever going to be true.  Somehow it doesn't think the assignments in the Promise callbacks would change the flow value of `state`.  I honestly think it's buggy in the compiler/language and want to investigate more.

nodejs/awsx/ec2/vpc.ts
CyrusNajmabadi: ah ok.  Will stick with the this pattern then.

ellismg:
> If we take a native dep this could impact us moving forward to future nodejs versions. As above, i think this risk is extremely tiny. But we should consider and discuss if it's acceptable to us.

It looks like the library itself is using the NAPI stuff, which attempts to provide a stable ABI across Node versions, which is great for us.

What wasn't 100% obvious to me was if installing this thing actually requires having a c++ compiler on your machine.  I looked briefly at their package.json and it sure felt like they were going to invoke the compiler.  Do you know if this is the case?  It feels like requiring someone has a c++ compiler on their machine to be able to use awsx is not ideal.

swgillespie:
> What wasn't 100% obvious to me was if installing this thing actually requires having a c++ compiler on your machine. I looked briefly at their package.json and it sure felt like they were going to invoke the compiler.

This is likely to be the case, yes.

nodejs/awsx/utils.ts
CyrusNajmabadi: Talked to Daniel!  Unfortunately by design due to https://github.com/microsoft/TypeScript/issues/9998.

CyrusNajmabadi:
Hrmm... how can i tell?  I wiped node_modules, and then did the yarn step.  I didn't see anything crazy.  But it might have been so fast i didn't notice it (though i tend to think of C++ compile steps as uber slow).  Is there a good way for me to find out?

CyrusNajmabadi:
Ok.  trying verbose yarn.  Interesting stuff:

```
verbose 1.659 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-0.12".
verbose 1.659 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-0.10".
verbose 1.659 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-10".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-0.11".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-4".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-12".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-6".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-11".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-7".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-9".
verbose 1.66 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-0.11".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-8".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-0.12".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-0.10".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/darwin-x64-node-5".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-5".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-4".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-10".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-6".
verbose 1.661 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-8".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-0.10".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-9".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-0.12".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-11".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-0.11".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-10".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-ia32-node-7".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-4".
verbose 1.662 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-5".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-7".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-6".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-9".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-12".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-0.11".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-11".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-0.12".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-10".
verbose 1.663 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-4".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-0.10".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/linux-x64-node-8".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-12".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-7".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-6".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-8".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-0.10".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-ia32-node-5".
verbose 1.664 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-0.11".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-10".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-12".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-11".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-0.12".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-5".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-6".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-7".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-9".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-8".
verbose 1.665 Creating directory "/mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/node_modules/deasync/bin/win32-x64-node-4".
```

i looked through all the logs and don't see any c++ step explicitly.  Going to spelunk more, but i'm guessing they prebuild and keep with the package?

CyrusNajmabadi:
Downloaded the package that npm distributes from https://registry.npmjs.org/deasync/-/deasync-0.1.15.tgz

Unpacking shows they have deasync.node binaries for:

```
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-0.10
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-0.11
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-0.12
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-10
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-11
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-12
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-4
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-5
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-6
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-7
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-8
05/16/2019  12:46 PM    <DIR>          darwin-x64-node-9
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-0.10
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-0.11
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-0.12
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-10
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-4
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-5
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-6
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-7
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-8
05/16/2019  12:46 PM    <DIR>          linux-ia32-node-9
05/16/2019  12:46 PM    <DIR>          linux-x64-node-0.10
05/16/2019  12:46 PM    <DIR>          linux-x64-node-0.11
05/16/2019  12:46 PM    <DIR>          linux-x64-node-0.12
05/16/2019  12:46 PM    <DIR>          linux-x64-node-10
05/16/2019  12:46 PM    <DIR>          linux-x64-node-11
05/16/2019  12:46 PM    <DIR>          linux-x64-node-12
05/16/2019  12:46 PM    <DIR>          linux-x64-node-4
05/16/2019  12:46 PM    <DIR>          linux-x64-node-5
05/16/2019  12:46 PM    <DIR>          linux-x64-node-6
05/16/2019  12:46 PM    <DIR>          linux-x64-node-7
05/16/2019  12:46 PM    <DIR>          linux-x64-node-8
05/16/2019  12:46 PM    <DIR>          linux-x64-node-9
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-0.10
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-0.11
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-0.12
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-10
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-11
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-12
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-4
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-5
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-6
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-7
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-8
05/16/2019  12:46 PM    <DIR>          win32-ia32-node-9
05/16/2019  12:46 PM    <DIR>          win32-x64-node-0.10
05/16/2019  12:46 PM    <DIR>          win32-x64-node-0.11
05/16/2019  12:46 PM    <DIR>          win32-x64-node-0.12
05/16/2019  12:46 PM    <DIR>          win32-x64-node-10
05/16/2019  12:46 PM    <DIR>          win32-x64-node-11
05/16/2019  12:46 PM    <DIR>          win32-x64-node-12
05/16/2019  12:46 PM    <DIR>          win32-x64-node-4
05/16/2019  12:46 PM    <DIR>          win32-x64-node-5
05/16/2019  12:46 PM    <DIR>          win32-x64-node-6
05/16/2019  12:46 PM    <DIR>          win32-x64-node-7
05/16/2019  12:46 PM    <DIR>          win32-x64-node-8
05/16/2019  12:46 PM    <DIR>          win32-x64-node-9
```

So i think we *should* be good here as this covers osx, linux, windows.

ellismg:
> Unpacking shows they have deasync.node binaries for:

Ahh, slick.  Okay.  Seems reasonable to me as a thing to start experimenting with, but I'm not concerned anymore about this having a huge impact across our user base.

joeduffy:
Thank you for responding so quickly!

For `deasync`, I _do_ think this can cause issues for existing code, because it pushes a message pump where it didn't previously exist.

For example, some code like this:

```typescript
let k = 0;
setTimeout(function() { k++; }, 0);
new Vpc("vpc", { numberOfAvailabilityZones: "all" });
assert(k === 0);
```

previously could never fire the assert. Now, with `deasync`, it can.

That said, I'm not sure we care. I figured I'd mention it for completeness. I very well may be mistaken.

CyrusNajmabadi:
I think you are def correct in your analysis @joeduffy, and for completeness we should acknowledge this.   I'm also in the bucket of saying that we likely can accept this.  We can wait to hear both if this turns out to be a problem (unlikely IMO) for some customer, and if fixing this is difficult for them (also unlikely).  If we turn out to be incorrect, we can rollback.  But i think i'm comfortable here with the current level of risk to at least try this targetted change out.

Issue #291, opened 2019-06-01
LoadBalancers seem to default to external = true


joeduffy:
Unless I'm missing something, it appears that LoadBalancers default `external` to `true`, despite the documentation saying that they default to `false`.

If I create one simply as

```
new awsx.elasticloadbalancingv2.NetworkLoadBalancer("web-traffic");
```

and then do an update

```
new awsx.elasticloadbalancingv2.NetworkLoadBalancer("web-traffic", { external: true });
```

there is no change to the underlying resource configs.

If I instead then change it to

```
new awsx.elasticloadbalancingv2.NetworkLoadBalancer("web-traffic", { external: false });
```

a number of things get changed in the update.

joeduffy:
Also, a related, but slightly different question. Is it intentional that ApplicationLoadBalancers automatically open the right ports in the SecurityGroup

https://github.com/pulumi/pulumi-awsx/blob/4136dd6fbd0539b1069553a7147f0d4b92e33a4e/nodejs/awsx/elasticloadbalancingv2/application.ts#L206-L217

but NetworkLoadBalancers don't offer this same option?

CyrusNajmabadi:
> Unless I'm missing something, it appears that LoadBalancers default external to true, despite the documentation saying that they default to false.

Docs for LoadBalancerArgs say this for me:

```
    /**
     * Whether or not the load balancer is exposed to the internet. Defaults to `true` if
     * unspecified.
     */
    external?: boolean;
```

Unfortunately, looks like that didn't copy over properly into the ALBArgs or NetworkArgs.  I can fixup the docs there, but likely we can't change the actual default due to that breaking people who just want to upgrade.

> but NetworkLoadBalancers don't offer this same option?

My understanding is that NLBs don't have security groups.  See this comment on the forums:

https://forums.aws.amazon.com/thread.jspa?threadID=263245

> As far as I understand, the NLB itself doesn't have any security group. Instead, you control access using the security groups(s) attached to the EC2 instances. The source IP address is preserved, so you work with security group configuration (and other firewalls so to speak) as if the client had connected directly to your machine. The load balancer is kind of more transparent than in the ELB/ALB case. 

Note that this is likely an area we can try to excel at making easier.  It just hasn't been done yet :)

CyrusNajmabadi:
More docs on this here: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-register-targets.html

[...]

PR #292, opened 2019-06-01
Fix docs

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/291


joeduffy:
Great, thank you!

Also may be worth tweaking the code sample in README: https://github.com/pulumi/pulumi-awsx/issues/291#issuecomment-497988146

Issue #296, opened 2019-06-03
why does awsx.elasticloadbalancingv2.loadbalancer use as hash for the name?


keen99:
https://github.com/pulumi/pulumi-awsx/blob/master/nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts#L30-L38

when awsx is used to create a loadbalancer, the `name` is hashed into `shortName` and passed to aws.elasticloadbalancingv2.LoadBalancer as `name`.   

this leads to human friendly names like `371ba2bf-6d27008` and urls like `371ba2bf-6771c8d-1928391563.us-east-1.elb.amazonaws.com`

similar happens with targetGroup.

speculation would be that this is an attempt to force a url-friendly and elb-allowed name, or something on the tf side, since a pure string could exceed those limits..... but honestly I'd rather just see a failure than a gibberish URL.    This behavior basically means I can't use awsx resources in my environment.

I looked into doing a PR for this, but couldn't grok the local development environment (not to mention that I didnt look at the other language flavors) - maybe awsx needs a CONTRIBUTING.md.....      

For backwards compat, it seems like this should probably be an optional change, so something like `useNameForLbName = true` or something.   or perhaps just allow `shortName` to be passed as an override - `elbName` or something.

looking closer at aws.elb2, the autogen terraform wrapper suggests `name` is ending up at `namePrefix`.    when I pass name as name, I get `net-lb-dev-71c9ccf`  - suggesting the namearg  is ending up as namePrefix, and name is unset (hence a random suffix).   I'm not clear on the path of why that's happening though.

CyrusNajmabadi:
Working on fix.  Sorry about this.  it's an unfortunate holdover from some code we copied over from a previous version of the library.

CyrusNajmabadi:
> I looked into doing a PR for this, but couldn't grok the local development environment (not to mention that I didnt look at the other language flavors) - maybe awsx needs a CONTRIBUTING.md.....

I'll open a bug to help track this.  https://github.com/pulumi/pulumi/blob/master/CONTRIBUTING.md  is a starting point.  But individual repos need to point back to this, as well as add any particular specific info relevant to them.

> For backwards compat, it seems like this should probably be an optional change, so something like useNameForLbName = true or something. or perhaps just allow shortName to be passed as an override - elbName or something

[...]

PR #299, opened 2019-06-03
Allow users to specify the names used for LBs and TGs

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/296

When we ported code down from 'cloud' we unfortunately kept some code that unnecessarily hashes an LB's name to produce the resource name for the actual cloud.  This leads to gibberish in the AWS console for those names, which isn't great.

We can't change this behavior due to the impact it would have on existing stacks.  So, for now, we allow a person to workaround this by explicitly providing the name they'd like to use like so:  `new ApplicationLoadBalancer("PreferredName", { name: "PreferredName" })`.

### DIFF
--- a/nodejs/awsx/elasticloadbalancingv2/application.ts
+++ b/nodejs/awsx/elasticloadbalancingv2/application.ts
@@ -267,7 +267,7 @@ export interface ApplicationLoadBalancerArgs {
      * begin or end with a hyphen. If not specified, the [name] parameter passed into the
      * LoadBalancer constructor will be hashed and used as the name.
      */
-    name?: pulumi.Input<string>;
+    name?: string;
 
     /**
      * Whether or not the load balancer is exposed to the internet. Defaults to `true` if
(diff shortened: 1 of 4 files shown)


ellismg:
Does this mean that the sub-resources are also no longer autonamed with a unique suffix when passing a `name` property?  That feels right, but I just wanted to check.

It's a little interesting that when using this `name` trick, the URN for the `LoadBalancer` component still ends up having the hashed name.

Just for my own edification, do we think we would do something differently if we were willing to use aliases?  I'm guessing this still would not be sufficient because we used the hashed name also as the name of the physical resource, and we'd need a way to lock that in as well, which `aliases` is of no use for.

CyrusNajmabadi:
> Does this mean that the sub-resources are also no longer autonamed with a unique suffix when passing a name property? 

That's correct.  

> That feels right, but I just wanted to check.

Agreed.  

> It's a little interesting that when using this name trick, the URN for the LoadBalancer component still ends up having the hashed name.

Yes.  I suppose we could decide that *this* name is used and we don't hash at all.  I think i would actually prefer that tbh... 

> Just for my own edification, do we think we would do something differently if we were willing to use aliases? I'm guessing this still would not be sufficient because we used the hashed name also as the name of the physical resource, and we'd need a way to lock that in as well, which aliases is of no use for.

@lukehoban and i discussed this, but could not figure a way where this would work.  We didn't think too heavily on it though.  If you have an idea on how it might be expanded, woudl be happy to discuss more.

CyrusNajmabadi:
Ok.  Made the change such that we override the 'hash' name for the resource as well if you provide 'name' explicitly.

PR #311, opened 2019-06-10
Expand on the details an api integration can include to apigateway

CyrusNajmabadi:
The goal here is to be able to take @d-nishi's code and rewrite it simply as:

```ts
const ingestStream = new aws.kinesis.Stream("ingestAssets", {
 shardCount: 1,
 retentionPeriod: 72
});

new awsx.apigateway.Api("x", {
    routes: [{
        path: "/ingest",
        method: "POST",
        target: ingestStream,
    }],
});
```

we still need to imbue 'Stream' with the information to allow it to be passed in as a `target`.  Will link to that PR once done.


CyrusNajmabadi:
Tagging @lukehoban

ekrengel:
Are there tests for this?

Issue #304, opened 2019-06-08
Export CloudWatch Dashboard URLs


joeduffy:
It would be really nice if the `awsx.cloudwatch.Dashboard` class had a URL property that made it easy to export the URL of your dashboard after creating it.

Doing so manually is easy enough, but it's formulaic and I would imagine fairly common for folks to want easy access to.

```
const name = "myDashboard";
const dashboard = new awsx.cloudwatch.Dashboard(name, ...);
export const dashboardUrl =
    `https://${aws.config.region}.console.aws.amazon.com/cloudwatch/home?` +
    `region=${aws.config.region}#dashboards:name=${name}`;
```

It'd be nice if this was simply:

```
const dashboard = new awsx.cloudwatch.Dashboard("myDashboard", ...);
export const dashboardUrl = dashboard.url;
```

PR #318, opened 2019-06-11
Publish dashboard url

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/304

### DIFF
--- a/nodejs/awsx/cloudwatch/dashboard.ts
+++ b/nodejs/awsx/cloudwatch/dashboard.ts
@@ -78,6 +78,11 @@ export interface DashboardArgs {
  */
 export class Dashboard extends aws.cloudwatch.Dashboard {
     /**
+     * The url this [Dashboard] is published at.
+     */
+    public readonly url: pulumi.Output<string>;
+
+    /**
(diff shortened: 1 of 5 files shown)


nodejs/awsx/cloudwatch/dashboard.ts
```diff
@@ -87,6 +92,9 @@ export class Dashboard extends aws.cloudwatch.Dashboard {
             dashboardName: utils.ifUndefined(args.name, name),
             dashboardBody: getDashboardBody(args).apply(b => JSON.stringify(b)),
         }, opts);
+
+        this.url = pulumi.interpolate
+            `https://${aws.config.region}.console.aws.amazon.com/cloudwatch/home?region=${aws.config.region}#dashboards:name=${this.dashboardName}`;
```
ellismg: Feels like `aws.config.region` is not correct if this dashboard was created with a first class provider that was in a separate region.

CyrusNajmabadi: That's a good point.  Thanks!

PR #320, opened 2019-06-12
Make export internal.  It's not part of our public API.

CyrusNajmabadi:


CyrusNajmabadi:
Noticed as i was prepping for my demo.

Issue #323, opened 2019-06-12
subnets created by awsx.ec2.vpc do not get ipv6 netblocks assigned, no default routes created


keen99:
all subnets created are asigned ipv4 addresses, but not ipv6.  additionally,  created route tables are not given default ipv6 routes.  it might be worth mentioning that we create 1 route table per subnet, instead of 1 per vpc, which feels a little awkward.

sample created subnet:

```
        {
            "AvailabilityZone": "us-east-1b",
            "AvailabilityZoneId": "use1-az6",
            "AvailableIpAddressCount": 16378,
            "CidrBlock": "10.118.64.0/18",
            "DefaultForAz": false,
            "MapPublicIpOnLaunch": true,
            "State": "available",
            "SubnetId": "subnet-082c54f7d53c685f8",
            "VpcId": "vpc-06783ea793bd2cd76",
            "OwnerId": "568251075479",
            "AssignIpv6AddressOnCreation": false,
            "Ipv6CidrBlockAssociationSet": [],
            "Tags": [
                {
                    "Key": "type",
                    "Value": "public"
                },
                {
                    "Key": "Name",
                    "Value": "custom-public-1"
                }
            ],
            "SubnetArn": "arn:aws:ec2:us-east-1:568251075479:subnet/subnet-082c54f7d53c685f8"
        }
```

same route table:
```
       {
            "Associations": [
                {
                    "Main": false,
                    "RouteTableAssociationId": "rtbassoc-063b6c7d942cc30c5",
                    "RouteTableId": "rtb-03083e81772ca34dc",
                    "SubnetId": "subnet-0ce946c7c70d9b413"
                }
            ],
            "PropagatingVgws": [],
            "RouteTableId": "rtb-03083e81772ca34dc",
            "Routes": [
                {
                    "DestinationCidrBlock": "10.118.0.0/16",
                    "GatewayId": "local",
                    "Origin": "CreateRouteTable",
                    "State": "active"
                },
                {
                    "DestinationCidrBlock": "0.0.0.0/0",
                    "GatewayId": "igw-02042b572048e807a",
                    "Origin": "CreateRoute",
                    "State": "active"
                },
                {
                    "DestinationIpv6CidrBlock": "2600:1f18:4568:8700::/56",
                    "GatewayId": "local",
                    "Origin": "CreateRouteTable",
                    "State": "active"
                }
            ],
            "Tags": [],
            "VpcId": "vpc-06783ea793bd2cd76",
[...]

PR #329, opened 2019-06-14
Enable ipv6 for Vpc subnets.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/323

An `awsx.ec2.Vpc` with `assignGeneratedIpv6CidrBlock: true` will now set  `assignIpv6AddressOnCreation: true` by default for child subnets.  This can be overridden by  setting that value explicitly to `false` with the subnet's args.

This changes our behavior and falls under:
https://github.com/pulumi/home/wiki/Compatibility-philosophy-and-guidelines: 

> Changes to existing resources. We may decide at times to change certain defaults for a resource. Or to add information into a resource (for example 'tags'). Because this has the potential to change runtime behavior, we should let customers know and provide them guidance on what to do if they don't want their resource to change (for example, by manually providing the old value).

> If it is possible for the user to keep the resource from not changing through their own actions, then this can be released in a 'patch' release.

Because this behavior can be overridden, this is ok for a patch release.   Changelog has been updated to provide the appropriate informaiton on this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## 0.18.6 (Unreleased)
 
 - awsx.ecs.Cluster can be created from an existing aws.ecs.Cluster's id.
+- An `awsx.ec2.Vpc` with `assignGeneratedIpv6CidrBlock: true` will now set
+  `assignIpv6AddressOnCreation: true` by default for child subnets.  This can be overridden by
+  setting that value explicitly to `false` with the subnet's args.
 
 ## 0.18.5 (6/12/2019)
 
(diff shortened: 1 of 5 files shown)


CyrusNajmabadi:
Tagging @ellismg and @lukehoban .  Specifically of interest is that we can apply our compat guidelines here.  Tagging both of you just for eyes/awareness and to make sure we're all ok here.  As we get more comfortable with this moving forward, we can stat to make this call with a smaller group of people.

CyrusNajmabadi:
Note: i'm not actually certain this PR is complete.  have to figure out the cidr story.

CyrusNajmabadi:
I feel comfortable with this.  If anyone has any concerns, let me know!

jen20:
We should probably also support an egress only internet gateway here too?

Issue #294, opened 2019-06-02
Consider adding a LoadBalancer.attach function


joeduffy:
I wanted to write a sample that uses our new ELB support for a simple EC2 VM instance. I eventually got this working, which led me to wonder if we could make this nicer.

The complete example is:

```typescript
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";

// Create a security group to let traffic flow.
const sg = new awsx.ec2.SecurityGroup("web-sg", {
    vpc: awsx.ec2.Vpc.getDefault(),
    egress: [{ protocol: "-1", fromPort: 0, toPort: 0, cidrBlocks: [ "0.0.0.0/0" ] }],
});

// Creates an ALB associated with the default VPC for this region and listen on port 80.
const alb = new awsx.elasticloadbalancingv2.ApplicationLoadBalancer("web-traffic",
    { external: true, securityGroups: [ sg ] });
const listener = alb.createListener("web-listener", { port: 80 });

// For each subnet, and each subnet/zone, create a VM and a listener.
export const publicIps: any[] = [];
for (let i = 0; i < alb.vpc.publicSubnets.length; i++) {
    const vm = new aws.ec2.Instance(`web-${i}`, {
        ami: aws.getAmi({
            filters: [
                { name: "name", values: [ "ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*" ] },
                { name: "virtualization-type", values: [ "hvm" ] },
            ],
            mostRecent: true,
            owners: [ "099720109477" ], // Canonical
        }).then(ami => ami.id),
        instanceType: "t2.micro",
        subnetId: alb.vpc.publicSubnets[i].subnet.id,
        availabilityZone: alb.vpc.publicSubnets[i].subnet.availabilityZone,
        vpcSecurityGroupIds: [ sg.id ],
        userData: `#!/bin/bash
    echo "Hello World, from Server ${i+1}!" > index.html
    nohup python -m SimpleHTTPServer 80 &`,
    });
    publicIps.push(vm.publicIp);

    const attach = new aws.elasticloadbalancingv2.TargetGroupAttachment(`web-nlb-vm-${i}`, {
        targetId: vm.privateIp,
        targetGroupArn: alb.targetGroups[0].targetGroup.arn,
        availabilityZone: vm.availabilityZone,
    });
}

// Export the resulting URL so that it's easy to access.
export const endpoint = listener.endpoint;
```

[...]

PR #330, opened 2019-06-14
Add the ability to easily attach targets to an ALB/NLB

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/294

### DIFF
--- a/nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
+++ b/nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
@@ -15,6 +15,7 @@
 import * as aws from "@pulumi/aws";
 import * as pulumi from "@pulumi/pulumi";
 
+import * as mod from ".";
 import * as x from "..";
 import * as utils from "./../utils";
 
@@ -22,6 +23,9 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
     public readonly loadBalancer: aws.elasticloadbalancingv2.LoadBalancer;
     public readonly vpc: x.ec2.Vpc;
     public readonly securityGroups: x.ec2.SecurityGroup[];
+
+    public readonly listeners: mod.Listener[] = [];
+    public readonly targetGroups: mod.TargetGroup[] = [];
 
     constructor(type: string, name: string, args: LoadBalancerArgs, opts?: pulumi.ComponentResourceOptions) {
         super(type, name, {}, opts);
@@ -42,7 +46,18 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
             securityGroups: this.securityGroups.map(g => g.id),
             tags: utils.mergeTags(args.tags, { Name: longName }),
         }, parentOpts);
-   }
+    }
+
+    public attachTarget(
+            name: string,
+            args: pulumi.Input<LoadBalancerTarget> | LoadBalancerTargetProvider | aws.ec2.Instance,
+            opts: pulumi.CustomResourceOptions = {}) {
+        if (this.listeners.length === 0) {
+            throw new pulumi.ResourceError("Load balancer must have at least one [Listener] in order to attach a target.", this);
+        }
+
+        return this.listeners[0].attachTarget(name, args, opts);
+    }
 }
 
 function getSubnets(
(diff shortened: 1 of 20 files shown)


nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
```diff
@@ -42,7 +46,18 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
             securityGroups: this.securityGroups.map(g => g.id),
             tags: utils.mergeTags(args.tags, { Name: longName }),
         }, parentOpts);
-   }
+    }
+
+    public attachTarget(
+            name: string,
+            args: pulumi.Input<LoadBalancerTarget> | LoadBalancerTargetProvider | aws.ec2.Instance,
```
CyrusNajmabadi: note @lukehoban I'm hardcoding in aws.ec2.Instance here (but i don't like it).   The goal is to easily allow clients to just call `lb.attachTarget(ec2Instance)`.  However, that won't work because ec2Instance doesn't implement LoadBalancerTargetProvider.  I *could* make it implement that interface by using mixins in the aws package.  But that would be a first where we do that.  

I'm curious what you think here.  

1. Should we just recognize this super common type, like i've done here?
2. Should we require the user to call `new awsx.elasticloadbalancingv2.Ec2InstanceTarget` themselves from code?
3. Should I go monkeypatch tihs interface into `aws`?

Something else?

CyrusNajmabadi:
Tagging @lukehoban @ellismg .  Note that there's an open question i have as to the best way to accomplish connecting ec2 and elbv2.

nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
```diff
@@ -138,3 +153,52 @@ export interface LoadBalancerSubnets {
...
+}
+
+/**
+ * Allows an EC2 instance to simply be used as the target of an ALB or NLB.  To use, just call:
+ *
+ * ```ts
+ *  lb.attachTarget(new Ec2InstanceTarget(instance));
+ * ```
+ */
+export class Ec2InstanceTarget implements LoadBalancerTargetProvider {
```
CyrusNajmabadi: helper class to turn an aws.ec2.Instance into a LoadBalancerTargetProvider.  Won't be necessary if we can just monkeypatch aws.ec2.Instance to expose the `loadBalancerTarget` method.

nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
```diff
@@ -138,3 +153,52 @@ export interface LoadBalancerSubnets {
...
+    constructor(public readonly instance: aws.ec2.Instance) {
+    }
+
+    public loadBalancerTarget(targetType: pulumi.Input<mod.TargetType>): pulumi.Output<LoadBalancerTarget> {
+        const result = pulumi.output([targetType, this.instance.id, this.instance.privateIp, this.instance.availabilityZone])
+                             .apply(([targetType, instanceId, privateIp, availabilityZone]) => {
+            return {
+                targetId: targetType === "instance" ? instanceId : privateIp,
+                availabilityZone: availabilityZone,
```
CyrusNajmabadi: @joeduffy was looking but couldn't find good docs here.  Do you know of any?  It will help me determine if this behavior is the one we want for all cases, or if we need to expose some level of customization to control things more.

nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
lukehoban: I personally think this is okay as is.  I think in practice it will be okay to handle the universe of things that can be load balancer targets in one place here.

nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
lukehoban: Good docs on what exactly?

nodejs/awsx/elasticloadbalancingv2/targetGroup.ts
```diff
@@ -212,11 +233,24 @@ export interface TargetGroupArgs {
...
+ * The type of target that you must specify when registering targets with a target group. The
+ * possible values are `instance` (targets are specified by instance ID) or `ip` (targets are
+ * specified by IP address) or `lambda` (targets are specified by lambda arn). The default is
+ * `ip`. Note that you can't specify targets for a target group using both instance IDs and IP
+ * addresses. If the target type is `ip`, specify IP addresses from the subnets of the virtual
+ * private cloud (VPC) for the target group, the RFC 1918 range (10.0.0.0/8, 172.16.0.0/12, and
+ * 192.168.0.0/16), and the RFC 6598 range (100.64.0.0/10). You can't specify publicly routable IP
+ * addresses.
+ */
+export type TargetType = "instance" | "ip" | "lambda";
```
lukehoban: It doesn't look like we are handling the `"lambda"` case yet?

nodejs/examples/alb/ec2Instance/index.ts
```diff
@@ -0,0 +1,56 @@
...
+        subnetId: alb.vpc.publicSubnets[i].subnet.id,
+        availabilityZone: alb.vpc.publicSubnets[i].subnet.availabilityZone,
+        vpcSecurityGroupIds: [ sg.id ],
+        userData: `#!/bin/bash
+echo "Hello World, from Server ${i+1}!" > index.html
+nohup python -m SimpleHTTPServer 80 &`,
+    });
+    publicIps.push(vm.publicIp);
+
+    alb.attachTarget("target-" + i, vm);
```
lukehoban: I have a bit of trouble following the code above - but how does this target get connected to a listener?

It would be interesting to recreate something like the graph at https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html which includes all the main permutations of relationships between these as a test.

nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
joeduffy: I think https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#target-type is the best. Note this part:

> When the target type is ip, you can specify IP addresses from one of the following CIDR blocks:
> 
> The subnets of the VPC for the target group
> 
> * 10.0.0.0/8 (RFC 1918)
> * 100.64.0.0/10 (RFC 6598)
> * 172.16.0.0/12 (RFC 1918)
> * 192.168.0.0/16 (RFC 1918)
> 
> **Important**
> You can't specify publicly routable IP addresses.

CyrusNajmabadi: yes, that's very helpful.  let me read through this to figure out what else may need to change.  thanks!

nodejs/examples/alb/ec2Instance/index.ts
CyrusNajmabadi: will document, but we take the approach of defaulting to the first listener for the alb.  You can also just specify the listener you want to attach to, and it will pick the default target group for that listener.  finally, you can just specify the exact target group you want to add to.

nodejs/awsx/elasticloadbalancingv2/loadBalancer.ts
CyrusNajmabadi: ok, sounds good.

CyrusNajmabadi:
Lambda support has been added.

nodejs/awsx/elasticloadbalancingv2/targetGroup.ts
CyrusNajmabadi: We are now!

Issue #314, opened 2019-06-11
Explicit provider not being used for awsx.ec2.Vpc


clstokes:
Setting an explicit provider does not work for `awsx.ec2.Vpc` and potentially child resources. The ambient provider is still used.

```
$ pulumi config set aws:region does-not-exist
```

```
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";

const east = new aws.Provider("east", { region: "us-east-1" });

const vpc = new awsx.ec2.Vpc("east", {}, {
    providers: {
        "aws": east,
    }
})
```

```
07:49:27 [pulumi-awsx-region-provider] $ pulumi up
Previewing update (dev):

     Type                     Name                             Plan       Info
 +   pulumi:pulumi:Stack      pulumi-awsx-region-provider-dev  create     24 messages
 +   ├─ awsx:x:ec2:Vpc        east                             create
 +   ├─ pulumi:providers:aws  east                             create
     └─ aws:ec2:Vpc           east                                        1 error

Diagnostics:
  pulumi:pulumi:Stack (pulumi-awsx-region-provider-dev):
    Error: invocation of aws:index/getAvailabilityZones:getAvailabilityZones returned an error: Invalid AWS Region: does-not-exist
        at /Users/clstokes/cc/scratch/pulumi-awsx-region-provider/node_modules/@pulumi/pulumi/runtime/invoke.js:74:33
        at Object.onReceiveStatus (/Users/clstokes/cc/scratch/pulumi-awsx-region-provider/node_modules/grpc/src/client_interceptors.js:1205:9)
        at InterceptingListener._callNext (/Users/clstokes/cc/scratch/pulumi-awsx-region-provider/node_modules/grpc/src/client_interceptors.js:
...
```

CyrusNajmabadi:
Easy to fix, but easy bug to make.  I'm going to bring this up at our backcompat discussion tomorrow.  If we're willing to break things, we might want to put on the table the idea of just outright moving to a model where "parent" is required.  It's basically a footgun for us that keeps hitting us over and over again.

lukehoban:
For testing this - I’d suggest we make all tests use 1st class providers and set an invalid `aws:region` for the default provider. That will ensure that we see immediate failures if providers are not being passed through correctly.

CyrusNajmabadi:
[...]

PR #332, opened 2019-06-18
Properly parent Vpc resources.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/314

### DIFF
--- a/nodejs/awsx/aws.ts
+++ b/nodejs/awsx/aws.ts
@@ -15,20 +15,27 @@
 // TODO[pulumi/pulumi-aws#40]: Move these to pulumi-aws.
 
 import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
 
 // Compute the availability zones only once, and store the resulting promise.
 let zones: Promise<string[]> | undefined;
 
 // Export as a function instead of a variable so clients can pass one AZ as a promise to a resource.
-export async function getAvailabilityZone(index: number): Promise<string> {
-    const azs = await getAvailabilityZones();
+export async function getAvailabilityZone(index: number, opts?: pulumi.InvokeOptions): Promise<string> {
+    const azs = await getAvailabilityZones(opts);
     return azs[index];
 }
 
-export function getAvailabilityZones(): Promise<string[]> {
-    if (!zones) {
-        zones = aws.getAvailabilityZones().then(r => r.names);
+export function getAvailabilityZones(opts?: pulumi.InvokeOptions): Promise<string[]> {
+    if (opts === undefined) {
+        // Only cache the results if 'opts' is not passed in.  If there are opts, it may change the
+        // results and we can't necessarily reuse any previous results.
+        if (!zones) {
+            zones = aws.getAvailabilityZones(/*args:*/ undefined).then(r => r.names);
+        }
+
+        return zones;
     }
 
-    return zones;
+    return aws.getAvailabilityZones(/*args:*/ undefined, opts).then(r => r.names);
 }
--- a/nodejs/awsx/aws_test.go
+++ b/nodejs/awsx/aws_test.go
@@ -39,11 +39,11 @@ import (
 )
 
 func Test_Examples(t *testing.T) {
-	region := os.Getenv("AWS_REGION")
-	if region == "" {
+	envRegion := os.Getenv("AWS_REGION")
+	if envRegion == "" {
 		t.Skipf("Skipping test due to missing AWS_REGION environment variable")
 	}
-	fmt.Printf("AWS Region: %v\n", region)
+	fmt.Printf("AWS Region: %v\n", envRegion)
 
 	cwd, err := os.Getwd()
 	if !assert.NoError(t, err, "expected a valid working directory: %v", err) {
@@ -52,7 +52,8 @@ func Test_Examples(t *testing.T) {
 
 	testBase := integration.ProgramTestOptions{
 		Config: map[string]string{
-			"aws:region": region,
+			"aws:region":    "INVALID_REGION",
+			"aws:envRegion": envRegion,
 		},
 		Dependencies: []string{
 			"@pulumi/awsx",
@@ -63,7 +64,8 @@ func Test_Examples(t *testing.T) {
 
 	shortTests := []integration.ProgramTestOptions{
 		testBase.With(integration.ProgramTestOptions{
-			Dir: path.Join(cwd, "../examples/cluster"),
+			Config: map[string]string{"aws:region": envRegion},
+			Dir:    path.Join(cwd, "../examples/cluster"),
 		}),
 		testBase.With(integration.ProgramTestOptions{
 			Dir: path.Join(cwd, "../examples/dashboards"),
@@ -81,13 +83,10 @@ func Test_Examples(t *testing.T) {
 		testBase.With(integration.ProgramTestOptions{
 			Dir:                    path.Join(cwd, "../examples/nlb/fargateShort"),
 			StackName:              addRandomSuffix("fargate"),
-			ExtraRuntimeValidation: containersRuntimeValidator(region, true /*isFargate*/, true /*short*/),
+			ExtraRuntimeValidation: containersRuntimeValidator(envRegion, true /*isFargate*/, true /*short*/),
 		}),
 		testBase.With(integration.ProgramTestOptions{
 			Dir: path.Join(cwd, "../examples/api"),
-			Config: map[string]string{
-				"aws:region": region,
-			},
 			Dependencies: []string{
 				"@pulumi/awsx",
 			},
@@ -211,13 +210,14 @@ func Test_Examples(t *testing.T) {
 			Dir:       path.Join(cwd, "../examples/nlb/fargate"),
 			StackName: addRandomSuffix("fargate"),
 			Config: map[string]string{
-				"aws:region":               region,
+				"aws:region":               "INVALID_REGION",
+				"aws:envRegion":            envRegion,
 				"containers:redisPassword": "SECRETPASSWORD",
 			},
 			PreviewCommandlineFlags: []string{
 				"--diff",
 			},
-			ExtraRuntimeValidation: containersRuntimeValidator(region, true /*isFargate*/, false /*short*/),
+			ExtraRuntimeValidation: containersRuntimeValidator(envRegion, true /*isFargate*/, false /*short*/),
 		}),
 
 		// {
--- a/nodejs/awsx/aws_test.go
+++ b/nodejs/awsx/aws_test.go
@@ -63,10 +63,15 @@ func Test_Examples(t *testing.T) {
 	}
 
 	shortTests := []integration.ProgramTestOptions{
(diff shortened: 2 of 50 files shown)


CyrusNajmabadi:
Discovered another issue when talking to luke.  Not ready for review just yet.

CyrusNajmabadi:
@lukehoban this is ready for review.

nodejs/awsx/aws.ts
```diff
@@ -15,20 +15,27 @@
...
-        zones = aws.getAvailabilityZones().then(r => r.names);
+export function getAvailabilityZones(opts?: pulumi.InvokeOptions): Promise<string[]> {
+    if (opts === undefined) {
+        // Only cache the results if 'opts' is not passed in.  If there are opts, it may change the
+        // results and we can't necessarily reuse any previous results.
+        if (!zones) {
+            zones = aws.getAvailabilityZones(/*args:*/ undefined).then(r => r.names);
+        }
+
+        return zones;
```
CyrusNajmabadi: note: i would also be fine with no caching at all.

nodejs/awsx/aws_test.go
```diff
@@ -62,9 +63,15 @@ func Test_Examples(t *testing.T) {
...
+		integration.ProgramTestOptions{
+			Config: map[string]string{"aws:region": envRegion},
+			Dir:    path.Join(cwd, "../examples/cluster"),
+			Dependencies: []string{
+				"@pulumi/awsx",
+			},
+			Quick:       true,
+			SkipRefresh: true,
+		},
 		testBase.With(integration.ProgramTestOptions{
```
lukehoban: The mixing of tests based on `testBase` with others seems like it could be fragile in the future.  Should we rename or restructure a little to make it clearer what tests are testing what (and how new tests should be added)?

nodejs/awsx/elasticloadbalancingv2/application.ts
```diff
@@ -122,12 +122,15 @@ export class ApplicationTargetGroup extends mod.TargetGroup {
     }
 
     public createListener(name: string, args: ApplicationListenerArgs,
-                          opts?: pulumi.ComponentResourceOptions): ApplicationListener {
+                          opts: pulumi.ComponentResourceOptions = {}): ApplicationListener {
+        // We didn't use to parent the listener to the target group.  Now we do.  Create an alias
+        // from the old parent to the current one if this moves over.
+        const aliases = [pulumi.createUrn(name, "awsx:x:elasticloadbalancingv2:ApplicationListener", opts.parent)];
```
lukehoban: Technically - I think we need to merge this new alias into any aliases already provided in `opts`?  If the user provides any aliases, the code below will overwrite `aliases` with their array instead of this array - right?

nodejs/awsx/elasticloadbalancingv2/listenerRule.ts
```diff
@@ -33,9 +33,11 @@ export class ListenerRule extends pulumi.ComponentResource {
 
     constructor(name: string, listener: x.elasticloadbalancingv2.Listener,
                 args: ListenerRuleArgs, opts: pulumi.ComponentResourceOptions = {}) {
-        super("awsx:x:elasticloadbalancingv2", name, {}, { parent: listener, ...opts });
+        // We forgot to add the `ListenerRule` part of the name.  Add it in and create an alias from
+        // the previous incorrect name.
+        const aliases = [pulumi.createUrn(name, "awsx:x:elasticloadbalancingv2", opts.parent)];
```
lukehoban: I think this can (and should to be correct here - since the wrong parent happens to being passed here) be:

```ts
const aliases = [{ type: "awsx:x:elasticloadbalancingv2"})];
```

nodejs/awsx/elasticloadbalancingv2/listenerRule.ts
```diff
@@ -33,9 +33,11 @@ export class ListenerRule extends pulumi.ComponentResource {
...
+        super("awsx:x:elasticloadbalancingv2:ListenerRule", name, {}, { parent: listener, aliases, ...opts });
```
lukehoban: This `aliases, ...opts` pattern is repeated quite a bit, and unfortunately I don't think it's quite safe.  It's unfortunate that there's not an easier way to express this - but I think for now we need to be a little more careful in merging these.

nodejs/awsx/elasticloadbalancingv2/network.ts
```diff
@@ -144,7 +144,7 @@ export class NetworkListener
             protocol,
             loadBalancer,
             defaultActions,
-        }, { parent: loadBalancer, ...opts });
+        }, opts);
```
lukehoban: Is there a reason we are changing this?

lukehoban: In particular, it looks like we're making the opposite change to what we are doing elsewhere - like in `targetGroup`.

nodejs/awsx/utils.ts
```diff
@@ -108,6 +108,26 @@ export function hasTrueBooleanMember(obj: any, memberName: string | number | sym
...
+    return getRegionFromProvider(opts.provider);
+}
+
+
+/** @internal */
+export function getRegion(res: pulumi.Resource): pulumi.Output<aws.Region> {
+    return getRegionFromProvider(res.getProvider("aws::"));
+}
+
+function getRegionFromProvider(provider: pulumi.ProviderResource | undefined) {
```
lukehoban: Is this also intended to be `@internal`?

nodejs/awsx/utils.ts
```diff
@@ -108,6 +108,26 @@ export function hasTrueBooleanMember(obj: any, memberName: string | number | sym
...
+        return getRegion(opts.parent);
+    }
+
+    return getRegionFromProvider(opts.provider);
+}
+
+
+/** @internal */
+export function getRegion(res: pulumi.Resource): pulumi.Output<aws.Region> {
+    return getRegionFromProvider(res.getProvider("aws::"));
```
lukehoban: Perhaps a comment explaining what this is doing? :-) . I think it is passing a "fake" type token that will convince `getProvider` to lookup the AWS provider?

nodejs/awsx/utils.ts
```diff
@@ -108,6 +108,26 @@ export function hasTrueBooleanMember(obj: any, memberName: string | number | sym
...
+}
+
+function getRegionFromProvider(provider: pulumi.ProviderResource | undefined) {
+    const region = provider ? (<any>provider).region : undefined;
+    return region || aws.config.region;
```
lukehoban: Are we guaranteed that `.region` will be populated on a provider?  If not, is this correct?  If so, can't we be more precise here and return `provider ? provider.region : aws.config.region`?  Also, why do we ever not get back a `Provider`?  Is that just a deign hole in the SDK where we do not materialize a `Provider` instance for the default provider?

cc @pgavlin

nodejs/examples/alb/lambdaTarget/index.ts
```diff
@@ -16,15 +16,18 @@ import * as pulumi from "@pulumi/pulumi";
 import * as aws from "@pulumi/aws";
 import * as awsx from "@pulumi/awsx";
 
+const config = new pulumi.Config("aws");
```
lukehoban: Perhaps here and elsewhere a comment to explain why we are doing this?

nodejs/examples/ec2/app/Dockerfile
```diff
@@ -1,2 +1,5 @@
+FROM alpine AS build
+COPY content/index.html /
+
 FROM nginx
-COPY content /usr/share/nginx/html
\ No newline at end of file
+COPY --from=build /index.html /usr/share/nginx/html
```
lukehoban: Why are we changing this?  The original seems strictly better.

nodejs/examples/ecr/app/Dockerfile
```diff
@@ -1,2 +1,5 @@
+FROM alpine AS build
+COPY content/index.html /
+
 FROM nginx
-COPY content /usr/share/nginx/html
\ No newline at end of file
+COPY --from=build /index.html /usr/share/nginx/html
```
lukehoban: Similar comment here.

lukehoban:
> Fixes #314

Would be great to augment the description of this PR to include details on all the categories of changes being made here.  There are enough things here that could potentially cause issues that it will be useful to understand what we were trying to accomplish here.

nodejs/awsx/elasticloadbalancingv2/listenerRule.ts
CyrusNajmabadi: you're right.  will fix.

nodejs/awsx/elasticloadbalancingv2/application.ts
CyrusNajmabadi: You're correct.  Have updated accordingly.

nodejs/awsx/elasticloadbalancingv2/listenerRule.ts
CyrusNajmabadi: You're correct.  Have switched to merging.

nodejs/awsx/elasticloadbalancingv2/network.ts
CyrusNajmabadi: Sorry, i can def see how that's unclear.  In both cases what i'm doing is trying to move this logic to a centralized location in the base class.  We were inconsistent with how we did this, and now i'm trying to do it in the same way across AppListener/NetworkListener/AppTargetGroup/NetworkTargetGroup.

nodejs/awsx/utils.ts
CyrusNajmabadi: it's not exported, so i didn't think we needed to mark it one way or hte other.

nodejs/awsx/utils.ts
CyrusNajmabadi: sure!

nodejs/awsx/utils.ts
CyrusNajmabadi: >Is that just a deign hole in the SDK where we do not materialize a Provider instance for the default provider?

Correct.  If there's no provider passed along, then trying to get the provider simply returns nothing back.  Later, in the engine, the right provider is found and used in that case.  But that doesn't help on the sdk side afaict.

> Are we guaranteed that .region will be populated on a provider?   I'm not certain we have that guarantee at all...

nodejs/examples/ec2/app/Dockerfile
CyrusNajmabadi: unintentional.  will revert.

CyrusNajmabadi:
> Would  be great to augment the description of this PR to include details on  all the categories of changes being made here.  There are enough things  here that could potentially cause issues that it will be useful to  understand what we were trying to accomplish here.

Will do.

CyrusNajmabadi:
@lukehoban Moved to the `{ parent: ... }` alias form.  Definitely a lot easier to follow, esp when comparing the before/after for any piece of code.  Thanks!

PR #337, opened 2019-06-19
Deprecate the original awsx 'Cluster' and 'Network' types

CyrusNajmabadi:


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
   `assignIpv6AddressOnCreation: true` by default for child subnets.  This can be overridden by
   setting that value explicitly to `false` with the subnet's args.
 
+- `awsx.Cluster` and `awsx.Network` are now deprecated and will no longer receive future changes.
```
ellismg: Is it possible for us to provide some guidance either in the `CHANGELOG.md` or have a link back to an issue with some guidance on how to migrate to these new types?  I assume migrating from `awsx.Network` to `aws.ec2.Vpc` can be a little tricky, especially if you don't want to force recreation of the VPC?

PR #342, opened 2019-06-21
Allow for arbitrary launch configs to be provided for awsx.autoscaling.AutoScalingGroup, not just a specific ec2 launch config.

CyrusNajmabadi:
WIP. Looking for feedback/thoughts.

We've unfortunately hardcoded in ec2 specific launch config capabilities. We could deprecate that and try to make thigns more general. But this approach here also gives an escape hatch for customers that want to supply their own launch configs.

### DIFF
--- a/nodejs/awsx/autoscaling/autoscaling.ts
+++ b/nodejs/awsx/autoscaling/autoscaling.ts
@@ -14,6 +14,7 @@
 
 import * as aws from "@pulumi/aws";
 import * as pulumi from "@pulumi/pulumi";
+import * as random from "@pulumi/random";
 
 import * as x from "..";
 import * as utils from "./../utils";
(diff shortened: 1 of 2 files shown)


nodejs/awsx/autoscaling/autoscaling.ts
```diff
@@ -60,24 +62,37 @@ export class AutoScalingGroup extends pulumi.ComponentResource {
...
         // Use the autoscaling config provided, otherwise just create a default one for this cluster.
-        if (args.launchConfiguration) {
-            this.launchConfiguration = args.launchConfiguration;
+        if (args.launchConfigurationId) {
+            launchConfigurationId = args.launchConfigurationId;
+            stackName = new random.RandomId(name, {
+                byteLength: 8,
+                prefix: name,
+                keepers: { launchConfigurationId },
+            }, { parent: this }).b64Url;
```
CyrusNajmabadi: not sure if this is the right way to do things. But currently we use a random bucket id to get a unique stack name. i figured we should likely switch to actually pulumi.Random for something cleaner.

nodejs/awsx/autoscaling/autoscaling.ts
```diff
@@ -60,24 +62,37 @@ export class AutoScalingGroup extends pulumi.ComponentResource {
...
 
         // Use cloudformation to actually construct the autoscaling group.
         this.stack = new aws.cloudformation.Stack(name, {
             ...args,
-            name: this.launchConfiguration.stackName,
+            name: stackName,
             templateBody: getCloudFormationTemplate(
                 name,
-                this.launchConfiguration.id,
+                launchConfigurationId,
```
CyrusNajmabadi: open question if it's sufficient to just provide a stack-name/launch-config-id. Or if there might be other information necessary to pass in by customers to get the experience they need.

PR #347, opened 2019-06-22
Properly parent component resources.

CyrusNajmabadi:
Many awsx components were both not parented properly and also did not correctly pass 'provider'
  information along.  For programs not explicitly passing a 'provider' along, this normally was not
  an issue.  However, programs that did want to use an explicit 'provider' (for example, to set a
  particular region for a resource), would commonly run into issues.  We tried to broadly fix these
  issues, while doing so in a way that should hopefully **not** have any impact on any existing
  stacks. Specifically, the 'aliases' feature was used so that while we may have reparented some
  resources, pulumi will know that that is just a representation change, and shouldn't cause any
  actual resources to be created/deleted.  If you see otherwise, please let us know.

  The specific resources/apis affected are:

  1. Lambda authorizers will now be parented by the `awsx.apigateway.Api` they are created for.
  2. `awsx.apigateway.Api` static routes will properly work when using a different provider
     (previously it wouldn't place the static route in the corresponding region for the provider).
  3. `awsx.apigateway.Api` UsagePlans/UsagePlanKeys now properly pass along a provider.

  4. `awsx.autoscaling.AutoScalingGroup` now properly passes along a provider.
  5. The `LaunchConfiguration` for an `AutoScalingGroup` will now be parented by the `AutoScalingGroup`

  6. `awsx.cloudwatch.Dashboard` now properly passes along a provider.

  7. The underlying `aws.ec2.Vpc` is now parented to the `awsx.ec2.Vpc` that created it.
  8. `awsx.ec2.Vpc.getDefault` has changed behavior.  It now takes in options that allow a provider
     to be passed in.  This provider is used to determine which region to lookup the default Vpc in.
     Note: when this function is called multiple times it will return the same Vpc instance if for
     the same region.
  9. `awsx.ec2.Vpc.getDefault` will now return Vpcs with the name `default-<actual vpc id>`.  An
     existing default vpc with the name `default-vpc` will be aliased to this new name.
  10. Subnets created by `awsx.ec2.Vpc` will be parented to the Vpc now.
  11. The InternetGateway and NatGateways created by `Vpc.fromExistingIds` will now be parented to
      the Vpc.

  12. `awsx.ecr.LifeCyclePolicy` is now parented by the `aws.ecr.Repository` it is created for.

  13. `awsx.ecs.Service/TaskDefinition` now respect providers.

  14. `awsx.elasticloadbalancingv2.ApplicationListener/ApplicationTargetGroup/NetworkListener/NetworkTargetGroup`
      will now all be parented by their respective `LoadBalancer` by default if a parent is not
      specified.
  15. The type name of an `awsx.elasticloadbalancingv2.ListenerRule` has been fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
     - sudo apt-get -y install docker-ce=17.09.0~ce-0~ubuntu
     - docker version
     # Install Pulumi
-    - curl -sSL https://get.pulumi.com | bash -s -- --version "$(curl -sSL http://registry.npmjs.org/-/package/@pulumi/pulumi/dist-tags | jq -r .[\"feature-dedupeAliases\"])"
+    - curl -sSL https://get.pulumi.com | bash -s -- --version "$(curl -sSL http://registry.npmjs.org/-/package/@pulumi/pulumi/dist-tags | jq -r .latest)"
     - export PATH=$HOME/.pulumi/bin:$PATH
 before_script:
     - ${PULUMI_SCRIPTS}/ci/ensure-dependencies
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
     - sudo apt-get -y install docker-ce=17.09.0~ce-0~ubuntu
     - docker version
     # Install Pulumi
-    - curl -sSL https://get.pulumi.com | bash -s -- --version "$(curl -sSL http://registry.npmjs.org/-/package/@pulumi/pulumi/dist-tags | jq -r .latest)"
+    - curl -sSL https://get.pulumi.com | bash -s -- --version "$(curl -sSL http://registry.npmjs.org/-/package/@pulumi/pulumi/dist-tags | jq -r .[\"feature-dedupeAliases\"])"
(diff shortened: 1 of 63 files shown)


.travis.yml
```diff
@@ -18,7 +18,7 @@ install:
     - sudo apt-get -y install docker-ce=17.09.0~ce-0~ubuntu
     - docker version
     # Install Pulumi
-    - curl -sSL https://get.pulumi.com | bash -s -- --version "$(curl -sSL http://registry.npmjs.org/-/package/@pulumi/pulumi/dist-tags | jq -r .latest)"
+    - curl -sSL https://get.pulumi.com | bash -s -- --version "$(curl -sSL http://registry.npmjs.org/-/package/@pulumi/pulumi/dist-tags | jq -r .[\"feature-dedupeAliases\"])"
```
CyrusNajmabadi:
```suggestion
    - curl -sSL https://get.pulumi.com | bash -s -- --version "$(curl -sSL http://registry.npmjs.org/-/package/@pulumi/pulumi/dist-tags | jq -r .latest)"
```

PR #353, opened 2019-06-26
 More cleanly split how region information is passed along in dashboards. 

CyrusNajmabadi:


nodejs/awsx/cloudwatch/dashboard.ts
```diff
@@ -88,20 +94,20 @@ export class Dashboard extends aws.cloudwatch.Dashboard {
...
-        const region = utils.getRegion(this);
-
         this.url = pulumi.interpolate
             `https://${region}.console.aws.amazon.com/cloudwatch/home?region=${region}#dashboards:name=${this.dashboardName}`;
     }
 }
 
 /** @internal */
-export function getDashboardBody(args: DashboardArgs, opts: pulumi.CustomResourceOptions = {}) {
+export function getDashboardBody(args: DashboardArgs, region: pulumi.Output<aws.Region>) {
```
CyrusNajmabadi: this is just the helper that hte main Dashboard code calls.  It's not exposed to users (hence @internal).  We just have it exported so tests can call it.

CyrusNajmabadi:
Merging in to get things green (and because this really is the better way to do things).

PR #355, opened 2019-07-01
Restore default around how many subnets we pull in for teh default Vpc for a region.

CyrusNajmabadi:
This broke in https://github.com/pulumi/pulumi-awsx/commit/004393885a8dadae5301c00ee04e526004b5e381 (which has not shipped publicly yet).

previously we hardcoded in: 

```ts
        const subnetIds = vpcId.then(id => aws.ec2.getSubnetIds({ vpcId: id }))
            .then(subnets => subnets.ids);
        const subnet0 = subnetIds.then(ids => ids[0]);
        const subnet1 = subnetIds.then(ids => ids[1]);
        const publicSubnetIds = [subnet0, subnet1];
```

Because i could move away from async to sync, i changed the above to:

```ts
const publicSubnetIds = utils.promiseResult(aws.ec2.getSubnetIds({ vpcId }, { provider }).then(subnets => subnets.ids))
```

We need to restore the original behavior as these public subnets are often queried to make other flow-control or other sorts of downstream decisions.

Note: we could always consider an opt-in flag here if customers want to see all public subnets in the default vpc.

### DIFF
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -232,7 +232,12 @@ export class Vpc extends pulumi.ComponentResource {
         const vpcId = utils.promiseResult(aws.ec2.getVpc({ default: true }, { provider }).then(v => v.id));
         let vpc = defaultVpcs.get(vpcId);
         if (!vpc) {
-            const publicSubnetIds = utils.promiseResult(aws.ec2.getSubnetIds({ vpcId }, { provider }).then(subnets => subnets.ids));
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi:
>  these are starting to disappear in new accounts right?

Can you link me to any info on that?  I thought it was teh reverse.  If it was after 2013, you got a default Vpc.

stack72:
@CyrusNajmabadi ignore that point - it's a pipe dream that the team have but it's just too entwined to actually happen - I thought they were starting to disappear but it's more that the bigger infra shops I know delete them in their account

PR #359, opened 2019-07-09
Fill in name and binaryMediaTypes in RestApi.

pgavlin:
Each of these fields is populated by the AWS provider, and any
differences between the read values and the inputs will cause spurious
diffs.

- If `body` is specified, the `name` field will be populated from the
  title present in the Swagger spec. Ensure that we fill in a matching
  name by parsing the spec and extracting its title.
- If `binaryMediaTypes` is unspecified, it will be populated with
  `[ "*/*" ]`, which causes a delete diff. To avoid this, just specify
  `[ "*/*" ]` explicitly.

### DIFF
--- a/nodejs/awsx/apigateway/api.ts
+++ b/nodejs/awsx/apigateway/api.ts
@@ -347,30 +347,39 @@ export class API extends pulumi.ComponentResource {
         super("aws:apigateway:x:API", name, {}, opts);
 
         let swaggerString: pulumi.Output<string>;
-        let swaggerSpec: SwaggerSpec | undefined;
+        let swaggerSpec: pulumi.Output<any>;
         let swaggerLambdas: SwaggerLambdas | undefined;
         if (args.swaggerString) {
-            swaggerString = pulumi.output(args.swaggerString);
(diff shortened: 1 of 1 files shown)


pgavlin:
https://github.com/terraform-providers/terraform-provider-aws/issues/7161 also indicates that the `name` field will be autogenerated if no title is present in the Swagger spec, so we may want to ensure that a title exists.

pgavlin:
(if the spec is provided by the user)

nodejs/awsx/apigateway/api.ts
```diff
@@ -370,6 +370,8 @@ export class API extends pulumi.ComponentResource {
 
         // Create the API Gateway Rest API, using a swagger spec.
         this.restAPI = new aws.apigateway.RestApi(name, {
+            name: swaggerString.apply(s => { const spec = JSON.parse(s); return spec.info.title; }),
```
lukehoban: We should handle `.info` or `.info.title` being undefined.

May be easiest to just fill it in if needed.

pgavlin: Done.

PR #371, opened 2019-07-22
Remove the deprecated Network/Cluster types

CyrusNajmabadi:
Open question on how to deliver this.  As awsx hasn't gotten to 1.0 we could just merge this in now and release prior to 1.0.  Or we can wait and do this as part of an explicit 1.0 piece of work.


CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+### Compatibility issues
+
+* Removing the deprecated awsx.Cluster and awsx.Network type (deprecated in 0.18.6). Code that uses
```
pgavlin: Does this warrant a minor version bump (rather than a patch version bump)? Are there more deprecated types that we plan to remove?

CyrusNajmabadi: @pgavlin: 

> Note: i'm putting this into a 1.0 branch as this is a breaking change

:)

lukehoban: Note that I think we’ll want to release this in our next minor version bump, but will likely want to do that before 1.0. 

We don’t want 0.x to 1.0 to be a meaningful breaking change, it should be more like the end of a 1.0-beta-x chain.

CyrusNajmabadi: Gotcha.  So should this be an 0.19.x change then?  Based on our compat guidelines, it should be, but i just want to make sure that we're on the same page here.

CyrusNajmabadi:
Tagging @pgavlin @mikhailshilkov @lukehoban .   Only open question here is how we deliver this change.  i.e. does it go into our 0.X line of awsx, or does it go into the 1.0/2.0 line.  My preference is on the former, but i wanted to get a read on how everyone else feels about it.  

If there are no concerns, I'll merge this into master for the 0.x line.

pgavlin:
I think putting this in the 0.x line is okay. @lukehoban ?

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ CHANGELOG
   most projects.  However, there may be some code changes needed.  For more details see:
   [#470](https://github.com/pulumi/pulumi-awsx/pull/470).
 
+### Compatibility issues
+
+* Removing the deprecated awsx.Cluster and awsx.Network type (deprecated in 0.18.6). Code that uses
```
pgavlin:
```suggestion
* The deprecated awsx.Cluster and awsx.Network type (deprecated in 0.18.6) have been removed. Code that uses
```

CyrusNajmabadi:
Merging in now.  @lukehoban If you have any concerns here, let us know.

lukehoban:
> I think putting this in the 0.x line is okay. @lukehoban ?

Yes - but it must be in a minor version bump (`0.19`).

Issue #385, opened 2019-08-13
Add method to get all sub resources of a vpc in awsx package


aavileli:
Can we get another method called **fromExistingVpcId** on **awsx.ec2.vpc** class so we don't need to specify all the subresource ids like a data block. This is so silly if the vpc was created using the awsx package you have already set the public, private and isolated subnets with included tags

ericrudder:
Cyrus, this is the one we just chatted about ...

CyrusNajmabadi:
This is doable, but definitely non-trivial.  Many of the resources we make are not necessarily tagged with any sort of information to know how they were created.  We would have to reinfer this stuff.  At a minimum it would involve:

1. walking all subnets to figure them out, and filtering down to only the ones we tag in our current manner.
2. walking all IGs to find those associated with this VPC id.
3. walking the subnets to figure out the routetable and routetableassociation info.
4. figuring out hte association between subnets and IGs in those tables.
5. findin the natgateways and eips, figuring out which correpond here based on the routing tables.

The idea here was that the information would be fairly easy to export if needed.  For example, you really only need to export 6 values from your Vpc, and then use those to rehydrate the Vpc later on.  We could make it easier to get the `ExistingVpcIdArgs` from an actual Vpc.  That way you could just export that for your stack and then read that in later.

PR #392, opened 2019-08-21
Draft of an approach to allow easier roundtripping of a pulumi vpc.

CyrusNajmabadi:
Possible feature to make it easy to export/import a pulumi component (vpc in this case).  Used to address:

https://github.com/pulumi/pulumi-awsx/issues/385

### DIFF
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -291,6 +291,33 @@ export class Vpc extends pulumi.ComponentResource {
 
         return vpc;
     }
+
+    /**
+     * Returns the necessary information about this Vpc to allow easy recreation of it in another
+     * Pulumi stack.  To use, just export like so in one stack:
+     *
+     * ```ts
+     * export const vpcIdArgs = myVpc.getVpcIdArgs();
+     * ```
+     *
+     * Then, from another stack call:
+     *
+     * ```ts
+     * const stackReference = new pulumi.StackReference(...);
+     * const vpcIdArgs: awsx.ec2.ExistingVpcIdArgs = stackReference.getOutputSync("vpcIdArgs");
+     * const vpc = awsx.ec2.Vpc.fromExistingIds(vpcIdArgs);
+     * ```
+     */
+    public getVpcIdArgs(): ExistingVpcIdArgs {
+        return {
+            vpcId: this.vpc.id,
+            publicSubnetIds: this.publicSubnetIds,
+            privateSubnetIds: this.privateSubnetIds,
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi:
Opening this up for feedback.  

Pros: very very very simple.  For complex aggregate components (like awsx.Vpc) the solution for roundtripping (or hydrating across stacks) is to literally let the component itself define how it dehydrates/rehydrates its necessary information.  

Cons: requires any components like this to write the import/export logic.

I can't think of other ways that woudl work here given that a component is not just an aggregation of resources, but also corresponds to real logic that may need to run.  Ultimately, given that, i think the component should have final control and say over how it does this.

--

Other design things to consider.  If a component needed to change, how would properly handle rehydrating from a previously dehydrated POJO?  Should we include a `__version` prop with the POJO output and (if it exists on construction) validate that it's something we know how to hydrate from?

CyrusNajmabadi:
Note: there are very few components in this category (currently).  Basically, only apigateway and Vpc.  there are other components you might want to hydrate, but they're simple enough that they can just be hydrated with a simple ID passed to them.  i.e. they're a component to expose a richer API on top of the low level resource.  but they don't really represent an *aggregatin* of resources.  It's really teh aggregations that are problematic.

CyrusNajmabadi:
@lukehoban @pgavlin  I'd like to discuss this a bit with you to see how we move forward.  I have two general approaches i think we could take.  Both are effectively similar to this PR.  Namely that we can persist/restore a component from a JSON description.  Where they differ is primarily around if we treat these as a black-box or white-box description of the resource.  i.e. we can have something like:

```ts
// black box

public exportDescription(): any { ... }
public static importFromDescription(description: any): TheComponent { ... }
```

or

```ts
// white box
public exportDescription() : TheComponentDescription { ... }
public static importFromDescription(description: TheComponentDescription): TheComponent { ... }
interface TheComponentDescription {
    id: string;
    childId: string;
    relatedResourceIds: string[];
}
```

This PR is an example of white-box approach.  This component already defines a way to hydrate it from existing resource-ids.  So we're just providing the corresponding 'export' function here.  Given that Vpc is one of more complex types, but still is pretty simple to provide this functionality, i'm of the opinion that white-box is a good way to go.

The latter approach gives users the flexibility to provide the data themselves.  However, it ties us down a bit in terms of committing to the import/export format.  The former approach is a bit more flexible for us, because we're not promising anything about the format.  We could change it in hte future if desirable.  

Note: in either case, because this is a persistence format, we likely need to store version-info as well as supporting the persistence format per-component for a very long time.  We'll also have to explicitly test these scenarios out heavily to make sure that this approach works properly across versions.

PR #393, opened 2019-08-23
Allow subnets to inherit opts

RichardWLaub:
In order to prevent pulumi from deleting `kubernetes.io/cluster/<cluster-name>` tags on my subnets I want to add an `ignoreChanges` on tags for my subnets. This PR allows me to do that with something like:

```
new awsx.ec2.Vpc(name, vpcArgs, { ignoreChanges: ['tags'] });
```

 If anyone has a better approach for solving this problem I'm open to ideas!

### DIFF
--- a/nodejs/awsx/ec2/subnet.ts
+++ b/nodejs/awsx/ec2/subnet.ts
@@ -59,7 +59,7 @@ export class Subnet extends pulumi.ComponentResource {
                 vpcId: vpc.id,
                 ...args,
                 assignIpv6AddressOnCreation,
-            }, { parent: this });
+            }, { ...opts, parent: this });
 
             this.routeTable = new aws.ec2.RouteTable(name, {
                 vpcId: vpc.id,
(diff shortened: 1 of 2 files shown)


CyrusNajmabadi:
Unfortunately, this is lkely *not* the way we want to do this here.  But we definitely *do* need a solution here.  Tagging @lukehoban so maybe we can get some thought on this while i'm away.  

What might actually work *fine* here is if we provide a way to pass along 'opts' in the SubnetArgs.  However, that's def something we should make sure we're ok doing as that's not a concept we've ever really had before.  

Another approach is that we supply callbacks or other mechanisms to configure things about child resources being created on behalf of a parent component.

RichardWLaub:
You're right, passing along 'opts' in the SubnetArgs is a better approach. If I get an ok from someone from pulumi I can take a stab at implementing that.

If not, a code example of using callbacks to configure a child resource would be super helpful.

PR #399, opened 2019-09-03
Replacing the go.mod version of apache thrift as old link is broken

stack72:


lukehoban:
Pulled this commit into #401 to try and fix both of the breaking changes that were blocking CI builds.

Issue #406, opened 2019-09-09
Problems upgrading to v1.0.0 (Invalid Identifier)


Blitz2145:
When upgrading from `^0.18.25` to `^1.1.0` of `pulumi-aws`, I am getting the following error during preview:

```
Diagnostics:
  aws:ecs:Cluster (cluster):
    error: Preview failed: refreshing urn:pulumi:qa::blah-site-next::awsx:x:ecs:Cluster$aws:ecs/cluster:Cluster::cluster: error reading ECS Cluster (arn:aws:ecs:us-east-1:149539663033:cluster/arn:aws:ecs:us-east-1:149539663033:cluster/blah-cluster-qa-6372e69): InvalidParameterException: Invalid identifier: Unexpected number of separators
        status code: 400, request id: 68c46836-6b60-4072-b1bb-758e20a4ed43
```

CyrusNajmabadi:
Here's a dump of hte information richard and i acquired while investigating this. 

First, this is not an awsx issue.  And, afaict, this is not an issue with the SDKs at all.  It may be an issue with the engine/provider (unlikely imo), or it's an issue in the terraform layer.

Specifically, we were able to isolate the issue just having Richard try out:

`aws.ecs.Cluster.get("whateveryouwant", "arn:aws:ecs:us-east-1:149539663033:cluster/blah-cluster-qa-6372e69")`

The ID

This fails with the above issue.  It looks like something in the stack is taking the ID up through the `/` and prepending it.  Creating the bogus ID:

`arn:aws:ecs:us-east-1:149539663033:cluster/` + `arn:aws:ecs:us-east-1:149539663033:cluster/blah-cluster-qa-6372e69`

Looking at the logs produced, the SDK properly calls into the engine with:

ReadResource RPC prepared: id=arn:aws:ecs:us-east-1:149539663033:cluster/blah-cluster-qa-6372e69, name=whateveryouwant<{%reset%}>)

So something is happening further down the line to cause things to go off hte rails.  I took a look at the terraform reading code and couldn't find anything that looked obviously amiss.  That said, there is this highly likely issue that was closed out: https://github.com/hashicorp/terraform/issues/20451

So someone else has run into this with terraform, but then nothing came of it.

pgavlin:
@Blitz2145 can you try using just the cluster name in the call to `aws.ecs.Cluster.get`? This is the workaround suggesting in the referenced TF issue.

CyrusNajmabadi:
[...]

PR #407, opened 2019-09-10
Improve docs

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-awsx/issues/406

### DIFF
--- a/nodejs/awsx/ecs/cluster.ts
+++ b/nodejs/awsx/ecs/cluster.ts
@@ -179,8 +179,10 @@ export interface ClusterArgs {
     vpc?: x.ec2.Vpc;
 
     /**
-     * An existing Cluster (or Cluster-Id) to use for this awsx Cluster.  If not provided, a default
-     * one will be created.
+     * An existing aws.ecs.Cluster (or the name of an existing aws.ecs.Cluster) to use for this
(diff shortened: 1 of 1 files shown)


nodejs/awsx/ecs/cluster.ts
```diff
@@ -179,8 +179,10 @@ export interface ClusterArgs {
     vpc?: x.ec2.Vpc;
 
     /**
-     * An existing Cluster (or Cluster-Id) to use for this awsx Cluster.  If not provided, a default
-     * one will be created.
+     * An existing aws.ecs.Cluster (or the name of an existing aws.ecs.Cluster) to use for this
+     * awsx.ecs.Cluster.  If not provided, a default one will be created.
+     *
+     * Note: if passing a string, it must be the *name* of an existing Cluster, not the *id* of one.
```
bermudezmt:
```suggestion
     * Note: If passing a string, use the *name* of an existing ECS Cluster instead of its *id*.
```

Wasn't sure if we wanted to be consistent with using `aws.ecs.Cluster` in the note but this should make it a little clearer.

CyrusNajmabadi: thanks!

PR #431, opened 2019-10-17
Examples: Change the examples to run as individual tests

stack72:


nodejs/awsx/aws_test.go
```diff
@@ -225,21 +259,50 @@ func Test_Examples(t *testing.T) {
...
+
+	integration.ProgramTest(t, &test)
+}
+
+func TestAccNlb_fargate(t *testing.T) {
+	skipIfShort(t)
+	envRegion := getEnvRegion(t)
+	test := getBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir:       path.Join(getCwd(t), "../examples/nlb/fargate"),
```
chrsmith: Nit: Since every test sets the directory `"../examples/"`, perhaps we should add the subfolder (e.g. `"nlb/fargate"`) as a second parameter to `getBaseOptions(t)`? That might clean things up a little bit, what do you think?

nodejs/awsx/aws_test.go
```diff
@@ -250,54 +313,70 @@ func Test_Examples(t *testing.T) {
...
+
+func getDefaultProviderOptions(t *testing.T) integration.ProgramTestOptions {
+	baseJS := integration.ProgramTestOptions{
+		Dependencies: []string{
+			"@pulumi/awsx",
+		},
+		Quick:       true,
+		SkipRefresh: true,
+		// TODO[pulumi/pulumi#1900]: This should be the default value, every test we have causes some sort of
+		// change during a `pulumi refresh` for reasons outside our control.
```
lukehoban: JFWIW - We decided to close 1900 as by design, so we probably can just remove this comment.

nodejs/awsx/aws_test.go
stack72: Good suggestion here - I will follow up and run this across our repo as a separate PR when I get some cleanup time

PR #455, opened 2019-12-03
Fix duplicate H1 tags

cnunciato:
When these get pulled into the docs site, they currently render multiple H1 tags, which flags them in our SEO alerts. This just turns the `H1`s into `H2`s, and moves everything below them one level in.


justinvp:
Alternatively, we keep these as-is in the `README.md`s all of our library repos, and update the doc generator to add an extra `#` to all headings.

PR #502, opened 2020-02-01
Fix ECS TaskDefinitionArgs type does not accept null

ypresto:
Documented as "If `null` no XXX will be created." but its type does not accept null. This PR fixes it!


ypresto:
Fixed in #515

PR #526, opened 2020-04-17
Upgrade to 2.0

leezen:


nodejs/awsx/package.json
```diff
@@ -11,8 +11,8 @@
     "homepage": "https://pulumi.io",
     "repository": "https://github.com/pulumi/pulumi-awsx",
     "dependencies": {
-        "@pulumi/pulumi": "^1.9.1",
-        "@pulumi/aws": "^1.0.0",
+        "@pulumi/pulumi": "^1.9.1 || ^2.0.0",
```
stack72, line 14: why have both versions here? why not just go straight for 2.0 and we can release fixes from the last tag?

leezen: This is a higher-level abstraction library, so it seems more odd to me to force future use onto 2.0.0. There will be some set of people using aws-1.x, in which case, they will also be using pulumi-1.x, so we should allow both.

@lukehoban I think you were also thinking of `awsx` as supporting both provider versions, but it'd be good to get your opinion if you feel otherwise.

leezen:
Oddly, I was getting the mixed test to fail somehow on my local build and not on Travis. Then, I decided to blast everything away to repro and of course, now can't get my local build to have the mixed test fail.

leezen:
This actually implies the mixed mode will work and that the compiler "does the right thing" -- it picks up the version of `@pulumi/pulumi` used by the underlying library of `aws` and so we don't end up with class type conflicts. I tried this even without symlinks and it still works.

nodejs/awsx/package.json
```diff
@@ -10,17 +10,23 @@
...
-        "@pulumi/aws": "^1.0.0",
-        "@pulumi/docker": "^1.0.0",
         "@types/aws-lambda": "^8.10.23",
         "mime": "^2.0.0"
     },
+    "peerDependencies": {
+        "@pulumi/aws": "^1.0.0 || ^2.0.0",
+        "@pulumi/docker": "^1.0.0 || ^2.0.0",
+        "@pulumi/pulumi": "^1.9.1 || ^2.0.0"
+    },
```
lukehoban, line 24: Why was this necessary as part of these changes?

It may well be the "right" thing to do - but it definitely could also introduce a new set of upgrade challenges (for example, many folks will now need to explicitly include `@pulumi/docker` a you had to do here in examples.

leezen: Let me test this real quick, but I don't think we can easily take `@pulumi/docker` out. The reason being that not making it a peerDependency introduces the possibility that `awsx` will now bring in its own version of `@pulumi/pulumi` to satisfy the `@pulumi/docker` dependency. This would then break the mixed version SxS test. If I cannot repro the break, then I'll take out `@pulumi/docker`

leezen: Looks like I was wrong about this. Reasoning through it, it's because there's no interaction between any potential `@pulumi/pulumi` that `@pulumi/docker` might bring in via ComponentResource (we never interact with `docker.Image` in `awsx`). I'll undo the `@pulumi/docker` stuff as a result.

Issue #535, opened 2020-05-15
Allow load balancer to be created using existing aws.lb.LoadBalancer


leezen:
There are cases where users want to use the convenience of awsx, but already have an existing load balancer either from another stack or already in their account. In these cases, they'd like to be able to use the load balancer with the rest of the awsx library, which is not possible today.

PR #541, opened 2020-06-16
feat: allow load balancer to be created using existing lb

eordano:
Addresses https://github.com/pulumi/pulumi-awsx/issues/535

### DIFF
--- a/nodejs/awsx/lb/loadBalancer.ts
+++ b/nodejs/awsx/lb/loadBalancer.ts
@@ -30,6 +30,20 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
     constructor(type: string, name: string, args: LoadBalancerArgs, opts: pulumi.ComponentResourceOptions) {
         super(type, name, {}, opts);
 
+        if (args.loadBalancer !== undefined) {
+            if (Object.keys(args).length > 1) {
+                throw new pulumi.ResourceError("When providing a loadBalancer argument, no other args are allowed", this);
+            }
+            this.loadBalancer = args.loadBalancer;
+            this.vpc = x.ec2.Vpc.fromExistingIds("lb-vpc", { vpcId: this.loadBalancer.vpcId });
+            const securityGroups = this.loadBalancer.securityGroups;
+            if (!securityGroups) {
+                throw new pulumi.ResourceError("Load Balancer is missing info about its security groups", this);
+            }
+            this.securityGroups = x.ec2.getSecurityGroups(this.vpc, name, securityGroups.get(), { parent: this }) || [];
+            return this;
+        }
+
         this.vpc = args.vpc || x.ec2.Vpc.getDefault({ parent: this });
         this.securityGroups = x.ec2.getSecurityGroups(this.vpc, name, args.securityGroups, { parent: this }) || [];
 
@@ -99,6 +113,15 @@ export interface LoadBalancerArgs {
      * unspecified.
      */
     external?: boolean;
+
+    /**
+     * Allows an existing load balancer to be used. There are cases where users want to use the
+     * convenience of awsx, but already have an existing load balancer either from another stack
+     * or already in their account.
+     *
+     * If this value is set, no other arguments are allowed.
+     */
+    loadBalancer?: aws.lb.LoadBalancer;
 
     /**
      * The type of load balancer to create. Possible values are `application` or `network`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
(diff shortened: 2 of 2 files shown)


nodejs/awsx/lb/loadBalancer.ts
```diff
@@ -30,6 +30,20 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
     constructor(type: string, name: string, args: LoadBalancerArgs, opts: pulumi.ComponentResourceOptions) {
         super(type, name, {}, opts);
 
+        if (args.loadBalancer !== undefined) {
+            if (Object.keys(args).length > 1) {
```
leezen, line 34: I don't think this check will work as intended because the concrete classes (`ApplicationLoadBalancer` and `NetworkLoadBalancer`) pass in a `loadBalancerType` to this constructor.

nodejs/awsx/lb/loadBalancer.ts
```diff
@@ -100,6 +114,15 @@ export interface LoadBalancerArgs {
...
     external?: boolean;
 
+    /**
+     * Allows an existing load balancer to be used. There are cases where users want to use the
+     * convenience of awsx, but already have an existing load balancer either from another stack
+     * or already in their account.
+     *
+     * If this value is set, no other arguments are allowed.
+     */
+    loadBalancer?: aws.lb.LoadBalancer;
```
leezen, line 124: Note that the concrete subclasses don't actually have args interfaces that build on the `LoadBalancerArgs` interface, so the `NetworkLoadBalancerArgs` and `ApplicationLoadBalancerArgs` would actually need a `loadBalancer` field as well.

nodejs/awsx/lb/loadBalancer.ts
benmccann: Yep. And `ApplicationLoadBalancer` also passes in default values for `vpc` and `securityGroups`. Arguably those could be removed from `ApplicationLoadBalancer` and we could rely on the defaults here. You may in that case want to set the description in the `securityGroups` here so that we don't lose the description being set in `ApplicationLoadBalancer` by passing in something like `` description: `Default security group for ${this.constructor.name}: ${name}` ``

nodejs/awsx/lb/loadBalancer.ts
```diff
@@ -30,6 +30,20 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
     constructor(type: string, name: string, args: LoadBalancerArgs, opts: pulumi.ComponentResourceOptions) {
         super(type, name, {}, opts);
 
+        if (args.loadBalancer !== undefined) {
+            if (Object.keys(args).length > 1) {
+                throw new pulumi.ResourceError("When providing a loadBalancer argument, no other args are allowed", this);
+            }
+            this.loadBalancer = args.loadBalancer;
+            this.vpc = x.ec2.Vpc.fromExistingIds("lb-vpc", { vpcId: this.loadBalancer.vpcId });
```
benmccann, line 38: @leezen does it matter if `lb-vpc` was not the original name of the vpc? This looks a bit surprising to me, but I'm new to Pulumi

leezen: This is the name of the resource (Pulumi URN) and not the actual VPC name, so it won't matter. However, it'd probably be better if this actually related to the LB name to make it clearer what this resource is, e.g. 
```
`${name}-lb-vpc`
```
or something along those lines.

benmccann: Thanks for clarifying! I like your suggestion, but might tweak it to just `` `${name}-vpc` `` since `name` might already include some test like "lb" or "loadBalancer"

nodejs/awsx/lb/loadBalancer.ts
```diff
@@ -30,6 +30,20 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
...
+            const securityGroups = this.loadBalancer.securityGroups;
+            if (!securityGroups) {
+                throw new pulumi.ResourceError("Load Balancer is missing info about its security groups", this);
```
benmccann, line 41: I don't think it's possible that this condition would be hit

benmccann:
> You might want to take a look at getOrCreateCluster in ecs/cluster.ts which does some checks and also allows one to pass in the existing resource by ID or resource reference as something you might want to also do.

Good suggestion for consistency, but it's not really feasible to use that pattern here. In the case where you pass in a `loadBalancer` you will get the `vpc` and `securityGroups` from it. In the case you're creating a new one you will first create the `vpc` and `securityGroups` and then create the new load balancer using them. So it's harder to share code paths here. The `cluster` case is much simpler because the underlying `cluster` isn't passed the `vpc` and `securityGroups` and so you can use the same code paths more easily.

Issue #543, opened 2020-06-18
No Way to "Get" LoadBalancer resources in Another Stack


WonderPanda:
We are trying to split up our AWS resources across a couple different stacks so isolate things that change often versus resources that are more stable. We have a "core" Stack that defines things like our ECS Cluster, Network Load Balancer, VPC, etc.

Now we're trying to build a separate stack for deploying a Fargate Service onto the cluster but it appears that there's no exposed API for being able to retrieve a reference to the NLB or Target Groups/Listeners. Usually there's some kind of `get` method for being able to look up an already existing resource. Even the `awsx.ecs.Cluster` has the option to construct it using an existing ECS cluster resource passed in to the `cluster` arg which effectively then acts as a reference.

Is there no way to use Load Balancer resources in a multi stack scenario without falling back to just manually creating all of the AWS components ourselves and dropping awsx?

leezen:
Is this a duplicate of https://github.com/pulumi/pulumi-awsx/issues/535 or are you thinking of a different scenario?

WonderPanda:
They're related but I think this is a bit different. In this case I have a Load Balancer created with `awsx` so I'm not trying to adopt an existing LB. I just want the ability to reference the `awsx` resources wrappers inside of another Pulumi Stack so that I can have the ECS Cluster and Load Balancer in one stack and then reference those in another stack to set up a Fargate Service to consume those resources

WonderPanda:
Hmmm appreciate the insight. Rehydrating resources like what is outlined in 4133 definitely seems like it will be a welcome addition but right now I guess an equivalent to #535 might get us to where we need to be. I'd add that it seems like having the equivalent for `TargetGroups` and `Listeners` would be very helpful. In general there's an expectation that there is at least some way to get a reference to a resource in another stack using `.get` or an equivalent from it's Args object.

Any ideas on when #535 might be tackled?

benmccann:
This seems like a duplicate of https://github.com/pulumi/pulumi-aws/issues/824

Can you explain how #535 would help? I see that there's a pending PR for it https://github.com/pulumi/pulumi-awsx/pull/541

benmccann:
[...]

PR #548, opened 2020-07-19
Creation of awsx LoadBalancer from aws LoadBalancer

benmccann:
Updated version of @eordano's https://github.com/pulumi/pulumi-awsx/pull/541 based on code review feedback from @leezen. 

@eordano I hope you don't mind me taking this initiative. I'm in desperate need of this feature, so wanted to lend a hand to move things quickly if possible

Closes https://github.com/pulumi/pulumi-awsx/pull/541
Closes https://github.com/pulumi/pulumi-awsx/issues/535
Closes https://github.com/pulumi/pulumi-awsx/issues/543
Closes https://github.com/pulumi/pulumi-aws/issues/824

The Travis failure is coming from `cloudwatch` and looks unrelated to this PR

### DIFF
--- a/nodejs/awsx/lb/loadBalancer.ts
+++ b/nodejs/awsx/lb/loadBalancer.ts
@@ -30,6 +30,13 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
     constructor(type: string, name: string, args: LoadBalancerArgs, opts: pulumi.ComponentResourceOptions) {
         super(type, name, {}, opts);
 
+        if (args.loadBalancer) {
+            this.loadBalancer = args.loadBalancer;
+            this.vpc = x.ec2.Vpc.fromExistingIds(`${name}-vpc`, { vpcId: this.loadBalancer.vpcId });
+            this.securityGroups = x.ec2.getSecurityGroups(this.vpc, name, this.loadBalancer.securityGroups.get(), { parent: this }) || [];
+            return;
+        }
+
         this.vpc = args.vpc || x.ec2.Vpc.getDefault({ parent: this });
         this.securityGroups = x.ec2.getSecurityGroups(this.vpc, name, args.securityGroups, { parent: this }) || [];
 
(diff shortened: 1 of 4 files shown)


leezen:
There are some build failures here due to some recent changes in `@pulumi/aws` -- I'll open up a separate PR to fix those and we can re-base this PR off that.

leezen:
Opened https://github.com/pulumi/pulumi-aws/pull/1048 (and will need to push a patch release so we can fix the build)

nodejs/awsx/lb/loadBalancer.ts
```diff
@@ -30,6 +30,13 @@ export abstract class LoadBalancer extends pulumi.ComponentResource {
     constructor(type: string, name: string, args: LoadBalancerArgs, opts: pulumi.ComponentResourceOptions) {
         super(type, name, {}, opts);
 
+        if (args.loadBalancer) {
+            this.loadBalancer = args.loadBalancer;
+            this.vpc = x.ec2.Vpc.fromExistingIds(`${name}-vpc`, { vpcId: this.loadBalancer.vpcId });
+            this.securityGroups = x.ec2.getSecurityGroups(this.vpc, name, this.loadBalancer.securityGroups.get(), { parent: this }) || [];
```
leezen, line 36: I don't think `.get()` will work here since the `Output` won't necessarily be known at this point. `securityGroups` also only applies to ALBs, so I don't think you can indiscriminately attempt to get and set like this. Doing an `apply` against `loadBalancer.securityGroups` would be unideal as we'd end up with resource creation within that `apply`, so I'm not sure of a good way around this other than to require passing in the security groups.

benmccann: Updated. Thanks so much for taking a look at this and for fixing the issue in the aws repo as well

benmccann:
@leezen thanks for your work on the AWS package. Unfortunately, Travis still seems to be failing with the same error about the `cloudwatch` functionality. It has picked up the new `aws` npm package

>info Direct dependencies
└─ @pulumi/aws@2.13.1
info All dependencies
├─ @pulumi/aws@2.13.1
└─ builtin-modules@3.0.0

benmccann:
@leezen I rebased this against the latest commit and the build is passing now. thanks for your work on driving this forward!

PR #550, opened 2020-07-22
Update types to accommodate changes in pulumi-aws

leezen:
With `@pulumi/aws v2.10.0` `tags` and `dimensions` on `cloudwatch.MetricAlarm` changed shape. This change updates `awsx` for those shapes. I'm not sure why vpc and autoscaling changed, but I had to update the args for those as well.

I should note that I tested that this compiles cleanly against `@pulumi/aws 2.9.1` (before the dimensions/tags changes). The change here is technically breaking since we narrow from `Record<string, any>` to `[key: string]: string` but at least it appears to still work with previous versions of `@pulumi/aws` (since previous versions accepted `[key: string]: any`)


benmccann:
@lukehoban @stack72 would you be able to take a look at this? it'd be great to get the build working again!

stack72:
Love the changes to keep autoscaling up to date as well here @leezen - sorry it took me so long to get back to you!

Issue #457, opened 2019-12-07
Reconsider resource naming for higher level abstractions


jen20:
I'm currently trying to use `awsx` for some 'serverless' infrastructure, and I think some of the naming - specifically in API Gateway - should be reconsidered.

This program:

```
import * as awsx from "@pulumi/awsx";

async function main() {
    const api = new awsx.apigateway.API("redirector", {
        routes: [{
            path: "/{route+}",
            method: "GET",
            eventHandler: async (_: awsx.apigateway.Request): Promise<awsx.apigateway.Response> => {
                return {
                    statusCode: 200,
                    body: "Hello World",
                };
            }
        }],
    });

    return {
        endpoint: api.url,
    }
}

module.exports = main();
```

Gives:

```
Previewing update (dev):

     Type                                Name                         Plan
 +   pulumi:pulumi:Stack                dev                          create
 +   └─ aws:apigateway:x:API             redirector                   create
 +      ├─ aws:iam:Role                  redirector4fcc7b60           create
 +      ├─ aws:iam:RolePolicyAttachment  redirector4fcc7b60-32be53a2  create
 +      ├─ aws:lambda:Function           redirector4fcc7b60           create
 +      ├─ aws:apigateway:RestApi        redirector                   create
 +      ├─ aws:apigateway:Deployment     redirector                   create
 +      ├─ aws:lambda:Permission         redirector-86405973          create
 +      └─ aws:apigateway:Stage          redirector                   create
```

There are a few concerns about the naming here:

- Some resources have a hex suffix separated by a `-`, some do not. Although AWS names may have differing requirements, the Pulumi names need not be tied to them?
- One resource has _two_ hex suffixes,
- Some resources have no hex suffixes.

I can see an argument that the Pulumi names don't matter - but in that case we shouldn't display them. It seems that the only way to get any control over this is to drop down to raw resources from the `aws` provider which dramatically increases the amount of code required.

leezen:
In addition to this, we should be consistent about the type tokens for resources as there are some that are `awsx:x:...` while others are `awsx:...` or even `aws:...`

PR #557, opened 2020-07-30
Consistently use the awsx namespace

benmccann:
There are the only two that used `aws` instead of `awsx`. I'm not quite sure what the `:x:` is later on in the `apigateway` namespace and if I should leave it or if it's duplicative of the new `x` I just added

### DIFF
--- a/nodejs/awsx/lb/application.ts
+++ b/nodejs/awsx/lb/application.ts
@@ -51,7 +51,7 @@ export class ApplicationLoadBalancer extends mod.LoadBalancer {
             }, opts)];
         }
 
-        super("aws:lb:ApplicationLoadBalancer", name, argsCopy,
+        super("awsx:lb:ApplicationLoadBalancer", name, argsCopy,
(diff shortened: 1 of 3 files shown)


nodejs/awsx/lb/application.ts
```diff
@@ -51,7 +51,7 @@ export class ApplicationLoadBalancer extends mod.LoadBalancer {
             }, opts)];
         }
 
-        super("aws:lb:ApplicationLoadBalancer", name, argsCopy,
+        super("awsx:lb:ApplicationLoadBalancer", name, argsCopy,
             pulumi.mergeOptions(opts, { aliases: [{ type: "awsx:x:elasticloadbalancingv2:ApplicationLoadBalancer"}] }));
```
leezen, line 55: You'd have to do a type alias similar to what's done below already to ensure that resources aren't accidentally created/deleted.

leezen:
Yes, I agree. I'll close this for now. I've updated https://github.com/pulumi/pulumi-awsx/issues/457 to also include the issue with types.

PR #560, opened 2020-07-31
Creation of awsx TargetGroup from aws TargetGroup

benmccann:
Wrapping up from pulumi/pulumi-awsx#553 and https://github.com/pulumi/pulumi-awsx/pull/548. This is the last piece of the puzzle for rehydrating a load balancer

(fwiw, I've actually decided to architect my app a different way, but it seems good to wrap up this outstanding work since the previous PRs can't be used without this one)

### DIFF
--- a/nodejs/awsx/lb/targetGroup.ts
+++ b/nodejs/awsx/lb/targetGroup.ts
@@ -45,10 +45,6 @@ export abstract class TargetGroup
 
         this.vpc = args.vpc;
 
-        if (!args.targetGroup && args.port === undefined) {
-            throw new Error("[port] must be provided.");
-        }
-
         // We used to hash the name of an TG to keep the name short.  This was necessary back when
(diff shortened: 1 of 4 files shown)


benmccann:
@leezen what do you think about this PR?

nodejs/awsx/lb/targetGroup.ts
```diff
@@ -45,11 +45,15 @@ export abstract class TargetGroup
 
         this.vpc = args.vpc;
 
+        if (!args.targetGroup && args.port === undefined) {
```
leezen, line 48: Is there a reason to check `port` here? I believe it can be undefined if the `targetType` is lambda. I guess the old version of the args didn't actually allow this, but I'm not sure that that was actually correct?

benmccann: Ah, yeah, I was just doing it based off the previous args. I've updated it to remove this

PR #571, opened 2020-09-04
Initial scaffolding of GHA work

stack72:
Fixes: #572


stack72:
I am not going to merge this until after the long holiday weekend

nodejs/examples/metrics/mixed/package.json
```diff
@@ -8,8 +8,8 @@
         "build": "tsc"
     },
     "dependencies": {
-        "@pulumi/pulumi": "^1.0.0",
-        "@pulumi/aws": "^2.0.0",
+        "@pulumi/pulumi": "^2.0.0",
```
leezen, line 11: We actually want this to stick to pulumi 1.x -- the point of this test is to check how the potential for SxS pulumi SDK versions works.

.github/workflows/master.yml
```diff
@@ -0,0 +1,206 @@
...
+        run: make istanbul_tests
+    strategy:
+      fail-fast: true
+      matrix:
+        goversion:
+          - 1.14.x
+        language:
+          - nodejs
+        nodeversion:
+          - 13.x
```
leezen, line 88: Shall we bump this to 14?

Makefile
```diff
@@ -1,20 +1,55 @@
...
+#.PHONY: check_clean_worktree
+#check_clean_worktree:
+#	$$(go env GOPATH)/src/github.com/pulumi/scripts/ci/check-worktree-is-clean.sh
+#
+## The travis_* targets are entrypoints for CI.
+#.PHONY: travis_cron travis_push travis_pull_request travis_api
+#travis_cron: all
+#travis_push: only_build check_clean_worktree only_test publish_packages
+#travis_pull_request: only_build check_clean_worktree only_test_fast
+#travis_api: all
```
leezen, line 55: Why keep the commented out code?

leezen (commented):
Given how much we repeat ourselves w/ the yaml, I assume this is generated with some tooling? In which case, is there some way to trace back how the file is generated?

Makefile
stack72: Gone!

.github/workflows/master.yml
stack72: I was trying to make this like for like for now - I can follow up on this one though when we test for NodeJS 14 compatibility

nodejs/examples/metrics/mixed/package.json
stack72: AH great spot! Thanks for that - fixing up now

stack72:
@leezen I am going to address the YAML generation in a different move - the first thing was to get this working, then second will be to make the changes to the yaml generator in our ci-mgmt - that is the easy part :) But no changes will be visible for this repo

jaxxstorm (approved):

leezen (approved):

PR #590, opened 2020-10-20
Add cloudtrail.Trail component

bigkraig:
These are some basic extensions of the resources from `@pulumi/aws`, but they will generate some of the boilerplate resources required.


nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public readonly logGroup: aws.cloudwatch.LogGroup | undefined;
+
+    /**
+     * The managed S3 Bucket.
+     */
+    public readonly bucket: aws.s3.Bucket | undefined;
+
+    /**
+     * The CloudWatch Trail.
```
leezen, line 31:
```suggestion
     * The CloudTrail Trail.
```

nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -0,0 +1,115 @@
...
+import * as utils from "../utils";
+
+export class Trail extends pulumi.ComponentResource {
+    /**
+     * The managed CloudWatch Log Group.
+     */
+    public readonly logGroup: aws.cloudwatch.LogGroup | undefined;
+
+    /**
+     * The managed S3 Bucket.
```
leezen, line 26: I assume this is the S3 bucket that the trail is logged to. Do you mind updating the comment to clarify?

nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -0,0 +1,115 @@
...
+function bucketWithCloudTrailPolicy(name: string, parent: pulumi.Resource | undefined): aws.s3.Bucket {
+    const bucket = new aws.s3.Bucket(name, {}, {parent: parent});
+    const policy = new aws.s3.BucketPolicy(name, {
+        bucket: bucket.id,
+        policy: {
+            Version: "2012-10-17",
+            Statement: [{
+                Sid: "AWSCloudTrailAclCheck",
+                Effect: "Allow",
+                Principal: {Service: "cloudtrail.amazonaws.com"},
```
leezen, line 101: What do you think about using `aws.iam.principals.CloudtrailPrincipal` here instead?

nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.trail = new aws.cloudtrail.Trail(name, trailArgs, {parent: this});
+    }
+}
+
+
+type TrailArgs = utils.Overwrite<aws.cloudtrail.TrailArgs, {
+    /**
+     * If CloudTrail pushes logs to CloudWatch Logs in addition to S3.
+     *
+     * Disabled for cost out of the box.
```
leezen, line 74:
```suggestion
     * Disabled by default to reduce costs.
```

nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @default false
+     */
+    readonly sendToCloudWatchLogs?: pulumi.Input<boolean>;
+
+    /**
+     * If sendToCloudWatchLogs is enabled, provide these arguments to configure
+     * the log group.
```
leezen, line 82:
```suggestion
     * If sendToCloudWatchLogs is enabled, provide the log group configuration.
```

nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -0,0 +1,115 @@
...
+
+type TrailArgs = utils.Overwrite<aws.cloudtrail.TrailArgs, {
+    /**
+     * If CloudTrail pushes logs to CloudWatch Logs in addition to S3.
+     *
+     * Disabled for cost out of the box.
+     *
+     * @default false
+     */
+    readonly sendToCloudWatchLogs?: pulumi.Input<boolean>;
```
leezen, line 78: In the constructor above, we read `sendToCloudWatchLogs` as a prompt value, so this can't be an `Input`. I also wouldn't expect this or any of the arguments in this interface to be marked as `readonly`.

leezen (commented):
Thanks for the PR!

I had a couple minor nits on the `Trail` side of things. I also don't think any of the args need to be marked as `readonly`?

On the `Function` side, I'm curious if you think it might make more sense to actually work it into https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/lambda/lambdaMixins.ts instead? There, there's already some functionality to create a default role, etc. I wonder if that can be refactored out to provide the same functionality you're trying to add in this PR for `Function`. What do you think?

nodejs/awsx/cloudtrail/trail.ts
bigkraig: I'll remove the `readonly` and `Input`, but I was duplicating what has been done in the existing interface: https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/cloudtrail/trail.ts#L390-L405

bigkraig:
> On the `Function` side, I'm curious if you think it might make more sense to actually work it into https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/lambda/lambdaMixins.ts instead? There, there's already some functionality to create a default role, etc. I wonder if that can be refactored out to provide the same functionality you're trying to add in this PR for `Function`. What do you think?

Doesn't matter much to me. I wasn't aware of this and now i'm worried there may be more extensions to the AWS provider that I've been missing out on. How is the separation between `pulumi-aws` and `pulumi-awsx` defined? My understanding was that `pulumi-aws` was fairly raw and `pulumi-awsx` would contain higher level constructs.

nodejs/awsx/cloudtrail/trail.ts
leezen: Makes sense, but the rest of this package doesn't do that, so I think the consistency (of not having `readonly`) is preferable.

leezen:
> Doesn't matter much to me. I wasn't aware of this and now i'm worried there may be more extensions to the AWS provider that I've been missing out on. How is the separation between pulumi-aws and pulumi-awsx defined? My understanding was that pulumi-aws was fairly raw and pulumi-awsx would contain higher level constructs.

The function stuff in `pulumi-aws` is there to make it possible to provide closure serialization as a convenience for where you'd expect to be able to provide a callback. These typically hang off the resources where possible (search for 'mixins' in the repo). The other extension to the AWS provider that lives within `pulumi-aws` are various enums. Otherwise, higher level abstractions/components tend find to their way into `pulumi-awsx` (or other third party abstraction libraries).

The `Function` proposal you have here would make sense in the context of the other mixin capabilities in `pulumi-aws`.

bigkraig:
All set then 👍

leezen:
Great! Thanks for the contribution!

PR #594, opened 2020-11-16
Move away from GHA deprecated add-path

stack72:


stack72:
This is mechanical changes only - self merging as it only updates GHA syntax - no functionality was changed and we have a green build!

Issue #605, opened 2020-12-07
Unable to add path to HTTP healthcheck for NLB target group


adrianleonmorell:
Hi guys, I'm trying to add HTTP healtchecks to a NLB target group but I'm unable to add the path property. This is apparently possible using the AWS console
![image](https://user-images.githubusercontent.com/17031755/101342545-045ce480-3883-11eb-82e3-c9ca9af82b85.png)

This is what I'm trying with pulumi:
![image](https://user-images.githubusercontent.com/17031755/101342338-b5af4a80-3882-11eb-9094-d942ae76bc9a.png)
but the parameter path is of type never.

Is there something I'm missing? Can you guide my in the right direction?
Thanks in advance

leezen:
Thanks for reporting this. This looks to be a bug in the library where it hasn't been updated to keep up with the surface area of the underlying NLB API.

adrianleonmorell:
Hi @leezen, do you know if this will be released soon? Thank you, for the quick feedback.

leezen:
@adrianleonmorell I'll work on cutting a release today

adrianleonmorell:
Thanks @leezen

leezen:
@adrianleonmorell In case you didn't see, 0.23.0 is now available.

adrianleonmorell:
Hi @leezen I have being trying to create the NLB using awsx 0.23.0 and I have found 2 things:
1- The type for the timeout property for the healthchecks is set to never for the NLB so it's always created with 6 seconds
2- The matcher property is allowed but I get a runtime error if something different of "200-399" is specified (not a big deal for me since my healthchecks return 200 or 503)

This is the code sample I'm using
```
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";

const vpc = new awsx.ec2.Vpc(`test-vpc`, {
  numberOfAvailabilityZones: 2,
  subnets: [
    {
      type: "private",
      name: `private-subnet`,
    },
  ],
  tags: { Name: `test-vpc` },
});

const nlb = new awsx.lb.NetworkLoadBalancer("test-nlb", {
  vpc: vpc,
  subnets: vpc.privateSubnetIds,
  external: false,
});

nlb.createListener("test-listener", {
  port: 8080,
  protocol: "TCP",
  targetGroup: {
    port: 80,
    protocol: "TCP",
    healthCheck: {
      protocol: "HTTP",
[...]

PR #606, opened 2020-12-07
Update NetworkTargetGroupHealthCheck to match specification

leezen:
Based on https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html it's possible to set the path and matcher for NLB target group health checks.

Fixes https://github.com/pulumi/pulumi-awsx/issues/605

### DIFF
--- a/nodejs/awsx/lb/network.ts
+++ b/nodejs/awsx/lb/network.ts
@@ -309,27 +309,10 @@ export interface NetworkLoadBalancerArgs {
  */
 export interface NetworkTargetGroupHealthCheck extends mod.TargetGroupHealthCheck {
     /**
-     * Applies to Application Load Balancers only (HTTP/HTTPS), not Network Load Balancers (TCP)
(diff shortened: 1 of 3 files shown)


nodejs/awsx/lb/network.ts
```diff
@@ -308,28 +308,11 @@ export interface NetworkLoadBalancerArgs {
...
      * for TCP and HTTPS health checks and 6 seconds for HTTP health checks.
      */
     timeout?: never;
-
-    /**
-     * The number of consecutive health check failures required before considering the target
-     * unhealthy . For Network Load Balancers, this value must be the same as the
-     * healthy_threshold.
-     */
-    unhealthyThreshold?: pulumi.Input<number>;
```
leezen, line 332: Removed this as it's already in `TargetGroupHealthCheck`

viveklak (approved):

PR #612, opened 2020-12-22
Adding a cron build to ensure that provider breakages are picked up

stack72:
Fixes: #601


leezen (approved):
Whose 9am? Maybe just rip out that part of the comment? Kind of confusing.

stack72:
sure!

Issue #603, opened 2020-12-02
Task Definition log configuration being overwritten


pierskarsenbarg:
```typescript
import * as awsx from "@pulumi/awsx";

const cluster = new awsx.ecs.Cluster("pk-cluster", {
  name: "pk-cluster",
});

// Create a load balancer on port 80 and spin up two instances of Nginx.
const lb = new awsx.lb.ApplicationListener("nginx-lb", { port: 80 });

const fargateTask = new awsx.ecs.FargateTaskDefinition(`firelens-fargatetask`, {
  containers: {
    mycontainer: {
      image: "nginx",
      essential:true,
      memory: 128,
      portMappings: [lb],
      logConfiguration: {
        logDriver: "awslogs",
        options: {
          "awslogs-group": "nginx-container",
          "awslogs-region": "eu-west-1",
          "awslogs-stream-prefix": "nginx",
        },
      },
    },
  },
  tags: {
      Owner: "piers"
  }
});

const service = new awsx.ecs.FargateService("my-service", {
  cluster: cluster,
  taskDefinition: fargateTask,
  desiredCount: 1,
});

// Export the load balancer's address so that it's easy to access.
export const url = lb.endpoint.hostname;

```

When I run the above code, the logConfiguration is being overwritten:

![image](https://user-images.githubusercontent.com/109047/100887076-76f94900-34ac-11eb-9f6b-9758755e357c.png)

This is fine(ish) here, but if someone is using firelens for logging then the logDriver needs to be `awsfirelens`.

(I think it's being overwritten by [this code](https://github.com/pulumi/pulumi-awsx/blob/99678b954766dbca4aec53419568afc7e4b940ad/nodejs/awsx/ecs/container.ts#L58-L66) but I'm not sure what it _should_ be.

Example with firelens which isn't working:

```typescript
import * as awsx from "@pulumi/awsx";

const cluster = new awsx.ecs.Cluster("pk-cluster", {
  name: "pk-cluster",
});

// Create a load balancer on port 80 and spin up two instances of Nginx.
const lb = new awsx.lb.ApplicationListener("nginx-lb", { port: 80 });

const fargateTask = new awsx.ecs.FargateTaskDefinition(`firelens-fargatetask`, {
  containers: {
    mycontainer: {
      image: "nginx",
      essential:true,
      memory: 128,
      portMappings: [lb],
      logConfiguration: {
        logDriver: "awslogs",
        options: {
          "awslogs-group": "nginx-container",
[...]

PR #625, opened 2021-01-20
Allow a user to pass their own log configuration to ecs.TaskDefinitions

stack72:
Fixes: #603


nodejs/awsx/ecs/taskDefinition.ts
```diff
@@ -155,7 +155,10 @@ export abstract class TaskDefinition extends pulumi.ComponentResource {
 
     // Default policy arns for the Execution role.
     public static defaultExecutionRolePolicyARNs() {
-        return ["arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"];
+        return [
+            "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy",
+            aws.iam.ManagedPolicies.CloudWatchLogsFullAccess,
```
stack72, line 160: Without this policy, there is no way for our service to be able to create it's LogGroup

lukehoban: The `AmazonECSTaskExecutionRolePolicy` policy provides the needed CloudWatch actions.  This change should not be necessary.  What exactly did you hit that made you think it was necessary?

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html

stack72: It allows interacting with cloudwatch but it doesnlt allow creating a log group - that's handled by logs.CreateLogGroup

lukehoban: Why does it need to create a log group? What issue were you hitting?

pierskarsenbarg: It currently creates a log group. At the moment the end user has no control over the name and details (which is the issue that the user is having) and so they can't create one for the firelens logs

stack72: @lukehoban any extra thoughts here?

leezen: I think one thing we could do here if we want to keep the permissions scoped to a minimum is actually attach `AWSLambdaBasicExecutionRole` instead, which provides just:

```
"logs:CreateLogGroup",
"logs:CreateLogStream",
"logs:PutLogEvents"
```

which I think is all that's needed vs. giving full access to logs.

pierskarsenbarg: I think this will do. Or can we create a role that's not going to confuse people if they look up what resources are being used (if I saw a lambda role being attached to something ECS-related I might question it)?

stack72: Ok, I have updated to make this way less restrictive as @leezen suggests - I believe we are in a better shape now

leezen (approved):

PR #631, opened 2021-01-26
Update for Lambda full access deprecation

leezen:
- Update to `@pulumi/aws` 3.25.1 so that callback functions without specified policies create a copy of the deprecated policy.
- [Breaking] Update the default task role in `ecs.TaskDefinition` to use the more scoped down `LambdaFullAccess` policy (the new one that AWS recommends).

### DIFF
--- a/nodejs/awsx/ecs/taskDefinition.ts
+++ b/nodejs/awsx/ecs/taskDefinition.ts
@@ -146,8 +146,8 @@ export abstract class TaskDefinition extends pulumi.ComponentResource {
     // Default policy arns for the Task role.
     public static defaultTaskRolePolicyARNs() {
         return [
-            // Provides wide access to "serverless" services (Dynamo, S3, etc.)
-            aws.iam.ManagedPolicies.AWSLambdaFullAccess,
+            // Provides full access to Lambda
+            aws.iam.ManagedPolicy.LambdaFullAccess,
             // Required for lambda compute to be able to run Tasks
             aws.iam.ManagedPolicy.AmazonECSFullAccess,
         ];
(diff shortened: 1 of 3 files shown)


nodejs/awsx/ecs/taskDefinition.ts
```diff
@@ -146,8 +146,8 @@ export abstract class TaskDefinition extends pulumi.ComponentResource {
     // Default policy arns for the Task role.
     public static defaultTaskRolePolicyARNs() {
         return [
-            // Provides wide access to "serverless" services (Dynamo, S3, etc.)
-            aws.iam.ManagedPolicies.AWSLambdaFullAccess,
+            // Provides full access to Lambda
+            aws.iam.ManagedPolicy.LambdaFullAccess,
```
stack72, line 150: Is this in *any* way breaking? i.e. is there any potential issue that someone has already deployed a Task role with the old IAM Policy and then will be downgraded to his Policy?

leezen: Yes, this is breaking as noted in the description above.

stack72: Ok, do we need to give an indication of what they need to do if they still need the old role?

stack72 (approved):

nodejs/awsx/ecs/taskDefinition.ts
leezen: Good point, I'll add this to the changelog. After I merge this, I'll go ahead and cut a release.

stack72:
Happy to see this and Pulumi-cloud closed out - thanks so much for helping out with this @leezen

PR #638, opened 2021-02-15
Allow using specific zones when creating a VPC.

seivan:
Fixing: https://github.com/pulumi/pulumi-awsx/issues/493

Fixed the [review comments](https://github.com/pulumi/pulumi-awsx/pull/494) and made sure to base of v2. 
Took the liberty of adding deprecation warning to `numberOfAvailabilityZones` and introducing `requestedAvailabilityZone?: number | "all" | [string, ...string[]];`

That being said, I am not sure it's a great idea to rely on fresh calls to `getAvailabilityZones()`, maybe I'm wrong here, but couldn't there be a potential issue of getting different zones back, thus mutating an existing VPC? 

Assuming it does what I suspect, I'd think it's better to allow the user to decide if throwing an error is better than mutating an existing VPC when AWS returns different zones than what you're expecting. 

My take is, remove `"all"` & `number` and let the zones be explicit, or maybe use a cache with the argument as seed and if a zone is down, just chuck an error as it should, anyway that's out of scope from what I need right now.

### DIFF
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -522,8 +522,8 @@ async function getAvailabilityZones(
                 mappedZones.length === requestedZones.length ?
                     resolve(mappedZones) :
                     reject(new pulumi.ResourceError("Availability zones did not match requested zones", parent));
-            })
-        })
+            });
+        });
     }
     else {
         return descriptions.slice(0, requestedZones === "all" ? descriptions.length : requestedZones);
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -327,7 +327,7 @@ export class Vpc extends pulumi.ComponentResource<VpcData> {
     /** @internal */
     public async initializeVpcArgs(name: string, args: x.ec2.VpcArgs, opts: pulumi.ComponentResourceOptions): Promise<VpcData> {
         const provider = Vpc.getProvider(opts);
-        const availabilityZones = await getAvailabilityZones(opts.parent, provider, args.numberOfAvailabilityZones);
+        const availabilityZones = await getAvailabilityZones(opts.parent, provider, args.requestedAvailabilityZones ?? args.numberOfAvailabilityZones);
 
         return new VpcData(name, this, {
             ...args,
@@ -504,7 +504,7 @@ utils.Capture(Vpc.prototype).initializeVpcArgs.doNotCapture = true;
 async function getAvailabilityZones(
         parent: pulumi.Resource | undefined,
         provider: pulumi.ProviderResource | undefined,
-        requestedCount: "all" | number | undefined): Promise<topology.AvailabilityZoneDescription[]> {
+        requestedZones: string[] | number | "all" | undefined): Promise<topology.AvailabilityZoneDescription[]> {
 
     const result = await aws.getAvailabilityZones(/*args:*/ undefined, { provider, async: true });
     if (result.names.length !== result.zoneIds.length) {
@@ -516,11 +516,22 @@ async function getAvailabilityZones(
         descriptions.push({ name: result.names[i], id: result.zoneIds[i] });
     }
 
-    const count =
-        typeof requestedCount === "number" ? requestedCount :
-        requestedCount === "all" ? descriptions.length : 2;
+    if (Array.isArray(requestedZones)) {
+        const mappedZones = descriptions.filter(zone => requestedZones.includes(zone.name));
 
-    return descriptions.slice(0, count);
+        if (mappedZones.length !== requestedZones.length) {
+            throw new pulumi.ResourceError("Availability zones did not match requested zones", parent);
+        }
+        return mappedZones
+    }
+    else {
+        const count =
+            typeof requestedZones === "number" ? requestedZones :
+                requestedZones === "all" ? descriptions.length : 2;
+
+        return descriptions.slice(0, count);
+    }
+
 }
 
 function getExistingSubnets(vpcData: VpcData, vpc: Vpc, vpcName: string, type: VpcSubnetType, inputs: pulumi.Input<string>[] = []) {
@@ -702,10 +713,15 @@ export interface VpcArgs {
     subnets?: VpcSubnetArgs[];
 
     /**
-     * The maximum number of availability zones to use in the current region.  Defaults to `2` if
-     * unspecified.  Use `"all"` to use all the availability zones in the current region.
+     * @deprecated Use `requestedAvailabilityZones`
      */
-    numberOfAvailabilityZones?: number | "all";
+    numberOfAvailabilityZones?: number | "all" | string[];
+
+    /**
+     * The names of the availability zones to use in the current region. Defaults to `2` if
+     * unspecified. Use `"all"` to use all the availability zones in the current region.
+     */
+    requestedAvailabilityZones?: number | "all" | string[];
 
     /**
      * The max number of NAT gateways to create if there are any private subnets created.  A NAT
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -522,7 +522,7 @@ async function getAvailabilityZones(
         if (mappedZones.length !== requestedZones.length) {
             throw new pulumi.ResourceError("Availability zones did not match requested zones", parent);
         }
-        return mappedZones
+        return mappedZones;
     }
     else {
         const count =
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -514,13 +514,16 @@ async function getAvailabilityZones(
 
     const descriptions = result.names.map((name, idx) => ({ name, id: result.zoneIds[idx] }) )
 
-    if (Array.isArray(requestedZones)) {
-        const mappedZones = descriptions.filter(zone => requestedZones.includes(zone.name));
 
-        if (mappedZones.length !== requestedZones.length) {
-            throw new pulumi.ResourceError("Availability zones did not match requested zones", parent);
-        }
-        return mappedZones;
+    if (typeof requestedZones === "object") {
+        return new Promise((resolve, reject) => {
+            pulumi.Output.create(requestedZones).apply(requestedZones => {
+                const mappedZones = descriptions.filter(zone => requestedZones.includes(zone.name));
+                mappedZones.length === requestedZones.length ?
+                    resolve(mappedZones) :
+                    reject(new pulumi.ResourceError("Availability zones did not match requested zones", parent));
+            })
+        })
     }
     else {
         const count =
@@ -714,7 +717,7 @@ export interface VpcArgs {
      * The names of the availability zones to use in the current region. Defaults to `2` if
      * unspecified. Use `"all"` to use all the availability zones in the current region.
      */
-    requestedAvailabilityZones?: number | "all" | [string, ...string[]];
+    requestedAvailabilityZones?: number | "all" | pulumi.Input<[string, ...string[]]>;
 
     /**
      * @deprecated Use `requestedAvailabilityZones`
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -512,7 +512,7 @@ async function getAvailabilityZones(
     }
 
 
-    const descriptions = result.names.map((name, idx) => ({ name, id: result.zoneIds[idx] }) )
+    const descriptions = result.names.map((name, idx) => ({ name, id: result.zoneIds[idx] }) );
 
     if (Array.isArray(requestedZones) || typeof requestedZones === "object" ) {
         return new Promise((resolve, reject) => {
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -714,9 +714,6 @@ export interface VpcArgs {
      */
     requestedAvailabilityZones?: number | "all" | [string, ...string[]] | pulumi.Input<string[]>;
 
-    /**
-     * @deprecated Use `requestedAvailabilityZones`
-     */
     numberOfAvailabilityZones?: VpcArgs["requestedAvailabilityZones"];
 
     /**
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -504,7 +504,7 @@ utils.Capture(Vpc.prototype).initializeVpcArgs.doNotCapture = true;
 async function getAvailabilityZones(
         parent: pulumi.Resource | undefined,
         provider: pulumi.ProviderResource | undefined,
-        requestedZones: VpcArgs["requestedAvailabilityZones"]): Promise<topology.AvailabilityZoneDescription[]> {
+        requestedZones: VpcArgs["requestedAvailabilityZones"] = 2): Promise<topology.AvailabilityZoneDescription[]> {
 
     const result = await aws.getAvailabilityZones(/*args:*/ undefined, { provider, async: true });
     if (result.names.length !== result.zoneIds.length) {
@@ -526,11 +526,7 @@ async function getAvailabilityZones(
         })
     }
     else {
-        const count =
-            typeof requestedZones === "number" ? requestedZones :
-                requestedZones === "all" ? descriptions.length : 2;
-
-        return descriptions.slice(0, count);
+        return descriptions.slice(0, requestedZones === "all" ? descriptions.length : requestedZones);
     }
 
 }
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -514,8 +514,7 @@ async function getAvailabilityZones(
 
     const descriptions = result.names.map((name, idx) => ({ name, id: result.zoneIds[idx] }) )
 
-
-    if (typeof requestedZones === "object") {
+    if (Array.isArray(requestedZones) || typeof requestedZones === "object" ) {
         return new Promise((resolve, reject) => {
             pulumi.Output.create(requestedZones).apply(requestedZones => {
                 const mappedZones = descriptions.filter(zone => requestedZones.includes(zone.name));
@@ -713,7 +712,7 @@ export interface VpcArgs {
      * The names of the availability zones to use in the current region. Defaults to `2` if
      * unspecified. Use `"all"` to use all the availability zones in the current region.
      */
-    requestedAvailabilityZones?: number | "all" | pulumi.Input<[string, ...string[]]>;
+    requestedAvailabilityZones?: number | "all" | [string, ...string[]] | pulumi.Input<string[]>;
 
     /**
      * @deprecated Use `requestedAvailabilityZones`
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -504,7 +504,7 @@ utils.Capture(Vpc.prototype).initializeVpcArgs.doNotCapture = true;
 async function getAvailabilityZones(
         parent: pulumi.Resource | undefined,
         provider: pulumi.ProviderResource | undefined,
-        requestedZones: [string, ...string[]] | number | "all" | undefined): Promise<topology.AvailabilityZoneDescription[]> {
+        requestedZones: VpcArgs["requestedAvailabilityZones"]): Promise<topology.AvailabilityZoneDescription[]> {
 
     const result = await aws.getAvailabilityZones(/*args:*/ undefined, { provider, async: true });
     if (result.names.length !== result.zoneIds.length) {
@@ -512,7 +512,7 @@ async function getAvailabilityZones(
     }
 
 
-    const descriptions = result.names.map((name, idx) => ({name, id:result.zoneIds[idx]}))
+    const descriptions = result.names.map((name, idx) => ({ name, id: result.zoneIds[idx] }) )
 
     if (Array.isArray(requestedZones)) {
         const mappedZones = descriptions.filter(zone => requestedZones.includes(zone.name));
@@ -711,15 +711,15 @@ export interface VpcArgs {
     subnets?: VpcSubnetArgs[];
 
     /**
-     * @deprecated Use `requestedAvailabilityZones`
-     */
-    numberOfAvailabilityZones?: number | "all" | [string, ...string[]];
-
-    /**
      * The names of the availability zones to use in the current region. Defaults to `2` if
      * unspecified. Use `"all"` to use all the availability zones in the current region.
      */
     requestedAvailabilityZones?: number | "all" | [string, ...string[]];
+
+    /**
+     * @deprecated Use `requestedAvailabilityZones`
+     */
+    numberOfAvailabilityZones?: VpcArgs["requestedAvailabilityZones"];
 
     /**
      * The max number of NAT gateways to create if there are any private subnets created.  A NAT
--- a/nodejs/awsx/ec2/vpc.ts
+++ b/nodejs/awsx/ec2/vpc.ts
@@ -504,17 +504,15 @@ utils.Capture(Vpc.prototype).initializeVpcArgs.doNotCapture = true;
 async function getAvailabilityZones(
         parent: pulumi.Resource | undefined,
         provider: pulumi.ProviderResource | undefined,
-        requestedZones: string[] | number | "all" | undefined): Promise<topology.AvailabilityZoneDescription[]> {
+        requestedZones: [string, ...string[]] | number | "all" | undefined): Promise<topology.AvailabilityZoneDescription[]> {
 
     const result = await aws.getAvailabilityZones(/*args:*/ undefined, { provider, async: true });
     if (result.names.length !== result.zoneIds.length) {
         throw new pulumi.ResourceError("Availability zones for region had mismatched names and ids.", parent);
     }
 
-    const descriptions: topology.AvailabilityZoneDescription[] = [];
-    for (let i = 0, n = result.names.length; i < n; i++) {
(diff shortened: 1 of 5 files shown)


nodejs/awsx/ec2/vpc.ts
```diff
@@ -516,11 +516,22 @@ async function getAvailabilityZones(
...
-    const count =
-        typeof requestedCount === "number" ? requestedCount :
-        requestedCount === "all" ? descriptions.length : 2;
+    if (Array.isArray(requestedZones)) {
+        const mappedZones = descriptions.filter(zone => requestedZones.includes(zone.name));
+
+        if (mappedZones.length !== requestedZones.length) {
+            throw new pulumi.ResourceError("Availability zones did not match requested zones", parent);
+        }
+        return mappedZones
```
leezen, line 525:
```suggestion
        return mappedZones;
```

leezen (commented):
Overall, this seems reasonable. Do you mind adding an example and wiring it up so that it gets picked up as part of acceptance tests for this package?

seivan:
> Do you mind adding an example and wiring it up so that it gets picked up as part of 
Where and how exactly? 


The more I think about this, the more I am starting to realize that allowing a number is a really bad idea. 
It just dawned on me as I was setting up RDS cluster and needed to associate `availabilityZones`.

There is a large chance that the zones you get for your subnets in the VPC by declaring a number alone, will be different than the zones you've set up on your RDS cluster. 

Just to be clear, I am not suggesting to do anything now, but it could be a big deal.

leezen:
> There is a large chance that the zones you get for your subnets in the VPC by declaring a number alone, will be different than the zones you've set up on your RDS cluster.

I'm missing the use case. Typically, you'd setup the RDS instance within the VPC you're creating, right? So the cluster would have a dependency on the VPC you create.

seivan:
> I'm missing the use case. Typically, you'd setup the RDS instance within the VPC you're creating, right? So the cluster would have a dependency on the VPC you create.

You need to supply an explicit `availabilityZone` for each `ClusterInstance`
That might not match the ones that you've gotten from the VPC, **in fact**, you can't tell which ones you got from the VPC to begin with.

seivan:
> Do you mind adding an example and wiring it up so that it gets picked up as part of 

Where and how exactly?

leezen:
> You need to supply an explicit availabilityZone for each ClusterInstance. That might not match the ones that you've gotten from the VPC, in fact, you can't tell which ones you got from the VPC to begin with.

This should be available as the `availabilityZone` property, e.g.:

```typescript
const vpc = new awsx.ec2.Vpc("vpc", {});
const cluster = new aws.rds.Cluster("cluster", {
    ...
    availabilityZones: vpc.privateSubnets.then(subnets => subnets.map(s => s.subnet.availabilityZone)),
})
```

Or are you trying to do something else?

> Where and how exactly?

You'll want to add an example to `nodejs/examples` and then modify `nodejs/awsx/aws_test.go` to include it as another test.

seivan:
> This should be available as the `availabilityZone` property, e.g.:
> 
> ```ts
> const vpc = new awsx.ec2.Vpc("vpc", {});
> const cluster = new aws.rds.Cluster("cluster", {
>     ...
>     availabilityZones: vpc.privateSubnets.then(subnets => subnets.map(s => s.subnet.availabilityZone)),
> })
> ```
> 
> Or are you trying to do something else?

No, this is exactly it, thank you!
But correct me if I am wrong here, but this relies on the VPC call always giving you the same zones for your subnets. 
If at any time you get different set of zones for whatever reason, then you're going to be giving different zones to your RDS cluster and it might have some pretty significant impact. 


> > Where and how exactly?
> 
> You'll want to add an example to `nodejs/examples` and then modify `nodejs/awsx/aws_test.go` to include it as another test.

Thanks, I'll get to it.

leezen:
> But correct me if I am wrong here, but this relies on the VPC call always giving you the same zones for your subnets.
If at any time you get different set of zones for whatever reason, then you're going to be giving different zones to your RDS cluster and it might have some pretty significant impact.

These zones shouldn't change. These are the defined outputs of the component and don't change unless you change the underlying VPC resource.

seivan:
> These zones shouldn't change. These are the defined outputs of the component and don't change unless you change the underlying VPC resource.

There's no guarantee of that. The issue is that it's a random selection of zones from a user PoV. There's nothing to indicate what happens when a zone is down and the response consist of a different set of zones, which might cause all sorts of havoc, when it should in reality just fail and throw an error. 

I am not saying this is necessary the case, I am saying from a API PoV there are no guarantees set on how and what is chosen. 

The `getAvailabilityZones` function doesn't indicate what the default set of arguments are from 
```ts
export interface GetAvailabilityZonesArgs {

    readonly allAvailabilityZones?: boolean;

    readonly excludeNames?: string[];

    readonly excludeZoneIds?: string[];

    readonly filters?: inputs.GetAvailabilityZonesFilter[];

    readonly state?: string;
}
```

And it's being passed `undefined` so I assume there is some sort of a default, but there's nothing to document what those are. 

That being said, a user could use this to throw some errors themselves
`availabilityZones: vpc.privateSubnets.then(subnets => subnets.map(s => s.subnet.availabilityZone))`

seivan:
So I added a test-case with three specific zones under `{ region: "us-east-1" }`, but I am not sure how to run the tests. 

Regarding `nodejs/awsx/aws_test.go`,  I did notice there was something that selected the entire VPC test, so I am guessing it might be included there on its own? 

```go
func TestAccVpc(t *testing.T) {
	test := getBaseOptions(t).
		With(integration.ProgramTestOptions{
			Dir:       path.Join(getCwd(t), "../examples/vpc"),
			StackName: addRandomSuffix("vpc"),
		})

	integration.ProgramTest(t, &test)
}
```

nodejs/examples/vpc/index.ts
```diff
@@ -53,4 +53,14 @@ const vpcWithLocations = new awsx.ec2.Vpc("custom6", {
...
+
+const vpcWithSpecificZones = new awsx.ec2.Vpc("custom7", {
+    requestedAvailabilityZones: ["us-east-1a","us-east-1b","us-east-1c"],
+    subnets: [
+        { type: "public"},
+        { type: "private"},
+        { type: "isolated", name: "db"},
+        { type: "isolated", name: "redis"},
+    ],
+}, { provider: new aws.Provider("prov2", { region: "us-east-1" }) });
```
leezen, line 66: Add newline

nodejs/examples/vpc/index.ts
```diff
@@ -53,4 +53,14 @@ const vpcWithLocations = new awsx.ec2.Vpc("custom6", {
         { type: "isolated", name: "db", location: "10.0.2.0/24" },
         { type: "isolated", name: "redis", location: "10.0.3.0/24" },
     ],
-}, providerOpts);
\ No newline at end of file
+}, providerOpts);
+
+const vpcWithSpecificZones = new awsx.ec2.Vpc("custom7", {
+    requestedAvailabilityZones: ["us-east-1a","us-east-1b","us-east-1c"],
```
leezen, line 59: spacing

nodejs/examples/vpc/index.ts
```diff
@@ -53,4 +53,14 @@ const vpcWithLocations = new awsx.ec2.Vpc("custom6", {
...
+    subnets: [
+        { type: "public"},
+        { type: "private"},
+        { type: "isolated", name: "db"},
+        { type: "isolated", name: "redis"},
```
leezen, line 64: space before `}`

leezen (commented):
Thanks for the updates. Could you please add a `CHANGELOG` entry as well? I think if you merge `master` then the tests should pass as well and we can get this merged in.

nodejs/examples/vpc/index.ts
seivan: Done

nodejs/examples/vpc/index.ts
seivan: Done

nodejs/examples/vpc/index.ts
seivan: Done, did I miss that you have formatter rules of some sort? I had to disable my own on save since it would change everything, but a lot of these could be resolved if you introduced one that vscode would respect.

seivan:
@leezen Resolved review comments, updated changelog, and merged in master. Thanks for taking a look!

seivan:
@leezen Again for my own curiosity sake, I would like some clarification on how this works in regards to my [comments](https://github.com/pulumi/pulumi-awsx/pull/638#issuecomment-783644314)

When `getRequestedAvailabilityZones` gets called, it has the potential to return different zones or even in different order, that's not really on you, but on AWS, however I'm wondering if Pulumi does something to manage this internally? 

Like would it first sort the list of zones returned and check that the existing VPC was created on a different zone, thus throwing an error, or would it just mutate the VPC zones?

leezen:
Ah, looks like the test failed with `pulumi:providers:aws prov2 **creating failed** error: Duplicate resource URN 'urn:pulumi:vpc-55c77514::vpc::pulumi:providers:aws::prov2'; try giving it a unique name` -- which makes sense.

nodejs/examples/vpc/index.ts
```diff
@@ -53,4 +53,14 @@ const vpcWithLocations = new awsx.ec2.Vpc("custom6", {
...
+
+const vpcWithSpecificZones = new awsx.ec2.Vpc("custom7", {
+    requestedAvailabilityZones: ["us-east-1a", "us-east-1b", "us-east-1c"],
+    subnets: [
+        { type: "public" },
+        { type: "private" },
+        { type: "isolated", name: "db" },
+        { type: "isolated", name: "redis" },
+    ],
+}, { provider: new aws.Provider("prov2", { region: "us-east-1" }) });
```
leezen, line 66: I actually don't think there's a need to allocate a different provider here, so you should be able to just use `providerOpts` as with the other cases?

nodejs/examples/vpc/index.ts
```diff
@@ -55,12 +56,23 @@ const vpcWithLocations = new awsx.ec2.Vpc("custom6", {
...
-        { type: "isolated", name: "redis" },
-    ],
-}, { provider: new aws.Provider("prov2", { region: "us-east-1" }) });
+
+export = async () => {
+    const allZones =   await aws.getAvailabilityZones(/*args:*/ undefined, { provider: providerOpts.provider, async: true })
+
+    const selectedZones = new random.RandomShuffle("custom7_az", {
+        inputs: allZones.names,
+        resultCount: allZones.names.length / 2,
```
seivan, line 65: @leezen The documentation doesn't mention what happens if you don't get an integer here, I could floor/ceil it if necessary.

nodejs/examples/vpc/index.ts
```diff
@@ -55,12 +56,23 @@ const vpcWithLocations = new awsx.ec2.Vpc("custom6", {
...
-    requestedAvailabilityZones: ["us-east-1a", "us-east-1b", "us-east-1c"],
-    subnets: [
-        { type: "public" },
-        { type: "private" },
-        { type: "isolated", name: "db" },
-        { type: "isolated", name: "redis" },
-    ],
-}, { provider: new aws.Provider("prov2", { region: "us-east-1" }) });
+
+export = async () => {
```
seivan, line 60: @leezen Since I am using the existing provider, and I don't know where that's located as it's an ENV variable on the CI, I am fetching all the zones and that seems to be a promise, so I looked around and this was used in the ELB tests, so I assumed it'll be fine?

nodejs/examples/vpc/index.ts
seivan: @leezen 
Well ok, but I guess I gotta be asking for available zones and shuffle/pick since the tests are using env variables for region and I can't hard code them since those could change. 

Fixed in https://github.com/pulumi/pulumi-awsx/pull/638/commits/3eaa387353d1e216ad92afe18d426843ab1d62f8

nodejs/examples/vpc/index.ts
seivan: @leezen Actually, I simplified by using promises directly as arguments to the `random.Shuffle` in https://github.com/pulumi/pulumi-awsx/pull/638/commits/da138a4efea0bc1a5e028f16c18f1b7c7380d8c1 - let me know how you prefer it, or if you want me to go back to using hard coded region and zones  in a new custom provider with a different name - since (I am guessing) the name was the reason it failed?

nodejs/examples/vpc/index.ts
```diff
@@ -53,4 +54,24 @@ const vpcWithLocations = new awsx.ec2.Vpc("custom6", {
...
+const vpcWithSpecificZones = new awsx.ec2.Vpc("custom7", {
+    requestedAvailabilityZones: selectedZones,
+    subnets: [
+        { type: "public" },
+        { type: "private" },
+        { type: "isolated", name: "db" },
+        { type: "isolated", name: "redis" },
+    ],
+}, providerOpts);
+// }
```
leezen, line 76:
```suggestion
```

nodejs/examples/vpc/index.ts
leezen: Yeah, this makes sense if you want to just fetch the available zones and shuffle them in. That works for me.

leezen:
Thanks for the work on this. This is looking good. Assuming the tests pass and you remove that one extraneous comment line, I think we're good to merge.

nodejs/examples/vpc/index.ts
seivan: done

nodejs/awsx/ec2/vpc.ts
```diff
@@ -702,10 +711,15 @@ export interface VpcArgs {
     subnets?: VpcSubnetArgs[];
 
     /**
-     * The maximum number of availability zones to use in the current region.  Defaults to `2` if
-     * unspecified.  Use `"all"` to use all the availability zones in the current region.
+     * The names of the availability zones to use in the current region. Defaults to `2` if
+     * unspecified. Use `"all"` to use all the availability zones in the current region.
+     */
+    requestedAvailabilityZones?: number | "all" | [string, ...string[]];
```
leezen, line 717: The test is currently failing b/c it's passing in an `Output<string[]>` in when this doesn't accept that. I don't see an easy way to fix this except to change the test to not use `RandomShuffle` and use a different deterministic ordering instead, e.g. alphabetical.

seivan: @leezen 
I see a few ways: 

- Maybe accept `Output<string[]>` 
- Hardcode region as I did initially, this removes need for `RandomShuffle`
- Use await to get a `string[]` - I need help here to understand how to add that in the unit tests.

leezen:
@seivan 

> Maybe change so requestedAvailabilityZones accepts Input<string[]>, it's not a bad idea in itself.

I agree this is ideal. I think if there's a way to restructure your code to allow this, let's aim for this.

> Hardcode region and zones as I did initially, this removes need for RandomShuffle.
> Use await to get a string[] - I need help here to understand how to add that in the unit tests.

seivan:
@leezen 

Now I know, there is a private implementation to make that into a `Promise` and I could probably just cast to any and use that, but since there is no type checking involved, I'd prefer to just resort to manually create a `Promise`

So to summarize, I just return a `Promise`, which turns an `Input<T>` into an `Output<T>` using `pulumi.Output.create()`, and call `apply` to resolve the values/error. 



```ts
        return new Promise((resolve, reject) => {
            pulumi.Output.create(requestedZones).apply(requestedZones => { 
```


PS: I shadow the argument, to make sure there are no mistakes.

nodejs/awsx/ec2/vpc.ts
seivan: Added in https://github.com/pulumi/pulumi-awsx/pull/638/commits/c90df31b3497d709a044934377bb1c1d36401fbd

nodejs/awsx/ec2/vpc.ts
```diff
@@ -504,23 +504,31 @@ utils.Capture(Vpc.prototype).initializeVpcArgs.doNotCapture = true;
...
+
+    if (typeof requestedZones === "object") {
+        return new Promise((resolve, reject) => {
+            pulumi.Output.create(requestedZones).apply(requestedZones => {
+                const mappedZones = descriptions.filter(zone => requestedZones.includes(zone.name));
+                mappedZones.length === requestedZones.length ?
+                    resolve(mappedZones) :
+                    reject(new pulumi.ResourceError("Availability zones did not match requested zones", parent));
+            })
+        })
```
leezen, line 526: nit: missing semicolons (the linter in actions is running against master and not this branch)

nodejs/awsx/ec2/vpc.ts
```diff
@@ -504,23 +504,31 @@ utils.Capture(Vpc.prototype).initializeVpcArgs.doNotCapture = true;
 async function getAvailabilityZones(
         parent: pulumi.Resource | undefined,
         provider: pulumi.ProviderResource | undefined,
-        requestedCount: "all" | number | undefined): Promise<topology.AvailabilityZoneDescription[]> {
+        requestedZones: VpcArgs["requestedAvailabilityZones"] = 2): Promise<topology.AvailabilityZoneDescription[]> {
```
leezen, line 507: But this type is no longer valid now, right? Won't this return a `Promise<Output<AvailabilityZoneDescription[]>>` with the code you have below? I think you'll want to actually want this function to return Output<AvailabilityZoneDescription[]> and then plumb that through the rest of the code?

nodejs/awsx/ec2/vpc.ts
seivan: Heh, I had to disable my local formatter and you guys don't got a configuration supplied, I'd make a PR myself, but I am not sure what your specs are :)

nodejs/awsx/ec2/vpc.ts
seivan: > But this type is no longer valid now, right? Won't this return a `Promise<Output<AvailabilityZoneDescription[]>>` with the code you have below? 

Are you sure about this? Afaik the type of the `Promise<T>`is whatever is within the resolve, which is a `topology.AvailabilityZoneDescription[]`. 
That, and TS doesn't seem to complain, but I am not sure what goes on with `Output`. 
The type *should* be `Promise<topology.AvailabilityZoneDescription[]>` as the resolve callbacks with the list, we don't return the output itself, just the resolved value. 

>  I think you'll want to actually want this function to return `Output<AvailabilityZoneDescription[]>` and then plumb that through the rest of the code?

Sure, but that would introduce breaking changes, and I quite understand the plumbing between `Input/Promise/Output`, in a side note, I feel like `Output` should at least behave like a `PromiseLike<T>` but that's a different topic. 


Regardless, do you see an actual issue running the test suite or in your compiler? Because I don't see any errors here, and I don't see why there would be since the type would be `Promise<topology.AvailabilityZoneDescription[]>`

seivan: @leezen Could you run the tests, or give me the permissions to run them for the duration of this PR, so I could have some faster feedback loop.

seivan:
@leezen Could you run the tests, or give me the permissions to run them for the duration of this PR, so I could have some faster feedback loop.

leezen:
Sorry for not responding earlier.

Regarding linting: you should be able to run `make lint` locally.

Regarding testing: you can also run `AWS_REGION=xxx make test` locally (note this will incur some AWS charges) or run the actual command in the `Makefile`:

`cd nodejs/awsx && go test -v -count=1 -cover -timeout 2h -parallel ${TESTPARALLELISM} .` with an additional `-run TestName` to only run that test.

seivan:
> Regarding linting: you should be able to run `make lint` locally.

Thanks! 


> Regarding testing: you can also run `AWS_REGION=xxx make test` locally (note this will incur some AWS charges) or run the actual command in the `Makefile`:

Yeah, I am not comfortable running resources on my AWS account, not for charge reason but because I hate to clean up after. 


Anyway, I thought `Output<T>` would be acceptable as argument for something that expects `Input<T>`, but it seems like there are some issues there related to either that, or the rest types in a tuple. 
Any takes on this?

leezen:
Is there a reason it's `Input<[string, ...string[]]>` instead of just `Input<string[]>`? I think with the latter, there's no issue with having an `Output<string[]>`

seivan:
> Is there a reason it's `Input<[string, ...string[]]>` instead of just `Input<string[]>`?

To disallow empty lists, before you respond, keep in mind I was eventually planning on using template literal types for actual names and eventually tie that in with selected regions, heh. ;-) 


>  I think with the latter, there's no issue with having an `Output<string[]>`
Yeah so I looked at it, I could tweak it so it would work, but you know what, it seems to me that when relying on `Input<T>` & `Output<T>` types we can assure that its machine generated and be less strict. 

But for my case, I am actually typing them in manually and even using different ones per stack, so I'd like the strictness of it when typed by a bag of flesh.

seivan:
@leezen 

So I was thinking, is there a way in the test case to assert that the correct zones were selected in the subnets.
Or at the very least they are `z.length / 2`, do you got any ideas on what to use there?

leezen:
@seivan You could do a check along the lines of `allZones.then(azs => selectedZones.apply(z => assert(z.length == azs.length / 2)))` but I don't think this is necessary; as at this point, you're more or less testing the functionality of `RandomShuffle` plus the division, which doesn't really seem necessary to me? Or did you have a different test in mind? Looking at what's in the test, it seems adequate to me and I'm happy to merge this if you are.

leezen (approved):

nodejs/awsx/ec2/vpc.ts
```diff
@@ -702,10 +709,15 @@ export interface VpcArgs {
...
     /**
-     * The maximum number of availability zones to use in the current region.  Defaults to `2` if
-     * unspecified.  Use `"all"` to use all the availability zones in the current region.
+     * The names of the availability zones to use in the current region. Defaults to `2` if
+     * unspecified. Use `"all"` to use all the availability zones in the current region.
+     */
+    requestedAvailabilityZones?: number | "all" | [string, ...string[]] | pulumi.Input<string[]>;
+
+    /**
+     * @deprecated Use `requestedAvailabilityZones`
```
leezen, line 718: One last comment I realized after I clicked 'approve' -- I think we want to avoid `@deprecated` here and make this a comment to avoid annoying the vast majority of users who've provided this parameter.

seivan: @leezen Are you sure about this? 

I even think removing it instead of deprecating it would have been better. I mean this is one the benefits of using TS instead of Python and Pulumi should embrace these things instead of polluting the API.
The benefits is that the documentation and code becomes less dense and offers a path to upgrade your code.

Allow me to make the case that removing or even deprecating has tangible benefits. 
Think of it this way, if you neither remove or deprecate it, the users will *never* figure out that they're using deprecated or "outdated" implementations. The only other approach is for someone to examine each changelog for various libs and manage that themselves.

Deprecation allows us to warn that something has changed, and a compile error will tell them that the previous method was improper. 

Are you really sure you want me to remove it?

seivan:
>  But I don't think this is necessary; as at this point, you're more or less testing the functionality of `RandomShuffle` plus the division, which doesn't really seem necessary to me? 

No that wasn't my intention. 

> Or did you have a different test in mind? Looking at what's in the test, it seems adequate to me and I'm happy to merge this if you are.

We're not testing that this particular code behaves as intended. 

```ts
requestedZones.includes(zone.name));
                 mappedZones.length === requestedZones.length ?
                     resolve(mappedZones) :
                     reject(new pulumi.ResourceError("Availability zones did not match requested zones", parent));
```

Nor are we testing the conditions for `numer`
```ts
descriptions.slice(0, requestedZones === "all" ? descriptions.length : requestedZones)
```

So we should test that you're getting the specific zones you're requesting *and* test that you're getting the `number` of zones requested otherwise. 


Checking the length was the "minimum" test case :)

nodejs/awsx/ec2/vpc.ts
leezen: Agree with pretty much everything you wrote, except that there's no user benefit to changing from one property to the other. The functionality will remain the same. So, we end up burdening users with additional cognitive overhead for something that isn't materially 'wrong' (there's no security issue, no incorrect functionality, etc.). Given it's likely to be a widely used parameter, I'm inclined to not mark it `@deprecated` until we cut a major version, at which point, we can simply drop the parameter completely as a hard breaking change.

leezen:
> More specifically that we're filtering out the zones properly or throwing an error if the inputted zones don't match the specific region which would be a runtime error(!) until you use some template literal type magic to tie zones to regions, but that's a different topic.

> Nor are we testing the conditions for number and all

Ah -- yes. Thanks for clarifying this. In this case, you could add a couple test cases to `/nodejs/awsx/tests/ec2/ec2.spec.ts` by exporting `getAvailabilityZones` (and marking as `@internal`)

nodejs/awsx/ec2/vpc.ts
seivan: > So, we end up burdening users with additional cognitive overhead for something that isn't materially 'wrong' (there's no security issue, no incorrect functionality, etc.). 

I'm convinced, thanks, it makes sense! 

> Given it's likely to be a widely used parameter, I'm inclined to not mark it `@deprecated` until we cut a major version, at which point, we can simply drop the parameter completely as a hard breaking change.

I am not sure how the TS compiler works for straight up removing it in terms of offering an alternative, maybe add deprecated before released a new major version for a transition period, either way you're right.

leezen:
I went ahead and rebased this against `master` and removed the deprecation warning.

seivan:
> I went ahead and rebased this against `master` and removed the deprecation warning.

@leezen 

I just want to apologise, I completely dropped the ball on this.  I did however write the tests we discussed, but I haven't had a chance to actually run them. Do you still want them in this PR or a separate PR?

leezen:
@seivan Not a problem. Thank you for getting this in in the first place. If you'd like to follow up on those tests, happy to take another look in a separate PR.

Issue #639, opened 2021-02-22
AWSX API Gateway RestApiArgs does not support all properties of AWS RestApi


catmeme:
Crosswalk API Gateway provides a property `RestApiArgs` with the description, "Additional optional args that can be passed along to the RestApi created by the awsx.apigateway.API."  While trying to configure `disableExecuteApiEndpoint` under `RestApiArgs`, we realized only a subset of properties is supported when compared to `aws.apigateway.RestApi`.

* https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/awsx/apigateway/#RestApiArgs

* https://www.pulumi.com/docs/reference/pkg/aws/apigateway/restapi/#inputs

## Expected behavior

We would expect any property supported by `aws.apigateway.RestApi` would be supported by `awsx.apigateway.RestApiArgs`.

## Current behavior

Only a subset of properties are supported.

## Steps to reproduce

```
const app = new awsx.apigateway.API("api-gateway", {
  routes: [{
    path: "/{proxy+}",
    method: "POST",
    eventHandler: new aws.lambda.CallbackFunction("endpoint", {
      callback: routes.doThing,
    }),
  }],
  restApiArgs: {
    disableExecuteApiEndpoint: true,
  },
});
```

## Context (Environment)

A workaround is to add `// @ts-ignore` to any property you want to passthrough regardless of properties in `RestApiArgs`.

PR #640, opened 2021-02-22
Ensuring apigateway.RestApiArgs are up to date with the pulumi-aws library

stack72:
Fixes: #639

### DIFF
--- a/nodejs/awsx/apigateway/api.ts
+++ b/nodejs/awsx/apigateway/api.ts
@@ -406,38 +406,7 @@ export interface APIArgs {
  * Additional optional args that can be passed along to the RestApi created by the
  * awsx.apigateway.API.
  */
-export interface RestApiArgs {
-    /**
-     * The name of the REST API.  Defaults to the name of the awsx.apigateway.Api if unspecified.
-     */
-    name?: pulumi.Input<string>;
-
-    /**
-     * The list of binary media types supported by the RestApi. Defaults to `* / *` if unspecified.
-     */
-    binaryMediaTypes?: pulumi.Input<pulumi.Input<string>[]>;
-
-    /**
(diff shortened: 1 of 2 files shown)


leezen (commented):
Since we don't ever modify the interface itself or require it to be used in a different way, why not just replace the interface definition with `type RestApiArgs = aws.apigateway.RestApiArgs;`?

stack72:
> Since we don't ever modify the interface itself or require it to be used in a different way, why not just replace the interface definition with `type RestApiArgs = aws.apigateway.RestApiArgs;`?

More than happy to change to do that - updating the PR now

catmeme:
Thanks for the quick fix guys!  I'm curious if other other types in this library could use the same solution?

stack72:
> Thanks for the quick fix guys! I'm curious if other types in this library could use the same solution?

@catmeme there will be for sure - I will open an issue to see what we can change here

nodejs/awsx/apigateway/api.ts
```diff
@@ -402,43 +402,6 @@ export interface APIArgs {
     deploymentArgs?: DeploymentArgs;
 }
 
-/**
- * Additional optional args that can be passed along to the RestApi created by the
- * awsx.apigateway.API.
- */
-export interface RestApiArgs {
```
leezen, line 409: I actually wasn't suggesting we remove this type completely as that would be a breaking change (since this is an exported type), but I believe exporting this as an alias:

`export type RestApiArgs = aws.apigateway.RestApiArgs;` would be non-breaking since any external references to it should continue to work.

stack72: Way too sensible! :D Updated ;)

leezen (approved):

Issue #599, opened 2020-11-28
FargateService cannot specify both launchType and capacityProviderStrategies


kennyjwilli:
The `launchType` is currently [explicitly set to `FARGATE`](https://github.com/pulumi/pulumi-awsx/blob/f4a7719f6354e9933be3c63de105c5a8b641462b/nodejs/awsx/ecs/fargateService.ts#L220). When using `capacityProviderStrategies`, you cannot also specify a `launchType`. If you try to, `pulumi up` will fail with this error. 

```
InvalidParameterException: Specifying both a launch type and capacity provider strategy is not supported. Remove one and try again.
```

Here is a minimal repro. 

```typescript
import * as awsx from "@pulumi/awsx";

new awsx.ecs.Cluster("cluster", {
    capacityProviders: ["FARGATE", "FARGATE_SPOT"],
    defaultCapacityProviderStrategies: [{
        base: 0,
        capacityProvider: "FARGATE",
        weight: 0
    }, {
        base: 0,
        capacityProvider: "FARGATE_SPOT",
        weight: 1
    }],
})

new awsx.ecs.FargateService("service", {
    desiredCount: 1,
    taskDefinitionArgs: {
        containers: {
            nginx: {
                image: "nginx:latest",
                memory: 512
            }
        }
    },
    capacityProviderStrategies: [{
        capacityProvider: "FARGATE_SPOT",
        weight: 1
    }]
})
```

Explicitly specifying the `launchType` also prevents the cluster default `capacityProviderStrategy` from being applied. 

From the [RunTask API doc on capacityProviderStrategy](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html#API_RunTask_RequestParameters): 
> If a `capacityProviderStrategy` is specified, the `launchType` parameter must be omitted. If no `capacityProviderStrategy` or `launchType` is specified, the `defaultCapacityProviderStrategy` for the cluster is used.

To fix this, I suggest allowing the `launchType` parameter to be passed in to the `FargateService`.

kennyjwilli:
service.ts applies a [default value](https://github.com/pulumi/pulumi-awsx/blob/f4a7719f6354e9933be3c63de105c5a8b641462b/nodejs/awsx/ecs/service.ts#L57) when no `launchType` is specified. That will need to be fixed too.

stack72:
Hey @kennyjwilli 

[...]

PR #643, opened 2021-02-24
fix: Make it possible to use capacity provider strategy for ecs services

danieldspx:
This PR fixes pulumi/pulumi-awsx#599 and should not break existing code. I have tested it and it works like a charm but I would suggest you guys to test it to double check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ CHANGELOG
 ## HEAD (Unreleased)
 _(none)_
 
+## 0.25.1 (2021-02-24)
+
+* Make it possible to use capacity provider strategy for `ecs.FargateService` and `ecs.Ec2Service`
+
 ## 0.25.0 (2021-02-12)
 
 * Allow passing of `forceNewDeployment` to `ecs.FargateService` and `ecs.Ec2Service`
--- a/nodejs/awsx/ecs/ec2Service.ts
+++ b/nodejs/awsx/ecs/ec2Service.ts
@@ -85,11 +85,18 @@ export class EC2Service extends ecs.Service {
             cluster.vpc, name, args.securityGroups || cluster.securityGroups, opts) || [];
         const subnets = args.subnets || cluster.vpc.publicSubnetIds;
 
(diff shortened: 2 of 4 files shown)


mukaschultze (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ CHANGELOG
 ## HEAD (Unreleased)
 _(none)_
 
+## 0.25.1 (2021-02-24)
```
leezen, line 7: You should include the changelog item under `HEAD (Unreleased)`. When we cut a tagged version, we'll update the changelog to include all the unreleased items under a version heading.

nodejs/awsx/ecs/ec2Service.ts
```diff
@@ -85,11 +85,18 @@ export class EC2Service extends ecs.Service {
             cluster.vpc, name, args.securityGroups || cluster.securityGroups, opts) || [];
         const subnets = args.subnets || cluster.vpc.publicSubnetIds;
 
+        const useClusterDefaultCapacityProviderStrategies = !!args.useClusterDefaultCapacityProviderStrategies;
```
leezen, line 88: In https://github.com/pulumi/pulumi-awsx/issues/599#issuecomment-768688605 I wasn't really in favor of going down this approach as it feels like it's confusing to be able to specify `useClusterDefaultCapacityProviderStrategies` on _either_ `EC2Service` or `FargateService` with  effectively end up with the same behavior (not to mention the name `EC2Service` is now inappropriate because the underlying capacity provider may not actually be EC2). Any reason you didn't want to model it as a separate class as discussed in the comment?

danieldspx: Yeap, I will comment in #599 to keep the conversation there. But I do agree with you.

danieldspx: Now I see your point and it does make sense. I will create another PR with this. I will close this one since I no longer think this is the best option.

danieldspx:
Closing this because I think that #644 solves the problem better.

danieldspx:
Closing this because I think that #644 solves the problem better.

PR #644, opened 2021-02-25
Create CapacityProviderService to make it possible to use capacity provider strategies

### LINKED ISSUE #599: FargateService cannot specify both launchType and capacityProviderStrategies

danieldspx:
From #599 I  modified @leezen  class name a bit and called it `CapacityProviderService` and you pass the taskDefinition that would either be a FargateTaskDefinition or EC2TaskDefinition. With that, the user would know exactly wich path he is going. Now underlying capacity provider can be anything and we can still leverage the cluster default capacity provider strategy without any extra flag.

Note that I've implemented a different idea before (#643) but as @leezen said, It makes more sense to have a different class for this. Also, from now on you can't define `capacityProviderStrategies` in `FargateService` nor `EC2Service`. In that way, this PR fixes #599.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+* Create `CapacityProviderService` to make it possible to use capacity provider strategies solving [#599](https://github.com/pulumi/pulumi-awsx/issues/599).
 
 ## 0.25.0 (2021-02-12)
 
--- a/nodejs/awsx/aws_test.go
+++ b/nodejs/awsx/aws_test.go
@@ -55,6 +55,15 @@ func TestAccCapacityProviderServiceWithClusterDefaultStrategies(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccCapacityProviderServiceEC2CustomStrategy(t *testing.T) {
+	test := getBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: path.Join(getCwd(t), "../examples/ecs/capacity-provider-service-ec2-custom-strategy"),
+		})
+
+	integration.ProgramTest(t, &test)
+}
+
 func TestAccTaskDefinitionWithProxyConfiguration(t *testing.T) {
 	test := getBaseOptions(t).
 		With(integration.ProgramTestOptions{
--- a/nodejs/awsx/aws_test.go
+++ b/nodejs/awsx/aws_test.go
@@ -46,6 +46,15 @@ func TestAccCluster(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccCapacityProviderServiceWithClusterDefaultStrategies(t *testing.T) {
+	test := getBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: path.Join(getCwd(t), "../examples/ecs/capacity-provider-service-with-cluster-default-strategies"),
+		})
+
+	integration.ProgramTest(t, &test)
+}
+
 func TestAccTaskDefinitionWithProxyConfiguration(t *testing.T) {
 	test := getBaseOptions(t).
 		With(integration.ProgramTestOptions{
--- a/nodejs/awsx/ecs/capacityProviderService.ts
+++ b/nodejs/awsx/ecs/capacityProviderService.ts
@@ -1,0 +1,253 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+import * as ecs from ".";
+import * as x from "..";
+import * as utils from "../utils";
+
+
+
+export class CapacityProviderService extends ecs.Service {
+    public readonly taskDefinition: ecs.FargateTaskDefinition | ecs.EC2TaskDefinition;
+
+    constructor(name: string,
+                args: CapacityProviderServiceArgs,
+                opts: pulumi.ComponentResourceOptions = {}) {
+
+        if (!args.taskDefinition) {
+            throw new Error("The [taskDefinition] must be provided");
+        }
+
+        const cluster = args.cluster || x.ecs.Cluster.getDefault();
+
+        const taskDefinition = args.taskDefinition;
+
+
+        const securityGroups = x.ec2.getSecurityGroups(
+            cluster.vpc, name, args.securityGroups || cluster.securityGroups, opts) || [];
+
+        let assignPublicIp, networkConfiguration, subnets: pulumi.Input<pulumi.Input<string>[]>;
+
+        if (taskDefinition instanceof ecs.FargateTaskDefinition) {
+            assignPublicIp = utils.ifUndefined(args.assignPublicIp, true);
+            subnets = ecs._getSubnets(cluster, args.subnets, assignPublicIp);
+            networkConfiguration = {
+                subnets,
+                assignPublicIp,
+                securityGroups: securityGroups.map(g => g.id),
+            };
+        } else {
+            assignPublicIp = false;
+            subnets = args.subnets || cluster.vpc.publicSubnetIds;
+            networkConfiguration = taskDefinition.taskDefinition.networkMode.apply(n => {
+                // The network configuration for the service. This parameter is required for task
+                // definitions that use the `awsvpc` network mode to receive their own Elastic
+                // Network Interface, and it is not supported for other network modes.
+                if (n !== "awsvpc") {
+                    return undefined!;
+                }
+
+                return {
+                    subnets,
+                    assignPublicIp: false,
+                    securityGroups: securityGroups.map(g => g.id),
+                };
+            });
+        }
+
+        super("awsx:x:ecs:CapacityProviderService", name, {
+            ...args,
+            taskDefinition,
+            securityGroups,
+            networkConfiguration,
+        }, opts);
+
+        this.taskDefinition = taskDefinition;
+
+        this.registerOutputs();
+    }
+}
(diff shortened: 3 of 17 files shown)


leezen (commented):
Thanks for doing this! Overall, looks good. If you could please add a test or two (see `nodejs/awsx/aws_test.go` for how this works), that'd be great so we can make sure this is working properly. I also added a comment about keeping `getSubnets` internal to the module.

danieldspx:
@leezen I guess that for EC2 capacity provider startegies I should create an autoscaling group, and with that create a capacity provider in the cluster and then use it in the CapacityProviderService . But I could not come up with a test easyli for this. I am more used with Fargate. Could you help me? I could not find any pulumi example for autoscaling group usage.

leezen:
https://www.pulumi.com/docs/guides/crosswalk/aws/ecs/#creating-an-auto-scaling-group-for-ecs-cluster-instances has an example of doing this

danieldspx:
The tests failed so I tryed to `pulumi up` the same code from test and I got the same error message. And I discovered that I had not passed the `launchType` on service creation. This would lead the default launch type to be used ("EC2") and we know that `assignPublicIp = true` (which is true for Fargate in this example) would cause that error message. Now that is fixed.

danieldspx:
The problem was the the ecs.Service was setting `EC2` as the launch type if `args.launchType` was undefined. I've tested both and they work fine.

For the EC2 I got an `service (service-name) was unable to place a task because no container instance met all of its requirements. Reason: No Container Instances were found in your capacity provider.` and the reason is clear, but I did not want to creat any instances to avoid costs. I think both tests now should pass :smile:

danieldspx:
Error now was with the custom strategy: `'One or more security groups in the launch configuration are not linked to the VPCs configured in the Auto Scaling group`. I am not so sure if the change I've made will fix the problem.

There is still the option to pass the Cluster's security groups to the launch configuration args. If you think I should do it, just tell me.
```ts
...
const asg = cluster.createAutoScalingGroup("asg", {
  ...
  launchConfigurationArgs: { instanceType: "t2.micro", securityGroups: cluster.securityGroups },
}, providerOpts);
...

**Comment by leezen** [comment]
`createAutoscalingGroups` defaults to passing in the cluster's groups (https://github.com/pulumi/pulumi-awsx/blob/ea0c2f3afd95dcf47c5bac187843e7b66418ff9d/nodejs/awsx/ecs/cluster.ts#L85), so I don't think that's the issue? I dug a little bit into this, but nothing stands out. I'll try to spend more time tomorrow or Monday or if you are able to try taking a look, that'd be appreciated.

**Comment by leezen** [comment]
Oh wait, I see... you have a commit since the test ran to pass in the default VPC to the cluster. Yes, that should make the difference because in the previous ordering the cluster will create its own VPC, which is different from the default. Let me try re-running the tests.

**Comment by leezen** [comment]
To make this work, I think we need to unwind some dependencies around the `CapacityProvider` which has to exist before the `Cluster` (because there's no way to modify the cluster w/ the `CapacityProvider` after the fact in a single shot program). But, the `CapacityProvider` requires the `AutoScalingGroup` to exist, which in turns relies on having the `Cluster` identifier so that the boot parameters can correctly register instances. Similarly, the ASG relies on sharing a security group with the `Cluster`.

In order to unwind these dependencies, we can set the cluster identifier and security groups upfront, which results in the following:

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as random from "@pulumi/random";

const config = new pulumi.Config("aws");
const providerOpts = { provider: new aws.Provider("prov", { region: <aws.Region>config.require("envRegion") }) };

const vpc = awsx.ec2.Vpc.getDefault(providerOpts);

const clusterName = "cluster";
const clusterIdentifier = new random.RandomString("cluster", {
  special: false,
  upper: false,
  length: 6,
}).result.apply(s => "cluster-" + s);

// Create the cluster's security group ahead of time so we can use it for the ASG.
const clusterSecurityGroup = awsx.ecs.Cluster.createDefaultSecurityGroup(clusterName, vpc, providerOpts);

const userData: awsx.autoscaling.AutoScalingUserData = {
  extraBootcmdLines: () => clusterIdentifier.apply(clusterId =>
    [{ contents: `- echo ECS_CLUSTER='${clusterId}' >> /etc/ecs/ecs.config` }]),
};

const asg = new awsx.autoscaling.AutoScalingGroup("asg", {
  vpc,
  subnetIds: vpc.publicSubnetIds,
  templateParameters: { minSize: 1 },
  launchConfigurationArgs: {
    instanceType: "t2.micro",
    securityGroups: [clusterSecurityGroup],
    userData,
  },
}, providerOpts);

const cp = new aws.ecs.CapacityProvider("capacity-provider", {
  autoScalingGroupProvider: {
    autoScalingGroupArn: asg.group.arn,
  },
}, providerOpts);

const cluster = new awsx.ecs.Cluster(clusterName, {
  name: clusterIdentifier,
  vpc,
  capacityProviders: [ cp.name ],
  securityGroups: [clusterSecurityGroup],
}, providerOpts);
```

I've swapped this in for the relevant parts of what's currently committed to this PR, tested it, and it works.

danieldspx:
> To make this work, I think we need to unwind some dependencies around the `CapacityProvider` which has to exist before the `Cluster` (because there's no way to modify the cluster w/ the `CapacityProvider` after the fact in a single shot program). But, the `CapacityProvider` requires the `AutoScalingGroup` to exist, which in turns relies on having the `Cluster` identifier so that the boot parameters can correctly register instances. Similarly, the ASG relies on sharing a security group with the `Cluster`.
> 
> In order to unwind these dependencies, we can set the cluster identifier and security groups upfront, which results in the following:
> 
> ```ts
> import * as aws from "@pulumi/aws";
> import * as awsx from "@pulumi/awsx";
> import * as pulumi from "@pulumi/pulumi";
> import * as random from "@pulumi/random";
> 
> const config = new pulumi.Config("aws");
> const providerOpts = { provider: new aws.Provider("prov", { region: <aws.Region>config.require("envRegion") }) };
> 
> const vpc = awsx.ec2.Vpc.getDefault(providerOpts);
> 
> const clusterName = "cluster";
> const clusterIdentifier = new random.RandomString("cluster", {
>   special: false,
>   upper: false,
>   length: 6,
> }).result.apply(s => "cluster-" + s);
> 
> // Create the cluster's security group ahead of time so we can use it for the ASG.
> const clusterSecurityGroup = awsx.ecs.Cluster.createDefaultSecurityGroup(clusterName, vpc, providerOpts);
> 
> const userData: awsx.autoscaling.AutoScalingUserData = {
>   extraBootcmdLines: () => clusterIdentifier.apply(clusterId =>
>     [{ contents: `- echo ECS_CLUSTER='${clusterId}' >> /etc/ecs/ecs.config` }]),
> };
> 
> const asg = new awsx.autoscaling.AutoScalingGroup("asg", {
>   vpc,
>   subnetIds: vpc.publicSubnetIds,
>   templateParameters: { minSize: 1 },
>   launchConfigurationArgs: {
>     instanceType: "t2.micro",
>     securityGroups: [clusterSecurityGroup],
>     userData,
>   },
> }, providerOpts);
> 
> const cp = new aws.ecs.CapacityProvider("capacity-provider", {
>   autoScalingGroupProvider: {
>     autoScalingGroupArn: asg.group.arn,
>   },
> }, providerOpts);
> 
> const cluster = new awsx.ecs.Cluster(clusterName, {
>   name: clusterIdentifier,
>   vpc,
>   capacityProviders: [ cp.name ],
>   securityGroups: [clusterSecurityGroup],
> }, providerOpts);
> ```
> 
> I've swapped this in for the relevant parts of what's currently committed to this PR, tested it, and it works.

Awesome work. I did the changes you suggested and hope this finally works :pray:

bbstilson:
I can't seem to find any documentation of this feature. The only reason I stumbled across it is because my IDE autocompleted. Is there a reason the [awsx/ecs](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/awsx/ecs/) module documentation hasn't been updated? Cheers

danieldspx:
> I can't seem to find any documentation of this feature. The only reason I stumbled across it is because my IDE autocompleted. Is there a reason the [awsx/ecs](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/awsx/ecs/) module documentation hasn't been updated? Cheers

Also could not find it. If I remember correctly it was on the docs before and now it is gone. Do you have any idea why @leezen ?

@bbstilson  if you really want to know a bit more about it you can refer to [this article](https://medium.com/@danieldspx/how-to-set-up-your-aws-ecs-cluster-using-pulumi-2340041dc197) I wrote (specifically in the point 4 of the article).

leezen:
Unfortunately, I'm not sure this was actually ever correctly updated (see commit history for https://github.com/pulumi/docs/tree/master/content/docs/reference/pkg/nodejs/pulumi/awsx -- it seems the last time we released, the workflow for updating the docs didn't complete correctly -- https://github.com/pulumi/docs/runs/3750842256?check_suite_focus=true).

leezen:
I've opened https://github.com/pulumi/docs/pull/7077 to correct this

bbstilson:
Thanks for the prompt response @leezen  and @danieldspx ! I appreciate it.

PR #649, opened 2021-03-02
Fix the copyright header for linting

leezen:
@stack72 Any chance you know what's going on here? Even locally, with `2016-2018` the linter is fine. With `2016-2021` or even just plain `2021` the linter complains.


seivan:
@leezen Is this related to the linting test breaking in https://github.com/pulumi/pulumi-awsx/pull/638 ?

PR #655, opened 2021-03-26
Add additional StorageTypes to awsx.s3.metrics

catmeme:
To use the S3 `numberOfObjects` metric, I needed the `storageType: "AllStorageTypes"`.  When I referenced the AWS documentation, I realized many other storageTypes were missing: https://docs.aws.amazon.com/AmazonS3/latest/userguide/metrics-dimensions.html#s3-cloudwatch-dimensions

This PR adds the most up-to-date list of storageTypes for S3 metrics.

I'm not sure what to do about lines `149-174` in `nodejs/awsx/s3/metrics.ts`, which repeats storage types, but I wasn't sure if that that was intended to be a limited subset of options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+* Add additional StorageTypes to `awsx.s3.metrics`
 
 ## 0.26 (2021-03-25)
 
--- a/nodejs/awsx/s3/metrics.ts
+++ b/nodejs/awsx/s3/metrics.ts
@@ -37,31 +37,74 @@ export namespace metrics {
          * of storage:
          *
          * * StandardStorage - The number of bytes used for objects in the STANDARD storage class.
+         * * IntelligentTieringAAStorage - The number of bytes used for objects in the Archive
+         *   Access tier of INTELLIGENT_TIERING storage class.
+         * * IntelligentTieringDAAStorage - The number of bytes used for objects in the Deep
+         *   Archive Access tier of INTELLIGENT_TIERING storage class.
          * * IntelligentTieringFAStorage - The number of bytes used for objects in the Frequent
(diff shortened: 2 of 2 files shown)


leezen:
Thanks for submitting! If you don't mind, would you be able to open a PR in `pulumi-aws` to add this as an enum (see https://github.com/pulumi/pulumi-aws/pull/1196/files for an example -- you should only need to modify `schema.json` and the other files in the PR are automatically generated as part of `make generate`). Then, we can just plumb that enum through to here.

catmeme:
Maybe I'm being dense, but it looks like CloudWatch Metrics are contained exclusively within the `pulumi-awsx` package.   The `storageType` dimension is an enumeration of values for a CloudWatch MetricChange on an S3 Resource.

https://www.pulumi.com/docs/reference/pkg/aws/cloudwatch/

[awsx/s3:S3MetricChange](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/awsx/s3/#S3MetricChange) (which contains the `storageType` dimension property) extends [awsx/cloudwatch:Metric](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/awsx/cloudwatch/#Metric)

In `pulumi-aws`, there is a similar property in s3 supporting types, `storageClass`, but those aren't enums either: 

https://www.pulumi.com/docs/reference/pkg/aws/s3/bucket/#bucketlifecyclerulenoncurrentversiontransition
https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/s3/bucketObject.ts#L197-L201

Also, there is no `make generate` command in the `pulumi-aws` package.

leezen:
@catmeme No, you are not being dense. My apologies as I totally misread this as storage types and not metrics. So I think what you have here is fine.

leezen (approved):

leezen:
Tests are failing for the same reason as in `master` -- since we're just updating the enums, I'm going to go ahead and merge.

PR #678, opened 2021-06-20
Allow constructing Vpc with no args object

lukehoban:
Since all input properties are optional, make the `args` bag itself optional, to align with the code-generation for custom resources with all-optional args bags.

This regressed in https://github.com/pulumi/pulumi-awsx/pull/470.

Because the test infrastructure in this repo currently requires passing explict providers, which means passing an options bag, there is unfortunately no simple way to test this directly.


jkodroff:
@lukehoban I think with the MLC version of AWSX, this PR is probably obsolete. Would you agree?

PR #710, opened 2021-10-08
Refactor API implementation

lukehoban:
This pulls out the RestAPI resource creation into a function that can be reused in other packages to construct the desired child resource directly parented to another resource.  This is expected to be used in exposing this as a multi-language component.

### DIFF
--- a/nodejs/awsx/package.json
+++ b/nodejs/awsx/package.json
@@ -19,7 +19,7 @@
         "mime": "^2.0.0"
     },
     "peerDependencies": {
-        "@pulumi/aws": "^4.6.0",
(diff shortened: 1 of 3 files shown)


nodejs/awsx/package.json
```diff
@@ -19,7 +19,7 @@
         "mime": "^2.0.0"
     },
     "peerDependencies": {
-        "@pulumi/aws": "^4.6.0",
+        "@pulumi/aws": "~4.23.0",
```
lukehoban, line 22: This change works around https://github.com/pulumi/pulumi-aws/issues/1660.  It should be removed before merging if that issue is fixed.

Issue #289, opened 2019-05-30
`cluster.createAutoScalingGroup` fails with "Property AvailabilityZones cannot be empty."


lukehoban:
Following docs at https://github.com/pulumi/pulumi-awsx/tree/master/nodejs/awsx/ecs#clusters:

```typescript
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";

const vpc = // ... create custom vpc
const cluster = new awsx.ecs.Cluster("custom", { vpc });

const asg = cluster.createAutoScalingGroup("custom", {
    templateParameters: { minSize: 20 },
    launchConfigurationArgs: { instanceType: "t2.medium" },
});
```

Leads to:

```
error: Plan apply failed: 1 error occurred:
        * creating urn:pulumi:fib-dev-ecs::fib::awsx:x:ecs:Cluster$awsx:x:autoscaling:AutoScalingGroup$aws:cloudformation/stack:Stack::app-asg: ROLLBACK_COMPLETE: ["The following resource(s) failed to create: [Instances]. . Rollback requested by user." "Property AvailabilityZones cannot be empty."]
```

And indeed, the outputted CloudFormation template is missing `AvailabilityZones`:

```yaml
    AWSTemplateFormatVersion: '2010-09-09'
    Outputs:
        Instances:
            Value: !Ref Instances
    Resources:
        Instances:
            Type: AWS::AutoScaling::AutoScalingGroup
            Properties:
                Cooldown: 300
                DesiredCapacity: 2
                HealthCheckGracePeriod: 120
                HealthCheckType: EC2
                LaunchConfigurationName: "app-asg-1ff1dbd"
                MaxSize: 100
                MetricsCollection:
                -   Granularity: 1Minute
                MinSize: 2
                VPCZoneIdentifier: []
                Tags:
                -   Key: Name
                    Value: app-asg
                    PropagateAtLaunch: true
            CreationPolicy:
                ResourceSignal:
                    Count: 2
                    Timeout: PT15M
            UpdatePolicy:
                AutoScalingRollingUpdate:
                    MaxBatchSize: 1
                    MinInstancesInService: 2
                    PauseTime: PT15M
                    SuspendProcesses:
                    -   ScheduledActions
                    WaitOnResourceSignals: true
```

lukehoban:
Okay - so the issue isn't actually `AvailabilityZones`.  The CF docs say:

```
AvailabilityZones
[...]

PR #713, opened 2021-11-05
Patch awsx.Cluster.createAutoScalingGroup default subnetIds

Lepozepo:
Closes https://github.com/pulumi/pulumi-awsx/issues/289

### DIFF
--- a/nodejs/awsx/ecs/cluster.ts
+++ b/nodejs/awsx/ecs/cluster.ts
@@ -80,6 +80,7 @@ export class Cluster
             opts: pulumi.ComponentResourceOptions = {}) {
 
         args.vpc = args.vpc || this.vpc;
+        args.subnetIds = args.subnetIds || this.vpc.publicSubnetIds;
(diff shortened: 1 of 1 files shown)


dino-rodriguez:
@danielrbradley Can we get a review on this? Still an issue 1 yr later.

Lepozepo:
+1 may need to recreate this though ><

danielrbradley:
Sorry, for loosing this in the noise of notifications. This whole component has been heavily reworked in preparation for a v1.0.0 release. Just rediscovered this while going through the backlog

Looked at quickly pulling this in, but I'm not confident that the public subnets are always a safe default. Within the Autoscaling implementation we attempt to fall-back to the privateSubnetIds from the vpc, although this also has pitfalls because the default VPC won't have any private subnets so will fail with an error. I think at this point we'll need to leave the implementation as-is given this isn't a blocking issue because it's possible to just set the subnetIds within the args when calling this method.

PR #733, opened 2022-03-03
Initial changes to allow MLC components to be used

viveklak:
deploy


nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -40,7 +40,7 @@ export class Trail extends pulumi.ComponentResource {
      * @param opts A bag of options that control this resource's behavior.
      */
     constructor(name: string, args: TrailArgs, opts?: pulumi.CustomResourceOptions) {
-        super("aws:cloudtrail:x:Trail", name, {}, opts);
+        super("awsx:cloudtrail:Trail", name, {}, opts);
```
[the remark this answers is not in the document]
justinvp: Add a type alias for the old type to opts?

nodejs/awsx/cmd/provider/index.ts
```diff
@@ -0,0 +1,142 @@
...
+class Provider implements pulumi.provider.Provider {
+    // A map of types to provider factories. Calling a factory may return a new instance each
+    // time or return the same provider instance.
+    private readonly typeToProviderFactoryMap: Record<string, () => pulumi.provider.Provider> = {
+        "awsx:cloudtrail:Trail": trailProviderFactory,
+    };
+
+    constructor(readonly version: string, readonly schema: string) {
+        // Register any resources that can come back as resource references that need to be rehydrated.
+        pulumi.runtime.registerResourceModule("eks", "index", {
```
justinvp, line 29:
```suggestion
        pulumi.runtime.registerResourceModule("awsx", "cloudtrail", {
```

nodejs/awsx/cmd/provider/index.ts
```diff
@@ -0,0 +1,142 @@
...
+                }
+            },
+        });
+    }
+
+    async call(token: string, inputs: pulumi.Inputs): Promise<pulumi.provider.InvokeResult> {
+        throw new Error(`unknown method ${token}`);
+    }
+
+    check(urn: pulumi.URN, olds: any, news: any): Promise<pulumi.provider.CheckResult> {
```
justinvp, line 46: The reason this has all the provider methods (e.g. check, diff, create, read, update, delete, etc.) in EKS is because EKS used to have some dynamic provider resources which don't work from MLCs, so they were changed to be implemented as regular custom resources in the EKS provider using these methods. But it doesn't look like awsx has any dynamic provider resources, so I suspect you can clean this up considerably here, deleting all the methods aside from `construct`.

viveklak: Thsnks! I was wondering the same. Good to know.

nodejs/awsx/cloudtrail/trail.ts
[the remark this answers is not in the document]
viveklak: Ah of course... Added.

danielrbradley (commented):
This looks good - keen to get it in!

Do we need the alias in both the SDK _and_ the implementation constructor? Could we do the alias in just the call to `super(...)` and remove from the SDK?

examples/go-trail/main.go
```diff
@@ -0,0 +1,67 @@
...
+							"Effect": "Allow",
+							"Principal": {"Service": "cloudtrail.amazonaws.com"},
+							"Action": "s3:PutObject",
+							"Resource": "arn:aws:s3:::%s/prefix/AWSLogs/%s/*",
+							"Condition": {"StringEquals": {"s3:x-amz-acl": "bucket-owner-full-control"}}
+						}
+					]
+				}`, args[0], args[0], args[1]))
+			}),
+		})
```
danielrbradley, line 47: This might be out of scope (at least for now) but perhaps this bucket policy should be able to be created automatically within the CloudTrail component?

viveklak: Yes, definitely,. This PR just sticks to the current POC logtrail component - no functional changes.

nodejs/awsx/cloudtrail/trail.ts
```diff
@@ -40,7 +40,7 @@ export class Trail extends pulumi.ComponentResource {
      * @param opts A bag of options that control this resource's behavior.
      */
     constructor(name: string, args: TrailArgs, opts?: pulumi.CustomResourceOptions) {
-        super("aws:cloudtrail:x:Trail", name, {}, opts);
+        super("awsx:cloudtrail:Trail", name, {aliases: [{type: pulumi.output("aws:cloudtrail:x:Trail")}]}, opts);
```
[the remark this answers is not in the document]
viveklak: Good catch, fixed.

viveklak:
> 

Removed the redundant aliases from the schema.

danielrbradley (approved):

danielrbradley:
Merged as part of #754

PR #747, opened 2022-03-31
Adding the v0.40.0 changelog entry

stack72:


stack72:
Self merging as this is only a CHANGELOG entry

PR #759, opened 2022-04-13
Adding the ability to check the schema between PRs

stack72:


stack72:
Need to do some work on schema tools to allow custom schema paths

PR #769, opened 2022-04-20
Implement Load Balancer MLC

### LINKED ISSUE #741: Lb MLC Components
We should take the existing LoadBalancing component as part of this package and make sure it works across multiple languages

- [x] Application Load Balancer
- [x] Network Load Balancer
- [x] TargetGroupAttachment
- [ ] Option to bring your own LB
- [ ] Prove compatibility with classic

danielrbradley:
Closes #741


stack72 (approved):
:shipit:

PR #780, opened 2022-04-25
ECR multi-lang component

danielrbradley:
- Keep build and push image as an invoke as there's no underlying resource for images.
- Remove implicit creation of repository within building of the image.
- Keep options for invoking 'build and push' via repository component or directly with zero coupling.


stack72 (approved):
<3

danielrbradley:
Tests consistently failing in CI and passing locally. Issue seems to be related to the repositoryUrl being undefined after the repository is created. Need to identify why CI resolves the repositoryUrl to undefined. 

It looks like this has been experienced by other users too: https://github.com/pulumi/pulumi/issues/6665#issuecomment-811173269

PR #787, opened 2022-04-27
Setting sane networking defaults on LB ApplicationLoadBalancer and FargateEcsService

stack72:


awsx/ecs/ec2Service.ts
```diff
@@ -79,7 +79,7 @@ export class EC2Service extends schema.EC2Service {
                     args.loadBalancers ?? taskDefinition?.loadBalancers,
                 waitForSteadyState: !utils.ifUndefined(
                     args.continueBeforeSteadyState,
-                    false,
+                    true,
```
danielrbradley, line 82: Actually, is this correct ... there's a `!` before the ifUndefined

If continue not defined, we use true, so wait is false.
I think this was actually already correct: if continue not defined, use false, so wait is true.

Not very clear either way 🤨

awsx/ecs/fargateService.ts
```diff
@@ -79,7 +84,7 @@ export class FargateService extends schema.FargateService {
                     args.loadBalancers ?? taskDefinition?.loadBalancers,
                 waitForSteadyState: !utils.ifUndefined(
                     args.continueBeforeSteadyState,
-                    false,
+                    true,
```
danielrbradley, line 87: Same as above

awsx/ecs/ec2Service.ts
stack72: fixed!

awsx/ecs/fargateService.ts
stack72: fixed!

PR #793, opened 2022-04-28
Fix tslint warning for obsolete setting.

jkodroff:


awsx-classic/tslint.json
```diff
@@ -28,7 +28,41 @@
         "member-access": false,
         "member-ordering": [
             true,
```
danielrbradley, line 30: Could we just remove`member-ordering` alltogether. In my opinion it adds nothing but pain. And also remove`ordered-imports` while we're there too?

jkodroff: Opening a new PR.

PR #796, opened 2022-04-28
Auto-sync go SDK dependency

danielrbradley:
Means all package dependencies are auto-synced from the provider package spec.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,9 +65,11 @@ build_python:: schema
 		cd ./bin && python3 setup.py build sdist
 
 build_go:: VERSION := $(shell pulumictl get version --language generic)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,9 +65,11 @@ build_python:: schema
 		cd ./bin && python3 setup.py build sdist
 
 build_go:: VERSION := $(shell pulumictl get version --language generic)
+build_go:: AWS_VERSION := $(shell node -e 'console.log(require("./awsx/package.json").dependencies["@pulumi/aws"])')
```
danielrbradley, line 68: Probably could've used jq, but couldn't remember if that's available on CI already 🤷

stack72: Oh nice!!

stack72 (approved):

PR #799, opened 2022-04-28
Remove default security group output on NLB

danielrbradley:
Was getting removed from inputs rather than outputs (was never added to inputs).


stack72 (approved):
great find! thank you

PR #804, opened 2022-04-29
Dev docs

danielrbradley:


Makefile
```diff
@@ -11,6 +11,8 @@ WORKING_DIR     := $(shell pwd)
 
 build:: provider build_nodejs build_python build_go build_dotnet
 
+sdks:: schema build_nodejs build_python build_go build_dotnet
```
jkodroff, line 14: FWIW, this is `build_sdks` in most other Makefiles. This does not especially matter to me, but is worth noting.

jkodroff (approved):

PR #808, opened 2022-04-29
Make all VPC outputs required

danielrbradley:
Makes it much easier to use the outputs to inputs of other things


stack72 (approved):
Do you think we should cut a -beta.2? Or wait and get another few fixes in place?

danielrbradley:
Wait til monday for the beta-2 probably. Just fixing what I'm seeing from writing the guide.

PR #821, opened 2022-05-03
Support chaining and shallow plain types in Go SDK

AaronFriel:
Updates the Go SDK codegen and dependencies to support:
- https://github.com/pulumi/pulumi/issues/9506
- https://github.com/pulumi/pulumi/issues/9507

This improves the Go SDK so that the following example can run without ApplyT:

```go
package main

import (
	aws_ecs "github.com/pulumi/pulumi-aws/sdk/v5/go/aws/ecs"
	"github.com/pulumi/pulumi-awsx/sdk/go/awsx/ecs"
	"github.com/pulumi/pulumi-awsx/sdk/go/awsx/lb"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		cluster, err := aws_ecs.NewCluster(ctx, "cluster", nil)
		if err != nil {
			return err
		}
		lb, err := lb.NewApplicationLoadBalancer(ctx, "lb", nil)
		if err != nil {
			return err
		}
		_, err = ecs.NewFargateService(ctx, "nginx", &ecs.FargateServiceArgs{
			Cluster: cluster.Arn,
			TaskDefinitionArgs: &ecs.FargateServiceTaskDefinitionArgs{
				Container: &ecs.TaskDefinitionContainerDefinitionArgs{
					Image:  pulumi.StringPtr("nginx:latest"),
					Cpu:    pulumi.IntPtr(512),
					Memory: pulumi.IntPtr(128),
					PortMappings: ecs.TaskDefinitionPortMappingArray{
						ecs.TaskDefinitionPortMappingArgs{
							ContainerPort: pulumi.IntPtr(80),
							TargetGroup:   lb.DefaultTargetGroup,
						},
					},
				},
			},
		})
		if err != nil {
			return err
		}
		ctx.Export("dnsName", lb.LoadBalancer.DnsName())
		return nil
	})
}
```

Without these changes, the `ecs.NewFargateService` would have to be run in an ApplyT (9507), and the export of `dnsName` would require an ApplyT as well, with a very unintuitive signature.

### DIFF
--- a/schemagen/go.mod
+++ b/schemagen/go.mod
@@ -4,8 +4,8 @@ go 1.17
 
 require (
 	github.com/pkg/errors v0.9.1
-	github.com/pulumi/pulumi/pkg/v3 v3.30.1-0.20220429024829-810756e7b968
-	github.com/pulumi/pulumi/sdk/v3 v3.30.1-0.20220429024829-810756e7b968
+	github.com/pulumi/pulumi/pkg/v3 v3.31.1-0.20220503073946-62a228d9b3f0
+	github.com/pulumi/pulumi/sdk/v3 v3.31.1-0.20220503073946-62a228d9b3f0
 )
 
 require (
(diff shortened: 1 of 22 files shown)


stack72 (approved):
This is fantastic! thank you so much @AaronFriel and @iwahbe

stack72:
@AaronFriel one thing this does not solve is that with a cloudTrail in Go, it returns a Bucket, LogGroup and a Trail and these are still marked as outputs:

```
type Trail struct {
	pulumi.ResourceState

	// The managed S3 Bucket where the Trail will place its logs.
	Bucket s3.BucketOutput `pulumi:"bucket"`
	// The managed Cloudwatch Log Group.
	LogGroup cloudwatch.LogGroupOutput `pulumi:"logGroup"`
	// The CloudTrail Trail.
	Trail cloudtrail.TrailOutput `pulumi:"trail"`
}
```

So the Go SDK is still not hugely usable as that Output would need to be cast right?

Issue #788, opened 2022-04-27
Fix getDefaultVpc


danielrbradley:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Can't invoke downstream providers within an invoke due to https://github.com/pulumi/pulumi/issues/7583. Once that's resolved, then we should remove the deprecation warnings in the schema and remove the DefaultVpc resource hack.

Issue #784, opened 2022-04-27
CI fails with repositoryUrl undefined


danielrbradley:
ECR test consistently failing in CI and consistently passing locally. Issue seems to be related to the repositoryUrl being undefined after the repository is created. Need to identify why CI resolves the repositoryUrl to undefined. 

It looks like this has been experienced by other users too: https://github.com/pulumi/pulumi/issues/6665#issuecomment-811173269

Given this appears functional most of the time, we could just skip the test for now, add an issue and revisit later?

_Originally posted by @danielrbradley in https://github.com/pulumi/pulumi-awsx/pull/780#issuecomment-1110698795_

PR #825, opened 2022-05-08
Change reason for skipping TestDefaultVpc to point to relevant issue

pgermishuys:
Change reason for skipping TestDefaultVpc to point to relevant issue.
The relevant issue is https://github.com/pulumi/pulumi-awsx/issues/788 and not https://github.com/pulumi/pulumi-awsx/issues/784

Note: I didn't add a changelog entry as I didn't think it was necessary as this change doesn't impact crosswalk.


stack72 (approved):
Thank you!!

PR #829, opened 2022-05-10
Add support for docker image tag (#829)

niravbhimani53:
I have added the support of the docker tag Image class for ECR in `awsx` package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 * Add multi-lang component support scaffolding.
+* Add support for docker tag for ECR in `awsx` package
 
 ## 0.40.0 (2022-03-24)
 * Compatibility with pulumi-aws v5.0.0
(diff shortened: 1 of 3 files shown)


niravbhimani53:
Can someone please advise about the command to generate the SDK? I followed the steps on this page https://github.com/pulumi/pulumi-awsx/blob/master/DEVELOPMENT.md#development-workflow. But command updated none of the files for .NET, Golang or Python.

danielrbradley (commented):
I've just been going through the PR backlog and found this outstanding work. Sorry we didn't manage to get this reviewed at the time you wrote it.

I can't quite tell from the description or code changes exactly what a user would have done differently to leverage the tags. I _think_ they would have specified the tag within the name of the resource, which would be extracted and passed into the image name signature. It then separates the image name into a "base" and "hash" version. 

The `baseImageName` is used when pushing the image in place of the `imageName`. The `hashImageName` only appears to be used for logging.

Looking at the latest version of the code it appears we have a new imageTag argument added in https://github.com/pulumi/pulumi-awsx/pull/1230 and https://github.com/pulumi/pulumi-awsx/pull/1229 rather than including it as part of the resource name. However, it doesn't appear that we include the tag within the image name. 

Most recently there was also support added for manually specifying the image name: https://github.com/pulumi/pulumi-awsx/pull/1257

@corymhall @t0yv0 as reviewers on that most recent change:
1. Should we be adding the new imageTag to the imageName, if present?
2. I think we can close this PR as all other aspects have been implemented via alternative approaches.

Issue #811, opened 2022-05-02
Loosen validation of VPC SubnetTypes


stack72:
`Private` is allowed yet `private` is not

jkodroff:
@danielrbradley Is there a way to make the input acceptance less strict while not overly complicating the code in TS?

danielrbradley:
```
spec.type.localeCompare("Private", undefined, { sensitivity: "accent" })
```

PR #831, opened 2022-05-10
Fix up validation of vpc configuration types

stack72:
Fixes: #811


Makefile
```diff
@@ -51,6 +51,7 @@ build_nodejs::
 		yarn install && \
 		yarn run tsc --version && \
 		yarn run tsc && \
+		yarn test && \
```
danielrbradley, line 54: This should be in the provider build not the sdk build

stack72: ah you are correct! fixing now

PR #836, opened 2022-05-10
Build native binaries

### LINKED ISSUE #834: Produce single binary executable
## Hello!

## Issue details

Use [Vercel PKG](https://github.com/vercel/pkg) to produce native binaries.

danielrbradley:
Fixes #834

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ pre_dist:: provider
 dist:: pre_dist
 	mkdir -p dist
 	cd bin && \
-		npx --yes -- pkg index.js --compress GZip --target node17-macos-x64,node17-macos-arm64,node17-linux-x64,node17-win-x64 --output ../dist/out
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -9,6 +9,10 @@ CODEGEN         := pulumi-gen-${PACK}
 
 WORKING_DIR     := $(shell pwd)
 
+GOPATH 		 ?= "$(HOME)/go"
```
stack72, line 12: Any reason why we are adding GOPATH / GOBIN here?

danielrbradley: Making the install work the same as our other go-based providers - rather than having to add ~/awsx/dist to the user's path. This logic is taken from the go docs on how to establish the GOBIN
> Executables (main packages) are installed to the directory named by the GOBIN environment variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH environment variable is not set.

stack72: 🙌🙌🙌🙌

Issue #285, opened 2019-05-27
Add VPC Endpoints options


jen20:
It's quite common to want to add [VPC Endpoints][vpce] to an AWS VPC in order to provide access to AWS services which does not transit the public internet or (from private subnets) the NAT gateway.

Adding an endpoint requires a list of routing tables to modify - generally (but not necessarily) all of the subnets in the VPC - to be passed into a `aws.ec2.VpcEndpoint` resource. Right now adding gateway endpoints for S3 and DynamoDB can be achieved with the following program:

```typescript
function addAmazonServiceGateways(namePrefix: string, vpc: awsx.ec2.Vpc) {
    const publicRouteTables = vpc.getSubnets("public").map(x => x.routeTable!.id);
    const privateRouteTables = vpc.getSubnets("private").map(x => x.routeTable!.id);
    const allRouteTables = publicRouteTables.concat(privateRouteTables);

    new aws.ec2.VpcEndpoint(`${namePrefix}-s3-endpoint`, {
        vpcId: vpc.id,
        serviceName: `com.amazonaws.${aws.config.region}.s3`,
        routeTableIds: allRouteTables,
    });

    new aws.ec2.VpcEndpoint(`${namePrefix}-dynamodb-endpoint`, {
        vpcId: vpc.id,
        serviceName: `com.amazonaws.${aws.config.region}.dynamodb`,
        routeTableIds: allRouteTables,
    });
}
```

It would be nice to add support for this directly to `awsx.ec2.Vpc`, such that a user can either opt into endpoint creation as part of `VpcArgs`, or provide method calls to add support for both gateway and interface endpoints and a set of helper functions for constructing the endpoint service names.

[vpce]: https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints.html

stack72:
I vote that we add this directly as a Boolean parameter than we pass to VPC that will not require a separate function call

It would nice nice sugar to add to the process

lukehoban:
What are the full list of such endpoints? (And where is that list documented?).

I believe there are now two kinds of endpoints - gateway and interface. How should these be handled?

levand:
A big +1 to this, in fact, I would go so far as to say that for private and isolated subnets, gateways for common data-intensive services should be created by default.

Otherwise you end up routing all S3 traffic (for example) through the NAT, which is probably not what was intended.

PR #837, opened 2022-05-11
Add support for VPC Endpoints

stack72:
Fixes: #285

```
const myVpc = new awsx.ec2.Vpc("awsx-with-service-endpoint", {
  vpcEndpointSpecs: [
    {
      serviceName: "com.amazonaws.us-west-2.s3",
    },
  ],
});
```


Makefile
```diff
@@ -90,7 +90,7 @@ istanbul_tests::
 	cd awsx-classic/tests && \
 		yarn && yarn run build && yarn run mocha $$(find bin -name '*.spec.js')
 
-install_nodejs_sdk:: build_nodejs
+install_nodejs_sdk:: #build_nodejs
```
danielrbradley, line 93: Think this should be reverted

```suggestion
install_nodejs_sdk:: build_nodejs
```

danielrbradley (approved):
Approved, just need to revert the makefile comment

PR #841, opened 2022-05-16
Add SDK dependency on Pulumi-docker

### LINKED ISSUE #742: Specify dependency on pulumi-docker in schema for Multi Language
## Hello!

## Issue details

For MLC support, AFAICT, the internal implementation of the provider depends on `pulumi-docker` (e.g. for [ECS](https://github.com/pulumi/pulumi-awsx/pull/734)), but this dependency is not specified in the schema for any of the languages, which will prevent plugin discovery from working correctly.

Note: When the AWSX plugin is initially installed, I would expect the Pulumi `docker` plugin to be installed as part of installing the AWSX plugin's NPM dependencies (i.e. installing `@pulumi/docker` will end up running `pulumi plugin install resource docker ...`). However, if someone deletes the `docker` plugin from their machine, and is using AWSX from Python, .NET, or Go, the next time they run `pulumi up`, it won't statically be able to determine that `docker` is a required plugin, so it won't be installed automatically.

For Python and .NET, the fix is to specify a dependency on `pulumi-docker` and `Pulumi.Docker` in the schema:

https://github.com/pulumi/pulumi-awsx/blob/69254186707fd9c1108c4c758a8db136be35d3c9/provider/cmd/pulumi-resource-awsx/schema.json#L164-L167

https://github.com/pulumi/pulumi-awsx/blob/69254186707fd9c1108c4c758a8db136be35d3c9/provider/cmd/pulumi-resource-awsx/schema.json#L151-L154

For Go, dependencies are specified manually in the Go SDK's `go.mod` (the `go.mod` file is not generated automatically). However, when manually specifying a dependency on `github.com/pulumi/pulumi-docker/sdk/v3` in the `go.mod`, if it's not actually being used in the Go SDK, it'll be removed by `go mod tidy`. https://github.com/pulumi/pulumi/issues/9180 tracks adding a way to specify Go dependencies in the schema, which will generate some dummy imports inside the generated Go SDK, so such dependencies won't be stripped from `go.mod` by `go mod tidy`.

## Tasks

- [ ] Python: Specify dependency on `pulumi-docker` in schema
- [ ] .NET: Specify dependency on `Pulumi.Docker` in schema
- [ ] Go: Specify dependency on docker (blocked on https://github.com/pulumi/pulumi/issues/9180)

danielrbradley:
Fixes #742


schemagen/pkg/gen/schema.go
```diff
@@ -52,15 +52,17 @@ func GenerateSchema(packageDir string) schema.PackageSpec {
...
+					"Pulumi.Aws":    "5.*",
+					"Pulumi.Docker": "3.*",
 				},
 				"liftSingleValueMethodReturns": true,
 			}),
 			"go": rawMessage(map[string]interface{}{
 				"generateResourceContainerTypes": true,
 				"importBasePath":                 "github.com/pulumi/pulumi-awsx/sdk/go/awsx",
 				"liftSingleValueMethodReturns":   true,
+				"internalDependencies":           []string{"github.com/pulumi/pulumi-docker/sdk/v3/go/docker"},
```
stack72, line 65: Is there any reason why we don't need to specify aws as well here?

stack72 (approved):
Left a small question about aws as a go dependency

awsx/schema.json
```diff
@@ -2842,12 +2842,16 @@
...
-                "Pulumi.Aws": "5.*"
+                "Pulumi.Aws": "5.*",
+                "Pulumi.Docker": "3.*"
             }
         },
         "go": {
             "generateResourceContainerTypes": true,
             "importBasePath": "github.com/pulumi/pulumi-awsx/sdk/go/awsx",
+            "internalDependencies": [
+                "github.com/pulumi/pulumi-docker/sdk/v3/go/docker"
```
stack72, line 2853: I think it needs to just point to GitHub.com/Pulumi/Pulumi-docker/sdk as that's where the go.mod sits

schemagen/pkg/gen/schema.go
danielrbradley: Because the SDK directly references types from the AWS provider - so it's got a natural dependency already: https://github.com/pulumi/pulumi-awsx/pull/841/files#diff-7b5028b779639efd98a7f47865b168ca39285e1c46d94140c5d38c516d00b127L8

awsx/schema.json
danielrbradley: Ah yup - simplifying now 👍

stack72: Interesting - let's merge this and then look at fixing upstream. That path that upstream suggests feels very incorrect but that's a job for a different day

PR #852, opened 2022-05-20
Improve local and CI build flow

danielrbradley:
- Use file targets to minimise rebuild work locally
- Avoid rebuilding on every CI step
- Combine SDK build & acceptance test jobs as they're 1:1
- Install nuget package as source via make (not just in CI)
- Capture and restore bins from provider build step


stack72 (approved):
The makefile changes are making me cringe... but that's makes fault not yours :)

jkodroff:
@stack72 @danielrbradley I would gladly support moving this repo to Task: clear upside and no downside I can think of since Task is not difficult to install and is supported on the platforms we need.

danielrbradley:
@jkodroff this would be interesting here so we've got a side-by-side of make done properly vs task done property.

The part I'm most happy with is that both local and CI build processes are much more snappy.

PR #861, opened 2022-06-01
use bash in PATH

afreakk:
Systems like nixos, does not have bash in `/bin/bash`.  
So previous script did not execute.
With `/usr/bin/env bash` it will use the systems default bash interpreter.
So this is a more flexible approach.

Weirdly, on `v1.0.0-beta. 5, 6 and 7` `~/.pulumi/plugins/resource-awsx-v1.0.0-beta.*/pulumi-resource-awsx` is no longer a bash script, but a dynamic executable.  
Is that intended?  
That does not work very well on nixos.

### DIFF
--- a/awsx/pulumi-resource-awsx
+++ b/awsx/pulumi-resource-awsx
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-node $SCRIPT_DIR $@
+node $SCRIPT_DIR/bin $@


danielrbradley (changes requested):
Thanks for the contribution!

Happy to make the switch to use `/usr/bin/env`, though to make this script work locally you'll also need to add `/bin` to the path as that's where the compilled js resides and this binary is not copied into the bin directory.

This entry point is now only used for the purpose of local testing - the actual distribution has moved to using binaries packaged through Vercel/pkg to avoid issues for users with missing or different versions of node locally. This is closer to the experience of all other providers which are built with go. 

Could you elaborate on why native binaries don't work well on nixos? It's a system I'm unfamiliar with.

danielrbradley (approved):

PR #862, opened 2022-06-02
Make binaries static for linux

afreakk:
@danielrbradley 
Not sure if this is the only place to change, but I  tested some locally, and the binaries created seem to work fine.


danielrbradley:
Thanks! Currently the binary being built won't be tested because the local binary is built using the default "node16" target which will then infers the platform and arch from the local system - and will use linux rather than linux static.

I'll make a quick change to the build to make sure we can test this before it being released.

danielrbradley:
Merged as a part of #865

PR #868, opened 2022-06-06
Pass opts to get the subnets through `getDefaultVpc` function

JoeStead:


stack72:
Thanks so much for this @JoeStead :)

PR #885, opened 2022-07-23
expose route table ids

jaxxstorm:
Route table ids can be important for VPCs after they've been created.

Exposing the route table ids as outputs allows us to easily support resources like NAT instances (cheaper than NAT Gateways), AWS Transit Gateways and AWS Client VPNs

### DIFF
--- a/awsx/schema.json
+++ b/awsx/schema.json
@@ -1689,6 +1689,24 @@
                     },
                     "description": "The Route Table Associations for the VPC."
                 },
+                "publicRouteTableIds": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "privateRouteTableIds": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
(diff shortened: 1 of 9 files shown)


jaxxstorm:
@danielrbradley am I missing something here? I generated the types locally so trying to understand the failures here

awsx/schema.json
```diff
@@ -1689,6 +1689,24 @@
...
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "isolatedRouteTableIds": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
```
danielrbradley, line 1709: The schema.json is generated by the `schemagen` program. The VPC types are defined using this code here: https://github.com/pulumi/pulumi-awsx/blob/master/schemagen/pkg/gen/ec2.go

blampe:
https://www.pulumi.com/registry/packages/awsx/api-docs/ec2/vpc/#routetables_nodejs

PR #901, opened 2022-08-23
[internal] Add a release script that will work for the v0.40.x series

stack72:
We need to push a patch of a regression of the v0.40.x series but
our currently release machinery is specifically set up for the
new way of working

### DIFF
--- a/.github/workflows/old-release.yml
+++ b/.github/workflows/old-release.yml
@@ -1,0 +1,206 @@
+env:
+  AWS_REGION: us-west-2
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GO111MODULE: "on"
+  NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/old-release.yml
```diff
@@ -0,0 +1,206 @@
+env:
```
viveklak, line 1: I assume this is mostly from the state of the repo as of 0.40 release and the makefile.old is verbatim are mostly copied from the 0.40.x release/branch?

.github/workflows/old-release.yml
```diff
@@ -0,0 +1,206 @@
...
+          - 1.16.x
+        language:
+          - nodejs
+        nodeversion:
+          - 13.x
+name: release
+"on":
+  push:
+    tags:
+      - v0.*.*
```
viveklak, line 206: We won't really know if things work until the tag is released. Is that what we want? Should we add some build/validation workflow for 0.x.x branch to catch any issues?

viveklak (approved):

PR #907, opened 2022-09-14
Do not throw an exception when NAT gateway strategy is None.

### LINKED ISSUE #884: Allow setting "None" in NAT Gateway Strategy for private subnets
There are valid situations whereby you'd want to create valid private subnets but not have NAT Gateways:

```
const vpc = new awsx.ec2.Vpc("natInstance", {
    numberOfAvailabilityZones: 2,
    cidrBlock: "172.20.0.0/22",
    natGateways: {
        strategy: "None"
    }
})
```

One such example is if you're deciding to use NAT instances for cost-effective egresss from a VPC

jkodroff:
Fixes #884

### DIFF
--- a/awsx/ec2/vpc.test.ts
+++ b/awsx/ec2/vpc.test.ts
@@ -91,8 +91,12 @@ describe("validateNatGatewayStrategy", () => {
   );
 
   describe("strategy is None", () => {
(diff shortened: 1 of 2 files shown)


stack72:
Thanks for this @jkodroff :)

PR #917, opened 2022-09-26
Name subnets after the AZ they are deployed in.

### LINKED ISSUE #913: VPC 1.0 Usability: Change naming of subnets from 1, 2, 3 to match AZ?
If we generate subnets with the default args, we name them, e.g. `vpc-public-1`, `vpc-public-2`, `vpc-public-3`. The names are more useful if we name them after the AZ they're in instead, e.g. `vpc-public-1a`, `vpc-public-1b`, `vpc-public-1c`.

jkodroff:
Fixes #913

Note that this change, as implemented, will cause a replace for existing resources as the name of the resource is currently tied to the `Name` tag. Happy to separate the two based on PR feedback, or we can include a release note indicating how to alias to prevent the replace.


jkodroff:
@lukehoban Do you have any thoughts on how we should communicate this change to users: Are migration (alias) instructions sufficient?

awsx/ec2/subnetDistributor.ts
```diff
@@ -98,7 +100,7 @@ export function getSubnetSpecs(
         azName: azNames[i],
         cidrBlock: publicSubnetCidrBlock,
         type: "Public",
-        subnetName: `${vpcName}-${publicSubnetsIn[j].name ?? "public"}-${i + 1}`,
+        subnetName: `${vpcName}-${publicSubnetsIn[j].name ?? "public"}-${azSuffix}`,
         tags: publicSubnetsIn[j].tags,
       });
```
danielrbradley, line 105: Let's add aliases here as you suggested to maintain backward compatibility with the old names.

danielrbradley (changes requested):

jkodroff:
@danielrbradley It's been dormant for a while, but I picked this back up. Unit tests look like they pass and I just need to run it against a test Pulumi program made before this change to ensure no resources get re-created.

danielrbradley (commented):
This looks great with the aliases added in! Let me know once you've confirmed the upgrade path works without recreation and I'll approve.

PR #921, opened 2022-09-28
feat: allow tagging public subnets

rawkode:
Security group rules are based on tagging and it's a common use-case to allow ingress and egress based on these tags, as such we should support tagging the public subnets which are most likely to require ingress.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 * Add multi-lang component support scaffolding.
 * Fix type errors in TypeScript checking awsx-classic properties.
 * Ensure that FargateService and EC2Service default `continueBeforeSteadyState` to false.
+* Allow tagging public subnets on VPC for security group matching
 
 ## 0.40.0 (2022-03-24)
 * Compatibility with pulumi-aws v5.0.0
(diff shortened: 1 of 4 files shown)


awsx/schema.json
```diff
@@ -1848,6 +1848,12 @@
                     },
                     "plain": true,
                     "description": "A list of VPC Endpoints specs to be deployed as part of the VPC"
+                },
+                "publicSubnetTags": {
+                    "type": "object",
+                    "additionalProperties": {
+                        "type": "string"
+                    }
```
lukehoban, line 1856: This will need a description to populate documentation.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 * Add multi-lang component support scaffolding.
 * Fix type errors in TypeScript checking awsx-classic properties.
 * Ensure that FargateService and EC2Service default `continueBeforeSteadyState` to false.
+* Allow tagging public subnets on VPC for security group matching
```
lukehoban, line 8: Should we add equivalents to this for all the other subsets of resources we're creating as well?  Or at least also `privateSubnetTags`?

jkodroff:
I believe that we have a bug in the existing implementation. `subnets` already has a `tags` field, but it's not actually applied to the subnets: https://github.com/pulumi/pulumi-awsx/issues/922

danielrbradley (changes requested):
I don't think this new property should be necissary - I think we're just missing applying the tags which we've already got to the subnets.

If we need more options around specific tags on specific subnets then we'll need a more complete design which also covers all other types of subnet.

PR #936, opened 2022-11-09
Default FargateService to non-public

### LINKED ISSUE #801: Consider defaulting FargateService to non-public
## Hello!

## Issue details

It might be a better default from a security perspective to set `assignPublicIp: false` on `awsx.ecs.FargateService`. We currently override this option almost everywhere in our code using awsx. Breaking change for next major if it makes sense.

### Affected area/feature

`awsx.ecs.FargateService`

justinvp:
Fixes #801


justinvp:
Hmm, I'll look into the test failures.

justinvp:
@danielrbradley, this is ready for another look

danielrbradley (approved):
This looks great! Let's merge :shipit:

PR #937, opened 2022-11-09
don't create routes to nat gateway if strategy=None

jaxxstorm:
If we're specifying no NAT gateway, we don't want to go ahead and create
routes to it because the NAT gateway doesn't exist


danielrbradley (approved):
Great fix thank you!

lukehoban:
Not to put the burden on @jaxxstorm for this. But as a general note - would be great to include test cases with changes like this to ensure we're locking in the correctness improvements!

PR #939, opened 2022-11-09
propagate tags to all resources

### LINKED ISSUE #922: VPC: Tags are not applied
### What happened?

When creating an AWSX VPC, tags are not applied to dependent resources.

### Steps to reproduce

```python
"""A Python Pulumi program"""

import pulumi_awsx as awsx

vpc = awsx.ec2.Vpc(
    "david-awsx-question",
    awsx.ec2.VpcArgs(
        subnet_specs=[
            awsx.ec2.SubnetSpecArgs(
                type=awsx.ec2.SubnetType.PUBLIC,
                tags={
                    "ingress": "HECK YEAH, BUDDY!",
                },
            ),
            awsx.ec2.SubnetSpecArgs(
                type=awsx.ec2.SubnetType.PRIVATE,
            ),
        ],
        tags={
            "createdby": "jkodroff",
        },
    )
)
```

### Expected Behavior

* `createdby` should (probably) be applied to all resources. It only seems to appear on the VPC itself.
* `ingress` should (definitely) be applied to the public subnets, and it is not.

### Actual Behavior

See above

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

jaxxstorm:
All resources should get the tags assigned to them.

In addition, I've added the subnet type to the route tables so we can differentiate them

Fixes #922

### DIFF
--- a/awsx/ec2/vpc.ts
+++ b/awsx/ec2/vpc.ts
@@ -222,8 +222,8 @@ export class Vpc extends schema.Vpc<VpcData> {
                 subnetId: subnet.id,
                 allocationId: createEip ? eips[i].allocationId : allocationIds[i],
                 tags: {
-                  Name: `${name}-${i + 1}`,
                   ...args.tags,
+                  Name: `${name}-${i + 1}`,
                 },
               },
               { parent: subnet, dependsOn: [subnet] },
--- a/awsx/ec2/vpc.ts
+++ b/awsx/ec2/vpc.ts
@@ -108,6 +108,7 @@ export class Vpc extends schema.Vpc<VpcData> {
       {
         vpcId: vpc.id,
         tags: {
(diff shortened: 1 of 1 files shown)


awsx/ec2/vpc.ts
```diff
@@ -218,6 +223,7 @@ export class Vpc extends schema.Vpc<VpcData> {
                 allocationId: createEip ? eips[i].allocationId : allocationIds[i],
                 tags: {
                   Name: `${name}-${i + 1}`,
+                  ...args.tags,
```
danielrbradley, line 226: This should be before the name to ensure Name doesn't get overwritten.

jaxxstorm: Fixed!

PR #946, opened 2022-11-21
1.0 docs

danielrbradley:


README.md
```diff
@@ -26,15 +25,11 @@ The AWS Infrastructure package exposes many high level abstractions.  Including:
...
+### Migration from 0.x to 1.0
 
-* `apigateway`
-* `autoscaling`
-* `ecs`
-* `ec2`
-* `cloudwatch`
-* `lb`
-* `cloudtrail`
+Before version 1, this package only supported components in Typescript. All the existing components from the 0.x releases are now available in under the `classic` namespace.
```
justinvp, line 30: Nit:

```suggestion
Before version 1, this package only supported components in TypeScript. All the existing components from the 0.x releases are now available in under the `classic` namespace.
```

README.md
```diff
@@ -26,15 +25,11 @@ The AWS Infrastructure package exposes many high level abstractions.  Including:
...
+
+When upgrading change your references from `@pulumi/awsx` to `@pulumi/awsx/classic`.
```
justinvp, line 32: Do we want to say anything along the lines of "or, move to the new components in `@pulumi/awsx`? To offer that as an option as well? Otherwise, folks may get the impression that they should only move to classic.

justinvp (approved):

README.md
thomas11, line 30: > available in under

"in" or "under"?

thomas11 (approved):

PR #956, opened 2022-12-02
add display name for registry

susanev:
@danielrbradley pls let me know if this is not the right change to make, thank you!

### DIFF
--- a/awsx/schema.json
+++ b/awsx/schema.json
@@ -1,5 +1,6 @@
 {
     "name": "awsx",
+    "displayName": "AWSx (Pulumi Crosswalk for AWS)",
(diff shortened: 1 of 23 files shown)


danielrbradley (changes requested):
Name looks fine, but this schema is generated, so we'll need to implemented the change elsewhere and regenerate.

I think this is the source file: https://github.com/pulumi/pulumi-awsx/blob/master/schemagen/pkg/gen/schema.go#L42

susanev:
@danielrbradley i ran `make`, and there are a lot of whitespace changes in there, let me know if i did something wrong.

PR #957, opened 2022-12-05
Remove explicit dotnet install

danielrbradley:
We're installing 3.1 on an image which already has 6.0. This has no effect as dotnet will automatically choose the newest version.

This was found because were getting an error on the acceptance tests where we used the install action but failed to specify the matrix.dotnetversion which meant the action automatically installed the latest version (7.0) which caused build errors.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -149,10 +149,6 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{matrix.pythonversion}}
-      - name: Setup DotNet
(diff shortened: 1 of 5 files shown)


danielrbradley:
Apparently .NET 7 has only recently been released - or at least been made avaialble in the GH action.

roothorp:
LGTM - can we run the tests on this PR? Don't seem to have run automagically...
Also looks like actions are failing anyway because of https://github.com/pulumi/pulumi-aws/issues/2243, so maybe we can't anyway.

roothorp (approved):

Issue #985, opened 2023-02-01
Name tag should not be applied to sub-resources of Vpc


peak-stephen:
### What happened?

I created a Vpc resource. Later I tried to set the `Name` tag of the VPC to a different value to change the display name in the AWS console, Pulumi tried to apply this to route tables and got stuck.

### Expected Behavior

The name of the VPC is updated.

### Steps to reproduce

1. Create a stack with a VPC. (I'm not sure if it needs to exist first or if this would happen on create as well).
2. Add `tags: { Name: "something-else" }` to the VPC. Run pulumi up
3. See that Pulumi wants to update every route table in the VPC to have the same name.
4. (optional) choose "yes" and observe that Pulumi will hang because you can't do that :)

### Output of `pulumi about`

CLI
Version      3.53.1
Go Version   go1.19.5
Go Compiler  gc

Plugins
NAME             VERSION
aws              5.16.2
awsx             1.0.1
command          0.7.0
docker           3.6.1
docker-buildkit  0.1.26
eks              1.0.1
kubernetes       3.23.1
python           unknown

Host
OS       ubuntu
Version  22.04
Arch     x86_64

8<--------------- snip ------------------

Dependencies:
NAME                    VERSION
pip                     22.3.1
pulumi-awsx             1.0.1
pulumi-command          0.7.0
pulumi-docker-buildkit  0.1.26
pulumi-eks              1.0.1
setuptools              66.1.1
typeguard               2.13.3
wheel                   0.38.4

### Additional context

I would guess that #939 introduced this behaviour.
#967 is about the lack of update to the VPC itself, and should probably be fixed at the same time.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Thanks for submitting this issue! It appears that you've identified the root cause and am attempting to submit a fix for it. We look forward to reviewing and submitting your PR!

t0yv0:
Fixed in https://github.com/pulumi/pulumi-awsx/pull/986 I believe.

PR #986, opened 2023-02-03
Re-order tag priority in ec2.Vpc

### LINKED ISSUE #967: If I set a `name` tag in the VPC component, use that instead of the resource name
### What happened?

```typescript
    cidrBlock: "10.0.0.0/16",
    numberOfAvailabilityZones: 3,
    subnetSpecs: [{
        type: awsx.ec2.SubnetType.Public,
        name: "public",
    }],
    natGateways: {
        strategy: "None"
    },
    tags: {
        name: "my-vpc"
    }
});
```

If I have the above code for a new VPC, the component is setting the "name" tag to be the resource name (in the above case `vpc` rather than the name tag, `my-vpc`).

It happens here: https://github.com/pulumi/pulumi-awsx/blob/35614a308aac3dfa7e0c7f4cbc5d685c2b696580/awsx/ec2/vpc.ts#L92-L95

### Steps to reproduce

Use above code

### Expected Behavior

VPC should use the name tag to set the vpc name

### Actual Behavior

Component is overwriting the name tag with the resource name instead of what I'm providing

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

peak-stephen:
Fixes #967, #985

As described in the above issues it's currently impossible to actually set the name of a VPC, for two different reasons.

This PR fixes both issues. I don't _think_ it will cause a regression that anybody notices, because right now setting a name causes creation of the VPC to fail.

---

A "nice to have" would be constructing the name tag of each subnet based on the name tag of the over all VPC, but that's a bigger change/feature so I'd rather treat this PR as simply a bug fix.

### DIFF
--- a/awsx/ec2/vpc.ts
+++ b/awsx/ec2/vpc.ts
@@ -84,15 +84,14 @@ export class Vpc extends schema.Vpc<VpcData> {
 
     validateNatGatewayStrategy(natGatewayStrategy, subnetSpecs);
 
+    const sharedTags = { Name: name, ...args.tags }
+    
     const vpc = new aws.ec2.Vpc(
       name,
       {
         ...args,
         cidrBlock,
-        tags: {
-          ...args.tags,
-          Name: name,
-        },
+        tags: sharedTags,
       },
       { parent: this },
     );
@@ -107,10 +106,7 @@ export class Vpc extends schema.Vpc<VpcData> {
       `${name}`,
       {
         vpcId: vpc.id,
-        tags: {
-          ...args.tags,
-          Name: name,
-        },
+        tags: sharedTags,
       },
       { parent: vpc, dependsOn: [vpc] },
     );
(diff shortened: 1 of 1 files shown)


danielrbradley:
Thanks for the contribution! I think this is all good but want to work through proving there's no risk of breaking changes here. Will follow up next week

awsx/ec2/vpc.ts
```diff
@@ -161,7 +157,7 @@ export class Vpc extends schema.Vpc<VpcData> {
               mapPublicIpOnLaunch: spec.type.toLowerCase() === "public",
               cidrBlock: spec.cidrBlock,
               tags: {
-                ...args.tags,
```
danielrbradley, line 164: This change has no impact on behaviour. The only difference between `args.tags` and `sharedTags` is adding a Name, but the name will always be overwritten below.

danielrbradley (approved):
Ok, worked through this in detail. I think the only way this could cause a breaking change is if anyone was accidentally passing a Name in their `args.tags` which would have previously been a no-op.

Therefore I'm happy for this to merge 👍

awsx/ec2/vpc.ts
peak-stephen: that's true, it was more for consistency/readability. No need to think about "why use `arg.tags` here instead of `sharedTags`

PR #1012, opened 2023-03-16
feat(classic): add GaugeWidget for cloudwatch dashboard

maniolias:
Gauge Widget is missing on cloudwatch types. Here is a proposition of implementation


maniolias:
up 👍

mjeffryes (commented):
Looks reasonable to me. If you add a test to https://github.com/pulumi/pulumi-awsx/blob/master/awsx-classic/tests/cloudwatch/cloudwatch.spec.ts I think we can ship this.

mjeffryes:
Closing this for inactivity.

Issue #877, opened 2022-07-07
CloudTrail trail creates unencrypted bucket by default


pierskarsenbarg:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

By default, creating a CloudTrail trail creates an S3 bucket to store the logs in, but that bucket is unencrypted. 

Basic program in python, but it's the same for all languages:

```python
import pulumi_awsx as awsx

trail = awsx.cloudtrail.Trail("trail")
```

requirements.txt:

```
pulumi>=3.0.0,<4.0.0
pulumi-aws>=5.0.0,<6.0.0
pulumi-awsx==1.0.0-beta9
```

Should it be encrypted (at least with an S3 managed key) by default?

dswartz-altana:
In my opinion, AES256 would suffice as a default, but an option to use a custom/generated/best-practice KMS key would be clutch. Additional considerations for bucket policy (enforce encryption in-transit) and public access block settings would also be much appreciated!

jkodroff:
@pierskarsenbarg Is this in 1.0 beta or the original version?

pierskarsenbarg:
This is for the v1 beta

dswartz-altana:
Any update on this, by chance?

dswartz-altana:
Additionally, ensuring unencrypted traffic is denied would be a mega-plus for us!

```
                {
                    "Sid": "AllowSSLRequestsOnly",
                    "Effect": "Deny",
                    "Principal": "*",
                    "Action": "s3:*",
                    "Resource": [
                        "BUCKET_ARN",
                        "BUCKET_ARN/*"
                    ],
                    "Condition": {
                        "Bool": {"aws:SecureTransport": "false"},
                    }
                }
```

thomas11:
Hi @dswartz-altana, meanwhile S3 buckets are [AES-encrypted by default](https://docs.aws.amazon.com/AmazonS3/latest/userguide/default-encryption-faq.html).

For other readers I just want to point out that the bucket configuration can be easily overridden when creating a new `awsx.cloudtrail.Trail`. I added some examples in #1031.

PR #1031, opened 2023-06-27
Expand CloudTrail example with bucket encryption variants

thomas11:
Prompted by #877


mikhailshilkov (approved):
LGTM. Is there an open issue that is being fixed here?

thomas11:
Resolves #877 (documentation and testing)

Issue #1010, opened 2023-03-08
Use new version of docker provider in ECR module


pierskarsenbarg:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Since the new docker.Image resource is faster and better we should use this instead

chris-stetter:
Any plans to release this any time soon? This blocks updating to `pulumi-docker` v4.

PR #1035, opened 2023-07-03
Updated docker provider to latest version

pierskarsenbarg:
- Updated docker provider
- Added "classic" docker package (the last one of v3) to classic AWSx, so no real update is required there (unless you want one?)

The one thing I wasn't sure what to do about was [this value](https://github.com/pierskarsenbarg/pulumi-awsx/blob/1750308ef9e8666d299606fa300a1146f0dd7995/awsx/ecr/image.ts#L83) as there wasn't really a comparable option in the old version. Happy to take comments

Fixes: #1010

### DIFF
--- a/awsx/ecs/fargateMemoryAndCpu.test.ts
+++ b/awsx/ecs/fargateMemoryAndCpu.test.ts
@@ -1,0 +1,51 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { calculateFargateMemoryAndCPU, maxMemGB, maxVCPU } from "./fargateMemoryAndCpu";
+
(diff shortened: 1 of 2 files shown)


awsx/ecr/image.ts
```diff
@@ -43,48 +43,49 @@ export function computeImageFromAsset(
...
+      args: dockerInputs.args,
+      cacheFrom: dockerInputs.cacheFrom
+        ? {
+            images: dockerInputs.cacheFrom,
+          }
+        : undefined,
+      platform: "linux/arm64",
+      target: dockerInputs.target,
+    },
+  };
```
danielrbradley, line 86: This appears to be unused.

danielrbradley (commented):
It looks like you're declaring `dockerImageArgs` but not using the value. 

On what to fill in the args with - I'd suggest we should just be transparently passing all arguments down to docker – we should change the definitions in the new awsx to reflect the new docker args exactly. This should be done programatically in the schema gen program.

This would cause the PR to technically be a breaking change, but I think this would be reasonable to ship in a minor version bump with changelog.

awsx/ecr/image.ts
pierskarsenbarg: Yeah that was me not being careful enough

sdk/nodejs/package.json
```diff
@@ -17,9 +17,10 @@
     },
     "dependencies": {
         "@pulumi/aws": "^5.35.0",
-        "@pulumi/docker": "^3.6.1",
+        "@pulumi/docker": "^4.3.0",
         "@pulumi/pulumi": "^3.0.0",
         "@types/aws-lambda": "^8.10.23",
+        "docker-classic": "npm:@pulumi/docker@3.6.1",
```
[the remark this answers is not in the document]
pierskarsenbarg: No nor me, but good to know

Issue #827, opened 2022-05-09
ECR buildAndPushImage does not work correctly for multi-stage Dockerfiles


ztmr:
### What happened?

When using a single `Dockerfile` with multiple named stages, `buildAndPushImage` on `awsx.ecr.Repository` does not work correctly if creating images for more than one of these stages from within the same stack (and parent component?).

Only the first image (=first referenced stage) is created and the others are just pointing to the first one.

This seems to be relevant in other cases where we want to pass custom build arguments too -- haven't tested this, but passing custom `buildOptions` for the same `Dockerfile` is most likely going to be wrong as well.

### Steps to reproduce

- `myRegistry.buildAndPublishImage({ context: '.', dockerfile: 'my-only.dockerfile', target: 'target-1' });` builds image as expected and returns a reference to it
- `myRegistry.buildAndPublishImage({ context: '.', dockerfile: 'my-only.dockerfile', target: 'target-2' });` results in the same image reference as the earlier, so pointing to a `target-1` image, not `target-2`

### Expected Behavior

- `myRegistry.buildAndPublishImage({ context: '.', dockerfile: 'my-only.dockerfile', target: 'target-1' });` builds image with `target-1` stage and returns a correct image reference to it
- `myRegistry.buildAndPublishImage({ context: '.', dockerfile: 'my-only.dockerfile', target: 'target-2' });` builds image with `target-2` stage and returns a correct image reference to it

### Actual Behavior

This is related to the way the internal image references are derived -- using a SHA1 of Docker command arguments where `target` property is not included in the hash input: https://github.com/pulumi/pulumi-awsx/blob/601f13634a0f7bf6472594ebdeea3260b71decc5/awsx/ecr/image.ts#L98

I thought I can perhaps manipulate the `imageName` directly, but `DockerBuild` structures from `pulumi-awsx` and `pulumi-docker` are confusing and incompatible and it turned out that it cannot be customized. So I tried to manipulate `extraOptions` but that does not seem to be included with `args` (internal) that is also used to calculate the hash. The last resort was to simply symlink my original Dockerfile with some other name which worked eventually...

### Versions used

My env details are listed below, but `master` seems to be the same...

```console
CLI          
Version      3.18.1
Go Version   go1.17.3
Go Compiler  gc

Plugins
NAME        VERSION
aws         4.5.1
docker      3.0.0
eks         0.30.0
kubernetes  3.2.0
nodejs      unknown

[...]

NAME            VERSION
[...]

Issue #473, opened 2019-12-15
Feature request: Use docker.Image, not just functions


joeduffy:
It appears we open code Docker build/publish by simply invoking functions in the Docker package, like `buildAndPushImage`, rather than creating an instance of the `docker.Image` component resource.

This means that a lot of the interesting metadata associated with a Docker build -- like the image name, digest, authentication info, etc -- is not captured by Pulumi, and the image itself isn't a node in the resource graph. It's kind of just a silent/hidden side effect of running your Pulumi update.

This might seem like a theoretical issue, but it meaningfully impacts what you can do with the state: for example, I would love to be able to write PaC rules that scan all `docker:image:Image` resources, no matter where they come from (the Docker package, AWS library, Azure or GCP, etc), but right now as it stands, I cannot.

PR #1044, opened 2023-07-17
Updated docker provider to latest version

pierskarsenbarg:
- Updated docker provider
- Added "classic" docker package (the last one of v3) to classic AWSx, so no real update is required there (unless you want one?)

The one thing I wasn't sure what to do about was [this value](https://github.com/pierskarsenbarg/pulumi-awsx/blob/1750308ef9e8666d299606fa300a1146f0dd7995/awsx/ecr/image.ts#L83) as there wasn't really a comparable option in the old version. Happy to take comments

Messed up my previous forked branch 🤦 (#1035 )

Fixes: #1010


thomas11 (approved):
I'm not too familiar with the Docker provider, whether v3 or v4, but FWIW this looks good.

It seems like this fixes #473?

Could this possibly fix #827 as well? Could you test?

awsx/ecr/image.ts
```diff
@@ -43,48 +43,43 @@ export function computeImageFromAsset(
...
+      target: dockerInputs.target,
+      builderVersion: dockerInputs.builderVersion,
     },
-  );
+    registry: registryCredentials,
+  };
+
+  const image = new docker.Image(`image`, dockerImageArgs, { parent });
+
+  const uniqueImageName = image.imageName;
```
lukehoban, line 82: Pretty sure this introduces a significant regression because `inageName` doesn’t change on rebuilds. So normal examples that use this with ECS will not update the service when the image is re-pushed. 

I believe we have to use `.repoDigest` to get correct behavior.

awsx/schema-types.ts
```diff
@@ -107,11 +107,12 @@ export abstract class Image<TData = any> extends (pulumi.ComponentResource)<TDat
 }
 export interface ImageArgs {
     readonly args?: pulumi.Input<Record<string, pulumi.Input<string>>>;
+    readonly builderVersion?: BuilderVersionInputs;
     readonly cacheFrom?: pulumi.Input<pulumi.Input<string>[]>;
+    readonly context?: pulumi.Input<string>;
     readonly dockerfile?: pulumi.Input<string>;
-    readonly env?: pulumi.Input<Record<string, pulumi.Input<string>>>;
```
lukehoban, line 112: What happened to env?  Isn’t this a breaking change? (And material loss of functionality?)

PR #1065, opened 2023-08-03
Pass `--schema-path` with docs-build requests to pulumi/registry

cnunciato:
~~Because this repo doesn't contain a `schema.json`, these workflow-dispatch calls [always fail](https://github.com/pulumi/registry/actions/runs/5749835730/job/15585412356) (and generate [P1 issues](https://github.com/pulumi/registry/issues/2920) for the docs team), so this change just removes them.~~

Adds a flag to pass the path to `schema.json` when dispatching actions to the Registry in the `release` workflow. 

Fixes https://github.com/pulumi/registry/issues/2920.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -367,19 +367,6 @@ jobs:
       - name: Add SDK version tag
         run: git tag sdk/v$(pulumictl get version --language generic) && git push origin
           sdk/v$(pulumictl get version --language generic)
-  create_docs_build:
-    name: Create docs build
-    needs: tag_sdk
-    runs-on: ubuntu-latest
-    steps:
-      - name: Install pulumictl
-        uses: jaxxstorm/action-install-gh-release@v1.1.0
-        with:
-          repo: pulumi/pulumictl
-      - env:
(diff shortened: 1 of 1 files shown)


thomas11:
@cnunciato actually it does contain a schema.json, but it's at [awsx/schema.json](https://github.com/pulumi/pulumi-awsx/blob/master/awsx/schema.json), not at the root of the repo. Can we accommodate that and make it work instead of disabling it?

cnunciato:
@thomas11 Ah, I didn't see that! Yes, I think we can. Lemme try.

cnunciato:
Looks like that's [already being set](https://github.com/pulumi/registry/blob/eb31af57703ff2eea964a2c040d4ac5301b10d38/themes/default/data/registry/packages/awsx.yaml#L11) -- not sure why it doesn't work. Will take a look.

cnunciato:
Looks like it should be passing `--schema-path` and isn't. I'll adjust. Thanks for the tip!

cnunciato:
@thomas11 All right, updated accordingly -- tested this with an ad-hoc call and it [seems to work](https://github.com/pulumi/registry/actions/runs/5754421588/job/15599667119).

thomas11 (approved):
Thank you!

Issue #1053, opened 2023-07-20
Unable to assign `name`, `appProtocol`, and `containerPortRange` to an ECS Task Definition's `portMappings`


Tomburgs:
### What happened?

It is not possible to assign a `name`, `appProtocol`, and `containerPortRange` on a `portMappings` value on the `ContainerDefinition` type, which is used by `awsx.ecs.EC2TaskDefinition` and `awsx.ecs.FargateTaskDefinition`.

This prevents users from being able to use ECS Service Connect to allow interservice communication.

- [PortMapping AWS Docs](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html)
- [PortMapping Pulumi Definition](https://github.com/pulumi/pulumi-aws/blob/f787a47a2349b5b9024f9ddf1246668539a648a1/sdk/nodejs/ecs/container.ts#L189)

### Expected Behavior

It should be possible to assign all valid `PortMappings` properties through pulumi for an ECS Task Definition.

### Steps to reproduce

Following code should create an ECS task with a port mapping named 'myContainer' and set app protocol to 'grpc', but instead these values will be omitted from the created resource and typescript will complain about invalid properties.

```typescript
const myTask = new awsx.ecs.EC2TaskDefinition('myTask', {
  container: {
    name: 'myContainerName',
    image: myContainerImageUri,
    portMappings: [{
      containerPort: 4200,
      hostPort: 4200,
      protocol: 'tcp',
      name: 'myContainer',
      appProtocol: 'grpc',
    }],
  }
});
```

### Output of `pulumi about`

CLI
Version      3.75.0
Go Version   go1.20.5
Go Compiler  gc

Host
OS       darwin
Version  13.0
Arch     x86_64

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Hi @Tomburgs thanks for reporting this issue! It looks like we're not picking up the latest types available in our upstream dependency [pulumi/pulumi-aws-native](https://github.com/pulumi/pulumi-aws-native) as we're generating our AWSx schema using [v0.13.0 of our AWS Native provider](https://github.com/pulumi/pulumi-awsx/blob/d798487f062748c4bc944c2f178f176fd617f029/schemagen/pkg/gen/schema.go#L31).

[...]

PR #1069, opened 2023-08-07
Bump aws-native types to v0.72.0

rquitales:
This PR upgrades this provider to use types from pulumi/pulumi-aws-native v0.72.0. Previously, we were using types from v0.13.0.

Changes in this PR:

- Use v0.72.0 of aws-native types for ECS
- TaskDefinitionContainerDefinition now requires the name and image fields
- Updated tests to conform to the required fields

Fixes: #1053


awsx/ecs/container.test.ts
```diff
@@ -52,15 +53,4 @@ describe("container naming for single container", () => {
     expect(n.container).toBeUndefined();
   });
 
-  it("single container without name", async () => {
```
mikhailshilkov, line 55: Why do we need to remove this test?

awsx/ecs/ec2TaskDefinition.test.ts
```diff
@@ -45,14 +45,20 @@ describe("validateNetworkMode", () => {
 
   it("the default network mode should be awsvpc", async () => {
     const taskDefinition = new EC2TaskDefinition("default", {
-      container: {},
+      container: {
+        name: "fake-name",
+        image: "fake-image",
+      },
```
mikhailshilkov, line 51: Do users have to add these properties? Won't old code work without doing so? Does it mean this change is breaking?

awsx/schema-types.ts
```diff
@@ -793,15 +793,15 @@ export interface TaskDefinitionContainerDefinitionOutputs {
...
-    readonly image?: pulumi.Output<string>;
+    readonly image: pulumi.Output<string>;
     readonly interactive?: pulumi.Output<boolean>;
     readonly links?: pulumi.Output<string[]>;
     readonly linuxParameters?: pulumi.Output<TaskDefinitionLinuxParametersOutputs>;
     readonly logConfiguration?: pulumi.Output<TaskDefinitionLogConfigurationOutputs>;
     readonly memory?: pulumi.Output<number>;
     readonly memoryReservation?: pulumi.Output<number>;
     readonly mountPoints?: pulumi.Output<TaskDefinitionMountPointOutputs[]>;
-    readonly name?: pulumi.Output<string>;
```
mikhailshilkov, line 804: Can you explain why this changed?

rquitales: The `name` and `image` fields in upstream AWS Native now [marks these as required fields](https://www.pulumi.com/registry/packages/aws-native/api-docs/ecs/taskdefinition/#taskdefinitioncontainerdefinition).

mjeffryes (approved):
Kind of amazing how little the schema changed given with this big of a version jump actually. I think this looks good as long as we document the breaking changes in the changelog. (I see this repo does not use CHANGELOG.md though; so I'm not entirely clear on the process for how we document the changes.)

PR #1079, opened 2023-09-06
tkappler/update ci

thomas11:
- Update actions/checkout to v4, v2 uses unsupported Node
- Update jaxxstorm/action-install-gh-release to v1.10.0, v2 uses unsupported Node
- Update actions/setup-go to v4
- Update actions/upload-artifact to v3
- Update aws-actions/configure-aws-credentials to v3
- Update pulumi/action-install-pulumi-cli to v2
- Update actions/download-artifact to v3
- Undo wrong change to master.yml


.github/workflows/cron.yml
```diff
@@ -30,22 +30,22 @@ jobs:
     continue-on-error: true
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
```
t0yv0, line 33: Curious, is there a good reason this set of actions needs to be different from ci-mgmt standard set of actions? Possibly just lack of time to prioritize folding it back?

.github/actions/install/action.yml
```diff
@@ -0,0 +1,43 @@
+name: Install Languages & Frameworks
```
t0yv0, line 1: These kind of things would be a lot more useful if we could share them across providers.

t0yv0 (approved):

.github/workflows/cron.yml
thomas11: awsx is simply not in ci-mgmt, and I didn't have the time to take that on. Not sure if it's even a great idea because awsx is not a provider in the same sense than the others.

As to the action versions, I simply chose the latest one if they worked.

.github/actions/install/action.yml
thomas11: I shared it, via the `cp` command :)

t0yv0: Duplication is eating away from our future productivity..

.github/workflows/cron.yml
t0yv0: Yeah, it's a provider. It provisions resources, has a schema and publishes to PyPI/NPM. You're right that the tradeoffs need to be carefully considered however I'd not dismiss it out of hand. For a recent example, keeping java versions up to date, migrating to python wheels, these little infra tasks roll trivially through standard providers managed in ci-mgmt but require special work for every provider like this one.

.github/actions/install/action.yml
thomas11: Sure. I reduced duplication here. One step at a time.

PR #1083, opened 2023-09-08
Dependency update: Pulumi 3.81.0

t0yv0:
This is a routine dependency update to Pulumi 3.81.0.


mjeffryes (approved):

sdk/dotnet/Ec2/GetDefaultVpc.cs
```diff
@@ -17,6 +17,12 @@ public static class GetDefaultVpc
         /// </summary>
         public static Task<GetDefaultVpcResult> InvokeAsync(GetDefaultVpcArgs? args = null, InvokeOptions? options = null)
             => global::Pulumi.Deployment.Instance.InvokeAsync<GetDefaultVpcResult>("awsx:ec2:getDefaultVpc", args ?? new GetDefaultVpcArgs(), options.WithDefaults());
+
+        /// <summary>
+        /// [NOT YET IMPLEMENTED] Get the Default VPC for a region.
```
thomas11, line 22: @jkodroff, @t0yv0 , this PR introduced this method to the SDK, although it had always been defined in `pkg/gen/ec2.go`. Is it intended to be in the public SDK?

t0yv0: Sorry for skimming over this in review. Do we need to revert?

danielrbradley: I _think_ we decided that it's okay to leave this in the SDK. The wrapping class is marked as Obsolete to provide a good warning to users. It looks like this was just adding another overload of the method - there's already another version of it above. I wouldn't revert.

PR #1086, opened 2023-09-13
Upgrade pulumi-aws to v6 and pulumi-docker to v4

thomas11:
- Upgrade awsx-classic to awsv6 since it shares dependencies and SDK with awsx
- Remove ecs.Cluster properties that were removed upstream
  - Delete legacy examples that cannot work anymore without ecs.Cluster capacity providers
  - Adjust example to removal of ecs.Cluster capacity providers
- Bump p-java to latest v0.9.7
- Log better errors from package spec handling
- Bump Go SDK to /v2
- Upgrade p/p to 3.81.0


guineveresaenger (commented):
That looks correct to me.

If I read it correctly, `getImageName` is actually more of a "create a unique image name" functionality.

I also really approve of the variable naming here. `canonicalIMageName` is _fantastic_.

.github/workflows/run-acceptance-tests.yml
```diff
@@ -164,6 +164,7 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: ${{ env.GOVERSION }}
+          check-latest: true
```
iwahbe, line 167: What does this change give us? We don't normally pass `check-latest: true` so I'm curious why this is necessary here but not in other places.

iwahbe (approved):
LGTM

interurban (approved):
Sounds like we're prepared from a Docs point of view

Issue #1109, opened 2023-10-19
Dockerfile argument to Image has no effect


mikeshultz:
### What happened?

I'm trying to setup Pulumi (with Python) to build a docker image for ECR to use with ECS using `pulumi_awsx.ecr.Image` using a Dockerfile with a different filename.

The `dockerfile` argument seems to have no effect, and it always tries to build `Dockerfile`.

### Example

I put together a repository with the simplest testcase:  https://github.com/mikeshultz/pulumi-dockerfile-testcase

Building an image like this:

```python
image = Image(
    "pulumi-dockerfile-testcase",
    repository_url=repository.url,
    # Relative
    #dockerfile="Dockerfile.testcase",
    # Absolute
    dockerfile=Path(__file__).parent.joinpath("Dockerfile.testcase").as_posix(),
)
```

Will throw an error like this:

```
Diagnostics:
  docker:index:Image (image):
    error: could not open dockerfile at relative path Dockerfile: stat Dockerfile: no such file or directory

  pulumi:pulumi:Stack (pulumi-dockerfile-testcase-dev):
    error: Resource monitor has terminated, shutting down
```

This also makes it impossible to build multiple image from different Dockerfiles unless they have different contexts.

### Output of `pulumi about`

```
$ pulumi about
CLI          
Version      3.89.0
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME    VERSION
aws     6.6.0
awsx    2.0.2
docker  4.4.3
python  unknown

Host     
OS       arch
Version  
Arch     x86_64

This project is written in python: executable='/home/mike/.venvs/pulumi-testcase/bin/python3' version='3.10.13'

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/mikeshultz
User           mikeshultz
Organizations  mikeshultz
Token type     personal

Dependencies:
NAME         VERSION
pip          23.0.1
pulumi-awsx  2.0.2
setuptools   65.5.0

Pulumi locates its logs in /tmp by default
warning: Failed to get information about the current stack: No current stack
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikeshultz:
For anyone that runs into this, it seems to be possible to get it running using the lower level `pulumi_docker` `Image` construct.  Should look something like this:

```python
from pulumi_docker import DockerBuildArgs, Image, RegistryArgs

[...]

PR #1115, opened 2023-10-23
Pass the dockerfile option to ECR image

mikhailshilkov:
Apparently, we don't pass the `dockerfile` option to the ECR image. This PR is a one-line change to actually use it. I added a new E2E test to validate, let me know if you see a better testing opportunity.

Fix https://github.com/pulumi/pulumi-awsx/issues/1109

### DIFF
--- a/examples/ts-ecr-dockerfile/index.ts
+++ b/examples/ts-ecr-dockerfile/index.ts
@@ -1,0 +1,25 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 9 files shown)


danielrbradley (approved):

examples/ts-ecr-dockerfile/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "fargate",
```
thomas11, line 2: Copy-paste name? Let's rename it and bump nodejs to 18+.

I've also found that I jsut had to change a number of other examples to move awsx from `peerDependencies` to actual `dependencies`, other yarn/npm won't install it.

examples/ts-ecr-dockerfile/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+import * as awsx from "@pulumi/awsx";
+
+const repository = new awsx.ecr.Repository("repository", { forceDelete: true });
+
+export const image = new awsx.ecr.Image("image", {
+  repositoryUrl: repository.repository.repositoryUrl,
+  context: "./app",
+  dockerfile: "./docker/Dockerfile",
```
thomas11, line 24: How does the test fail without your change? Can't find Dockerfile?

mikhailshilkov: It [fails](https://github.com/pulumi/pulumi-awsx/actions/runs/6632554026/job/18018645089?pr=1115#step:15:680) with

> error: could not open dockerfile at relative path app/Dockerfile: stat app/Dockerfile: no such file or directory

Issue #1112, opened 2023-10-23
Timeout while waiting for state to become 'tfSTABLE' (last state: 'tfPENDING', timeout: 20m0s)


t0yv0:
### What happened?

I'm receiving this error a lot when trying to test examples locally:

```

Diagnostics:
  pulumi:pulumi:Stack (ecs-node-p-it-antons-mac-nodejs-69324370):
    error: update failed

  aws:ecs:Service (my-service):
    error: 1 error occurred:
        * creating urn:pulumi:p-it-antons-mac-nodejs-69324370::ecs-node::awsx:ecs:FargateService$aws:ecs/service:Service::my-service: 1 error occurred:
        * waiting for ECS service (arn:aws:ecs:us-west-2:616138583583:service/cluster-e6c5e93/my-service-3c9c1de) to reach steady state after creation: timeout while waiting for state to become 'tfSTABLE' (last state: 'tfPENDING', timeout: 20m0s)

Outputs:
    url: "nginx-lb-f66cb5d-2145136225.us-west-2.elb.amazonaws.com"

Resources:
    + 33 created

Duration: 22m32s
```

This timeout happens when trying to record example baseline behavior, say for ecs/nodejs/ on AWS 5.42.0 and AWSX 1.x.x, but also when running examples on latest versions or the dependencies.  

I have seen this affect the aws:ecs/service:Service through FargateService and other component resource wrappers. 

For users affected by this issue, the current workaround per @danielrbradley is to apply a transformation that increases the custom timeout for the ECS service, see https://github.com/pulumi/pulumi-awsx/pull/1118 for a fully worked out example.

Please upvote this issue if this affects your workflow, and we can consider increasing default timeouts in the AWS provider.

### Example

N/A

### Output of `pulumi about`

```
CLI          
Version      3.86.0
Go Version   go1.21.1
Go Compiler  gc

Host     
OS       darwin
Version  14.0
Arch     x86_64

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/t0yv0
User           t0yv0
Organizations  t0yv0, pulumi
Token type     personal

Pulumi locates its logs in /var/folders/gk/cchgxh512m72f_dmkcc3d09h0000gp/T/ by default
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Possibly related:

[...]

PR #1118, opened 2023-10-24
Extend ECS service create timeout

danielrbradley:
Use a transformation to set the create custom timeout.

Fixes https://github.com/pulumi/pulumi-awsx/issues/1104

### DIFF
--- a/examples/ts-lb-with-subnets/index.ts
+++ b/examples/ts-lb-with-subnets/index.ts
@@ -1,3 +1,4 @@
+import * as pulumi from "@pulumi/pulumi";
 import * as aws from "@pulumi/aws";
 import * as awsx from "@pulumi/awsx";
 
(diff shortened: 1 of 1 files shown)


danielrbradley:
@t0yv0 does this also close #112 ?

t0yv0:
It seems #112 is closed already, not sure.

examples/ts-lb-with-subnets/index.ts
```diff
@@ -10,25 +11,41 @@ const lb = new awsx.lb.ApplicationLoadBalancer("nginx-lb", {
...
+                    },
+                ],
+            },
         },
     },
-});
+    {
+        transformations: [
+            (args) => {
+                if (args.type === "aws:ecs/service:Service") {
```
t0yv0, line 38: Ah, very interesting, so this is transforming the "invisible" resource that's not visible in the program. This is a great but not very obvious workaround.

t0yv0 (approved):

thomas11 (commented):
Wouldn't it be cleaner to expose `timeouts` in p-aws and then use it as a normal property in awsx, [here](https://github.com/pulumi/pulumi-awsx/blob/tkappler/more-upgrade-tests/awsx/ecs/fargateService.ts#L60-L61)? This would fix the problem in other places as well, and allow p-aws users access to the timeout.

t0yv0:
I like that could you throw that into an issue?

danielrbradley:
@thomas11 this is already exposed in the AWS provider ... but as a resource _option_ rather than a resource _argument_. We replicate arguments, but options are meant to be propegated within components automatically by the framework, but customTimeouts hasn't been implemented.

danielrbradley:
> It seems #112 is closed already, not sure.

Oops, missed a "1" ... I was meaning:
- https://github.com/pulumi/pulumi-awsx/issues/1112

t0yv0:
I'd like to keep that open to represent possibly user-facing issue.

t0yv0:
That's tempting.

PR #1126, opened 2023-10-27
Ensure the resource name uses the uniquely generated name to avoid conflicts

JoeStead:
By having `image` hard-coded, pulumi can never succeed with multiple images in the same stack because the name conflicts.

### DIFF
--- a/awsx/ecr/image.ts
+++ b/awsx/ecr/image.ts
@@ -81,7 +81,7 @@ export function computeImageFromAsset(
     registry: registryCredentials,
   };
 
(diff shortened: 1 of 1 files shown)


thomas11 (approved):
Looks good to me, thank you @JoeStead!

Unfortunately, your PR coincided with a regression in the underlying pulumi-aws. It's currently being fixed, then we can get green tests and merge your PR.

thomas11:
Hi @JoeStead & @pierskarsenbarg, unfortunately we had to revert the PR since it would be a breaking change, causing existing images with the `image` name to be replaced. I filed #1142 to keep track of the issue.

PR #1128, opened 2023-10-27
Fix up dependencies of examples/ecs/python

thomas11:


examples/ecs/python/requirements.txt
```diff
@@ -1,2 +1,3 @@
 pulumi>=3.0.0,<4.0.0
-pulumi_aws>=5.0.0,<6.0.0
+pulumi_aws>=6.0.2,<7.0.0
+pulumi_awsx>=2.0.2,<3.0.0
```
pierskarsenbarg, line 3: Isn't it supposed to be testing this so it gets this but not the AWS dependency from the provider on `PATH`?

PR #1134, opened 2023-10-30
Use imageName for the docker image

garyhuntddn:
When attempting to build and push multiple docker images to ECR I was getting a duplicate resource name issue.

This appeared to be because the docker build was always using a name of "index".

Using the created imageName variable has fixed my issue and seems a better solution that the fixed name of "index"


thomas11:
Hi @garyhuntddn, thank you for your submission! Your fix is correct, but was already applied by #1126 in the meantime. Feel free to re-open if I think I missed something.

PR #1139, opened 2023-10-31
Fix makefile

danielrbradley:
Avoid file dependencies on either committed files or directories.

For context see [Handling Non-File Targets](https://dev.to/danielrbradley/better-builds-with-make-a-beginners-guide-1n0#handling-nonfile-targets)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,20 +130,21 @@ bin/pulumi-java-gen::
 		go test -v ./... -check.vv
 	@touch $@
 
-sdk/dotnet/bin:: DOTNET_VERSION := $(shell pulumictl get version --language dotnet)
-sdk/dotnet/bin:: bin/${CODEGEN} .make/schema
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -215,7 +226,7 @@ test:: bin/gotestfmt
 	fi
 
 schemagen:: bin/${CODEGEN}
-schema: awsx/schema.json
+schema: .make/schema
```
thomas11, line 229: Do we need this sentinel even though the schema is just a single file?

thomas11 (commented):
LGTM so far!

Makefile
danielrbradley: Yup - because it's a committed file, so sometime the timestamp might be updated during checkout, but it's not actually reflecting the newly checked-out code, so it should be re-generation but isn't due to the timestamp.

The short rule is: if it's committed or a directory, use a sentinel file.

thomas11 (approved):

PR #1119, opened 2023-10-24
Add more upgrade tests

thomas11:

PR #1140, opened 2023-11-01
Enable more VPC upgrade tests

mikhailshilkov:
Adds a few more snapshot tests

Supersedes https://github.com/pulumi/pulumi-awsx/pull/1119 (rebased on 2.1.0)


examples/ts-lb-simple/index.ts
```diff
@@ -23,6 +28,14 @@ const service = new awsx.ecs.FargateService("my-service", {
...
                 },
             ],
+            logConfiguration: {
+                logDriver: "awslogs",
+                options: {
+                    "awslogs-group": "my-log-group",
+                    "awslogs-region": "us-west-2",
+                    "awslogs-stream-prefix": "my-app",
+                },
+            },
```
danielrbradley, line 38: Do we need these changes? cc: @thomas11

danielrbradley (approved):
Looks fine, possibly revert the changes to the task definition if they're not needed, but probably fine either way.

examples/ts-lb-simple/index.ts
thomas11: Without them, the container wouldn't start for me when running the tests locally, with error

```
There was an error while retrieving logs from log stream: nginx/nginx/8c06c430fc784875b8f5c28c0e9b5101 in log group: my-service-d975f63.
The specified log stream does not exist.
```

I didn't fully chase it down. Since it seems to work in CI, another region/account difference?

PR #1143, opened 2023-11-01
Record ts-lb-attach-ec2 upgrade test

danielrbradley:
- Replace peer-dependency with normal dependency.
- This adds coverage of the `awsx:lb:TargetGroupAttachment` component.


thomas11 (approved):
A bit painful to rework the test around the upgrade testing but LGTM for now.

PR #1145, opened 2023-11-02
Revert "Revert "Ensure the resource name uses the uniquely generated name to avoid conflicts (#1126)" (#1141)"

thomas11:
This reverts commit af69664ec8c6a46edfdbd4cca079d3c652381015.

Revive PR #1126 by @JoeStead since it was actually sound. The hard-coded image name was only introduced in v2, so backwards compatibility is not a concern.


rlukata:
Thank you so much!

Issue #1148, opened 2023-11-03
extraOptions is ignored when using awsx.ecr.Image


rlukata:
### What happened?

extraOptions is ignored when using awsx.ecr.Image

### Example

```
const image = new awsx.ecr.Image(`docker-image`, {
        repositoryUrl: repo.repository.repositoryUrl,
        context: context,
        dockerfile: dockerFile,
        platform: 'linux/amd64',
        builderVersion: awsx.ecr.BuilderVersion.BuilderBuildKit,
        extraOptions: ['--quiet']
    });
```

### Output of `pulumi about`

```
CLI
Version      3.86.0
Go Version   go1.21.1
Go Compiler  gc

Plugins
NAME            VERSION
aws             5.42.0
aws             5.42.0
aws-apigateway  1.0.1
awsx            2.1.1
command         0.8.2
docker          4.4.4
nodejs          unknown
tls             4.10.0

Host
OS       darwin
Version  14.1
Arch     arm64
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
It turns out that `extraOptions` are supported in pulumi-docker V4 anymore: https://github.com/pulumi/pulumi-docker/issues/424

We should have deprecated/removed the `extraOptions` on ecr:Image. We should still probably do it, since we aren't breaking any existing V2 behavior (V1 behavior is already broken in the new major version). @danielrbradley @thomas11 what do you think?

thomas11:
> It turns out that `extraOptions` are not supported in pulumi-docker V4 anymore: [pulumi/pulumi-docker#424](https://github.com/pulumi/pulumi-docker/issues/424)
> 
> We should have deprecated/removed the `extraOptions` on ecr:Image. We should still probably do it, since we aren't breaking any existing V2 behavior (V1 behavior is already broken in the new major version). @danielrbradley @thomas11 what do you think?

@mikhailshilkov yes, I think removing something that isn't working can be considered a non-breaking change, we have done this before.

PR #1150, opened 2023-11-09
Remove awsx.Ecr.ImageArgs.extraOptions

thomas11:
This would be a breaking change, but this argument is already not
working anymore since it was removed from the underlying Docker provider,
see pulumi/pulumi-docker#424.

Part of #1148

### DIFF
--- a/awsx/schema-types.ts
+++ b/awsx/schema-types.ts
@@ -112,7 +112,6 @@ export interface ImageArgs {
     readonly cacheFrom?: pulumi.Input<pulumi.Input<string>[]>;
     readonly context?: pulumi.Input<string>;
     readonly dockerfile?: pulumi.Input<string>;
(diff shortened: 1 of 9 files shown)


joeduffy:
Should we deprecate this instead for now?

thomas11:
> Should we deprecate this instead for now?

@joeduffy we certainly could. In my mind, deprecation is for something that still works but will not or might not work in the future. Here, `extraOptions` are already broken so the existence of the property is misleading. Do you prefer deprecation anyway?

mjeffryes (approved):
This is an unfortunate side effect of the breaking changes Docker v4.0.0, but I agree that it's probably better to remove so the behavior is explicit. Technically, this is breaking and should be released as a major version bump, however, there's a reasonable argument to be made that it's better to cause the break than leave users on a version that silently ignores the setting.

Issue #1057, opened 2023-08-01
apiKeySource not updated in awsx.classic.apigateway.API


albertocorrales:
### What happened?

In order to implement tenant throttling on API Gateway, we had to update our APIs to set the setting `apiKeySource` to `AUTHORIZER`.

We were using the pulumi/awsx package to create our APIs and when we updated the setting `apiKeySource` we realized the change was not applied on AWS.

### Expected Behavior

When we update an existing API and we set the property `apiKeySource` as `AUTHORIZER`, this change should be applied on AWS.

### Steps to reproduce

Create an API as:

``` javascript
import * as awsx from "@pulumi/awsx";

let endpoint = new awsx.classic.apigateway.API("example", {
  routes: [
    {
      path: "/",
      method: "GET",

      eventHandler: async (event) => {
        // This code runs in an AWS Lambda and will be invoked any time `/` is hit.
        return {
          statusCode: 200,
          body: "hello",
        };
      },
    },
  ],
});
```

Update the `apiKeySource` as `AUTHORIZER`:

``` javascript
import * as awsx from "@pulumi/awsx";

let endpoint = new awsx.classic.apigateway.API("example", {
  apiKeySource: "AUTHORIZER",
  routes: [
    {
      path: "/",
      method: "GET",

      eventHandler: async (event) => {
        // This code runs in an AWS Lambda and will be invoked any time `/` is hit.
        return {
          statusCode: 200,
          body: "hello",
        };
      },
    },
  ],
});
```

### Output of `pulumi about`

```
CLI
Version      3.74.0  
Go Version   go1.20.5
Go Compiler  gc      

Plugins        
NAME    VERSION
aws     5.42.0 
awsx    1.0.2  
docker  3.6.1  
nodejs  unknown

Host
OS       Microsoft Windows 10 Enterprise
Version  10.0.19045 Build 19045
Arch     x86_64

This project is written in nodejs: executable='C:\Program Files\nodejs\node.exe' version='v18.13.0'

Current Stack: Alberto-Corrales/ApiGatewayAwsx/dev

TYPE                                               URN
pulumi:pulumi:Stack                                urn:pulumi:dev::ApiGatewayAwsx::pulumi:pulumi:Stack::ApiGatewayAwsx-dev
aws:apigateway:x:API                               urn:pulumi:dev::ApiGatewayAwsx::aws:apigateway:x:API::example
[...]

PR #1159, opened 2023-11-20
Correctly support apiKeySource in classic provider

lukehoban:
We were previously passing apiKeySource only via Swagger.  Even though AWS API Gateway does include the API key source in the swagger on output, it apparently does not read it from the swagger on updates.

However, it does support passing it outside of the Swagger, so we just pass this along directly to the RestAPI constructor is specified.

We could in principle *stop* sending it in the Swagger as well (since I don't think it's read from there at all), but I am avoiding changing that in case it has other subtle impacts on behaviour.

Fixes https://github.com/pulumi/pulumi-awsx/issues/1057.


thomas11:
Thank you @lukehoban! Your PR description doesn't say, so I gotta ask, did you test this change manually?

It would be awesome to have a two-step test in `examples/` that repros the issue submitter's issue with editing the `apiKeySource` after `up` but the effort might not be warranted for an uncommon issue in awsx-classic.

lukehoban:
> It would be awesome to have a two-step test in examples/ that repros the issue submitter's issue with editing the apiKeySource after up but the effort might not be warranted for an uncommon issue in awsx-classic.

Agreed - I'll get a test added before we merge this.

lukehoban:
@thomas11 Test added - ready for review.

mjeffryes (approved):
Thanks, Luke! I've filed an issue for us to port this fix into pulumi-aws-apigateway as well

lukehoban:
> I've filed an issue for us to port this fix into pulumi-aws-apigateway as well

Great. I’ve got a PR nearly ready for that - so happy to finish that up once this is merged.

thomas11 (approved):

lukehoban:
> I’ve got a PR nearly ready for that - so happy to finish that up once this is merged.

Tracked in https://github.com/pulumi/pulumi-aws-apigateway/pull/112.

PR #1164, opened 2023-11-22
Add HTTP API component

danielrbradley:
Encapsulates:
1. The root API resource
2. A map of routes which can reference integrations and authorizers. The key is the route e.g. `GET /pets`
3. Optional map of integrations
4. Optional map of authorizers
5. Optional map of stages (defaults to a single `default` stage which will autodeploy)
6. Optional map of domain mappings. The key is the domain name to either create or link to.


lukehoban:
A few questions before reviewing details here:
1. How do we decide whether this lives in `awsx` or `aws-apigateway`?  I can see arguments for both, but generally feels like we need articulated principles for these and future new components
2. I believe https://github.com/pulumi/pulumi-aws-apigateway/issues/6 is the issue tracking this work? (Where we could discuss the approach to addressing). 
3. What would be our content goal once this is landed?  Would we update the serverless arch templates to use this?  Would we update the AWS guides to use this?  Will we update examples to use this?

pierskarsenbarg:
Couple of things:

1. Can we avoid AWS leakage by not having the v2 in the module name? This is an abstraction so doesn't need to match the underlying provider (plus people will ask if the aws-api-gateway package is going to be decommissioned and what should they use for rest api instead). We could therefore have `awsx.apigateway.HttpApi` and `awsx.apigateway.RestApi` which feels a lot better
2. Can we either lift and shift or rewrite the `aws-apigateway` into AWSx and actually archive the old one? If what you're suggesting means that this becomes more discoverable then that's good, right?

PR #1169, opened 2023-11-30
Fix getMappingInputs removing valid arguments

### LINKED ISSUE #1168: ecs task definition port mappings not applying all options
### What happened?

ecs task definitions are not applying most of the port mapping options, such as `name`, `appProtocol`, `containerPortRange`, and [some others](https://github.com/pulumi/pulumi-awsx/blob/master/awsx/schema-types.ts#L951).

### Example

setting port mappings like so on either an EC2 or Fargate task definition will result in an unnamed port definition being created, despite `name` being passed

```typescript
portMappings: [
  {
    name: 'my-port-name',
    containerPort: 80,
  },
],
```

this causes issues with service connect which **needs** port mappings to be named. this makes service connect unusable with pulumi.

### Output of `pulumi about`

CLI
Version      3.94.2
Go Version   go1.21.3
Go Compiler  gc

Host
OS       darwin
Version  14.0
Arch     arm64

### Additional context

_No response_

### Contributing

seems like these lines here are the culprits: https://github.com/pulumi/pulumi-awsx/blob/9f5b18b7be1b11f99d7d0dc4d1438bfd996bf996/awsx/ecs/containers.ts#L99C1-L108C2

Tomburgs:
Fixes #1168

### DIFF
--- a/awsx/ecs/container.test.ts
+++ b/awsx/ecs/container.test.ts
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
 
 import { normalizeTaskDefinitionContainers, getMappingInputs } from "./containers";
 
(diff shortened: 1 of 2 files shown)


Tomburgs:
@mjeffryes noticed that the acceptance tests are failing due to missing aws creds. is there something expected from my side to aid in resolving this?

mjeffryes:
> @mjeffryes noticed that the acceptance tests are failing due to missing aws creds. is there something expected from my side to aid in resolving this?

No, I'll work to sort this out, apologies for the delay!

mjeffryes (approved):
This seems like a reasonable change that will give users more control.

PR #1189, opened 2024-01-02
Fix a broken example due to missing AMI

blampe:
This example was failing because v1 Amazon Linux AMIs no longer exist.

https://aws.amazon.com/blogs/aws/update-on-amazon-linux-ami-end-of-life/

Refs https://github.com/pulumi/pulumi-aws/pull/3207
Fixes https://github.com/pulumi/pulumi-awsx/issues/1188


mjeffryes (approved):
Looks like you'll need to rerecord the state for the upgrade test, but otherwise looks good

PR #1192, opened 2024-01-03
Fix pypi credentials

blampe:
Our `publish_sdk` workflow is currently [broken](https://github.com/pulumi/pulumi-awsx/actions/runs/7400736741) because pypi no longer accepts passwords.

Unclear why we weren't alerted for this workflow failure.

Refs https://github.com/pulumi/ci-mgmt/issues/751
Refs https://github.com/pulumi/ci-mgmt/pull/767


mjeffryes (approved):
Thanks for catching these!

guineveresaenger (approved):

iwahbe:
It would be great to standardize on https://github.com/pulumi/pulumi-package-publisher so we can have a single source of truth for failure alerting.

Half the battle is getting alerted if something goes wrong.

PR #1198, opened 2024-01-10
Fix upgrade tests

blampe:
This AMI query was not pinned, so example tests started breaking when a new AMI was released.

I'm somewhat surprised these tests are recorded but still hit the live provider. I would have expected the test to use recorded requests/responses and only fail if a request wasn't recognized.


t0yv0:
We don't have that capability. They hit live provider with preview. Recording is for the baseline state.

t0yv0 (approved):

danielrbradley (approved):
This looks good for a quick fix.

The intention is to call the real configure, check and diff methods but replay invokes. There's no implementation yet for replaying the invokes, but this is now possible in the new testing library using the intercepting provider proxy. The only piece missing is the logic to pick which recorded response to use.

blampe:
> We don't have that capability. They hit live provider with preview. Recording is for the baseline state.

Is there a reason we don't also record preview requests, or is that just something we haven't gotten to yet? (I hope the latter!)

t0yv0:
You don't want to lose the system under test entirely by replacing it with a test double. At the very least we're interested in Diff call to be executed against a real provider. Now, in practice it's all coupled via internal stateful interaction as Configure sets up state that Diff might use, so we need to dispatch more calls to the real provider than just Diff. Found the hard way.

blampe:
> You don't want to lose the system under test entirely by replacing it with a test double. At the very least we're interested in Diff call to be executed against a real provider. Now, in practice it's all coupled via internal stateful interaction as Configure sets up state that Diff might use, so we need to dispatch more calls to the real provider than just Diff. Found the hard way.

@t0yv0 @danielrbradley to be clear I'm not saying we can get rid of tests against the provider, only that they don't need to happen in the critical path. If my tests pass today on SHA `123ABC`, then they should continue to pass even if I check out `123ABC` ten years from now. Or if I make a change that has zero impact on how I invoke the provider's APIs, then I should be able to assume the API contract is still satisfied.

As it stands today we don't have those guarantees -- provider APIs change, AMIs get removed, etc. Given the complexity of our CI pipelines and the number of repos involved, this non-reproducibility means we effectively _always_ have build failures lurking. I've definitely spent more time this week fixing build failures than actual product work, and that's a bummer.

We still need tests against the provider, but that can happen out of band as a periodic job.

t0yv0:
Yup, this resonates with prior discussions. That's a wonderful point but we're lacking a design that can give those guarantees. We need some form of cloud call intercept per-provider that's not standardized by Pulumi for that. We *could* do it but it's non-trivial work.

PR #1234, opened 2024-03-03
Support ECR images on localstack

ferguscan:
## The Issue

`aws.ecr.Image` makes assumptions about the ID of an ECR reigstry.  It assumes the registry ID [can be extracted from the repository URL](https://github.com/pulumi/pulumi-awsx/blob/9f18b3866d2a16a59d662384a0c745e2e5603463/awsx/ecr/image.ts#L36).

This assumption seems to hold true for the real AWS APIs, but I don't think AWS actually specify this behaviour in any of their documentation.

The assumption **does not** hold true for localstack, leading to [this issue](https://github.com/localstack/localstack/issues/7186).

## The Fix

This PR allows the caller to (optionally) specify a `registryId`, avoiding the inference logic mentioned above.

## Notes

I edited `awsx/schema-types.ts` by hand.  I couldn't figure out how to generate that file.

### DIFF
--- a/awsx/ecr/image.ts
+++ b/awsx/ecr/image.ts
@@ -30,10 +30,10 @@ export function computeImageFromAsset(
   args: pulumi.Unwrap<schema.ImageArgs>,
   parent: pulumi.Resource,
 ) {
-  const { repositoryUrl, imageTag, ...dockerInputs } = args ?? {};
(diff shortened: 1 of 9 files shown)


mjeffryes (approved):
This seems like a reasonable work around to me. Thanks for the contribution!

PR #1245, opened 2024-03-26
Fix order of code generation in schema-types.ts

t0yv0:
Fixes https://github.com/pulumi/pulumi-awsx/issues/1244

I've ran `make build provider` locally and it had the same diff as the CI.


corymhall (approved):

t0yv0:
Yes looks like one of the hand-written TypeScript files is not passing linter. Trying again.

corymhall (approved):

PR #1255, opened 2024-04-04
Update classic vpc nat gateway to use `domain: "vpc"` instead of deprecated `vpc: true`

eigilsagafos:


t0yv0:
This looks good, thank you @eigilsagafos ! It seems to address a deprecation. Let me try to run tests here.

eigilsagafos:
@t0yv0 I updated based on the failing lint test. Could you trigger the tests again?

t0yv0:
Let's try it. Thank you!

t0yv0 (approved):

PR #1257, opened 2024-04-09
Add imageName param to ECR image args

### LINKED ISSUE #1171: Building different targets on same Dockerfile causes a Duplicate resource URN error
### What happened?

I have a Dockerfile with two targets in it (migrator and production). I need to build both of these targets on every pulumi up. So I have two instances of awsx.acr.Image which are very similar, but have different names, and use different targets.

However, when I do this, I get the error "Duplicate resource URN"

This seems to be related to how the `createUniqueImageName` function works here: https://github.com/pulumi/pulumi-awsx/blob/52e698db4d85d7fd4fa292027d04766ce4b1ceba/awsx/ecr/image.ts#L93

Since all it considers is context, dockerfile, and args (and not resource name, or target) these two resource definitions end up with the same image name.

### Example

```ts
export const migratorRepo = new awsx.ecr.Repository(`${stackName}-migrator`, {
  lifecyclePolicy: {
    rules: [{ maximumNumberOfImages: 25, tagStatus: 'any' }],
  },
});

export const appsRepo = new awsx.ecr.Repository(`${stackName}-api`, {
  lifecyclePolicy: {
    rules: [{ maximumNumberOfImages: 25, tagStatus: 'any' }],
  },
});

const migratorImage = new awsx.ecr.Image(PREFIX + 'migrator', {
  repositoryUrl: migratorRepo.url,
  context: path.join(__dirname, '../../../../'),
  dockerfile: path.join(__dirname, '../../', 'Dockerfile'),
  platform: 'linux/amd64',
  target: 'migration',
  cacheFrom: ["base"],
  args: {
    BUILDKIT_INLINE_CACHE: "1",
  },
});

const apiImage = new awsx.ecr.Image(PREFIX + 'api', {
  repositoryUrl: appsRepo.url,
  context: path.join(__dirname, '../../../../'),
  dockerfile: path.join(__dirname, '../../', 'Dockerfile'),
  platform: 'linux/amd64',
  target: 'production',
  cacheFrom: ["base"],
  args: {
    BUILDKIT_INLINE_CACHE: "1",
  },
});
```

### Output of `pulumi about`

CLI          
Version      3.76.1
Go Version   go1.20.6
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host     
OS       darwin
Version  13.6
Arch     arm64

### Additional context

My guess is that at the very least the `createUniqueImageName` function should factor in things like target, platform, and maybe even cacheFrom.

### Contributing

sav-valerio:
Fixes #1171

This allows the user to manually choose the name of the underlying Pulumi Docker image resource

When not defined it will fall back to the original behavior (use the image tag as the name of the underlying resource)

### DIFF
--- a/awsx/ecr/image.ts
+++ b/awsx/ecr/image.ts
@@ -37,7 +37,9 @@ export function computeImageFromAsset(
 
   pulumi.log.debug(`Building container image at '${JSON.stringify(dockerInputs)}'`, parent);
 
-  const imageName = imageTag ? imageTag : createUniqueImageName(dockerInputs);
+  const imageName = args.imageName ? args.imageName :
+    (imageTag ? imageTag : createUniqueImageName(dockerInputs));
+
   // Note: the tag, if provided, is included in the image name.
   const canonicalImageName = `${repositoryUrl}:${imageName}`;
 
--- a/awsx/ecr/image.ts
+++ b/awsx/ecr/image.ts
@@ -37,8 +37,11 @@ export function computeImageFromAsset(
 
   pulumi.log.debug(`Building container image at '${JSON.stringify(dockerInputs)}'`, parent);
 
-  const imageName = args.imageName ? args.imageName :
-    (imageTag ? imageTag : createUniqueImageName(dockerInputs));
+  const imageName = args.imageName
+    ? args.imageName
+    : imageTag
+    ? imageTag
+    : createUniqueImageName(dockerInputs);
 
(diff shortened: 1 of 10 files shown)


t0yv0:
The failing test is a known flake, skipping in https://github.com/pulumi/pulumi-awsx/pull/1260

LGTM, thanks so much for this contribution!

t0yv0 (approved):

PR #1263, opened 2024-04-17
Update codegen

t0yv0:
Upgrades Pulumi core code-generators and SDK dependency from v3.99.0 to v3.113.0.

Fixes https://github.com/pulumi/pulumi-awsx/issues/1262


t0yv0:
I think upgrade tests are downloading the AWS provider once per test and not cleaning up. This is the reason we hit out of disk space. I was overdue for upgrading these to the latest recommended providertest version so I will attempt that and also try to make sure they clean up.

t0yv0:
closing in favor of https://github.com/pulumi/pulumi-awsx/pull/1266

Issue #1197, opened 2024-01-10
policyArns in RoleWithPolicyArgs should be input


dhilgarth:
### What happened?

Right now, the `policyArns` property of `RoleWithPolicyArgs`  is a simple string: https://github.com/pulumi/pulumi-awsx/blob/master/sdk/nodejs/types/input.ts#L309
This forces us to create the service within `apply`, if we want to use a policy that we also create in the same program.

### Example

What I would expect to work:

```
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx"

const policy = new aws.iam.Policy("put-metrics", {
    policy: aws.iam.getPolicyDocumentOutput({
        statements: [{
            actions: ["cloudwatch:PutMetricData"]
        }]
    }).json
});

const service = new awsx.ecs.FargateService("sample", {
    taskDefinitionArgs: {
        container: {
            image: "nginx",
            name: "sample"
        },
        taskRole: {
            args: {
                policyArns: [policy.arn]
            }
        }
    }
})
```

What I have to do instead:

```
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx"

const policy = new aws.iam.Policy("put-metrics", {
    policy: aws.iam.getPolicyDocumentOutput({
        statements: [{
            actions: ["cloudwatch:PutMetricData"]
        }]
    }).json
});

const service = policy.arn.apply(policyArn => new awsx.ecs.FargateService("sample", {
    taskDefinitionArgs: {
        container: {
            image: "nginx",
            name: "sample"
        },
        taskRole: {
            args: {
                policyArns: [policyArn]
            }
        }
    }
}));
```

### Output of `pulumi about`

CLI          
Version      3.99.0
Go Version   go1.21.5
Go Compiler  gc

Plugins
NAME    VERSION
aws     6.18.0
awsx    2.4.0
docker  4.5.0
docker  3.6.1
nodejs  unknown

Host     
OS       Microsoft Windows 11 Pro
Version  10.0.22631 Build 22631
Arch     x86_64

This project is written in nodejs: executable='C:\Program Files\nodejs\node.exe' version='v20.9.0'

Backend        
Name           laptop-dh
URL            file://~
User           AzureAD\DanielHilgarth
Organizations  
Token type     personal

Pulumi locates its logs in C:\Users\DANIEL~1\AppData\Local\Temp by default

### Contributing

Vote on this issue by adding a 👍 reaction. 
[...]

PR #1276, opened 2024-04-24
Change policyArns in RoleWithPolicyArgs to accept eventual types

flostadler:
Previously the `policyArns` attribute in RoleWithPolicyArgs only accepted plain types as inputs. This forced users to use workarounds like using `apply` in order to configure policies created in the same pulumi program.
This PR changes the `policyArns` attribute from a plain string array to an nested Input array (`Input<Input<string>[]>`) to allow users to pass the whole list as well as individual elements as eventual types.

Fixes https://github.com/pulumi/pulumi-awsx/issues/1197

### DIFF
--- a/awsx/role.ts
+++ b/awsx/role.ts
@@ -69,7 +69,7 @@ export function defaultRoleWithPolicies(
 ): {
   roleArn?: pulumi.Output<string>;
   role?: aws.iam.Role;
-  policies?: aws.iam.RolePolicyAttachment[];
+  policies?: pulumi.Output<aws.iam.RolePolicyAttachment[]>;
 } {
   if (inputs?.roleArn !== undefined && inputs?.args !== undefined) {
     throw new Error("Can't define role args if specified an existing role ARN");
@@ -90,13 +90,17 @@ export function defaultRoleWithPolicies(
   delete roleArgs.policyArns;
 
   const role = new aws.iam.Role(name, roleArgs, opts);
-  const policies = args.policyArns?.map(
-    (policyArn) =>
-      new aws.iam.RolePolicyAttachment(
-        `${name}-${utils.sha1hash(policyArn)}`,
-        { role: role.name, policyArn },
-        opts,
+
+  const policies = args.policyArns ?
+    pulumi.Output.create(args.policyArns).apply(unwrapped =>
+      unwrapped.map(policyArn =>
+        new aws.iam.RolePolicyAttachment(
+          `${name}-${utils.sha1hash(policyArn)}`,
+          { role: role.name, policyArn },
+          opts,
+        ),
       ),
-  );
+    ) : undefined;
+
   return { role, policies, roleArn: role.arn };
 }
--- a/awsx/role.ts
+++ b/awsx/role.ts
@@ -91,16 +91,18 @@ export function defaultRoleWithPolicies(
 
   const role = new aws.iam.Role(name, roleArgs, opts);
 
-  const policies = args.policyArns ?
-    pulumi.Output.create(args.policyArns).apply(unwrapped =>
-      unwrapped.map(policyArn =>
-        new aws.iam.RolePolicyAttachment(
-          `${name}-${utils.sha1hash(policyArn)}`,
-          { role: role.name, policyArn },
-          opts,
+  const policies = args.policyArns
+    ? pulumi.Output.create(args.policyArns).apply((unwrapped) =>
+        unwrapped.map(
+          (policyArn) =>
+            new aws.iam.RolePolicyAttachment(
+              `${name}-${utils.sha1hash(policyArn)}`,
+              { role: role.name, policyArn },
+              opts,
+            ),
         ),
-      ),
-    ) : undefined;
+      )
+    : undefined;
 
   return { role, policies, roleArn: role.arn };
 }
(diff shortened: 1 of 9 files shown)


sdk/dotnet/Awsx/Inputs/RoleWithPolicyArgs.cs
```diff
@@ -78,14 +78,14 @@ public InputList<string> ManagedPolicyArns
...
 
         [Input("policyArns")]
-        private List<string>? _policyArns;
+        private InputList<string>? _policyArns;
 
         /// <summary>
         /// ARNs of the policies to attach to the created role.
         /// </summary>
-        public List<string> PolicyArns
+        public InputList<string> PolicyArns
```
t0yv0, line 86: `public sealed class InputList<T> : Input<ImmutableArray<T>>, IEnumerable, IAsyncEnumerable<Input<T>>`

This is breaking I think.

sdk/go/awsx/awsx/pulumiTypes.go
```diff
@@ -2129,7 +2129,7 @@ type RoleWithPolicyArgs struct {
 	// ARN of the policy that is used to set the permissions boundary for the role.
 	PermissionsBoundary pulumi.StringPtrInput `pulumi:"permissionsBoundary"`
 	// ARNs of the policies to attach to the created role.
-	PolicyArns []string `pulumi:"policyArns"`
+	PolicyArns pulumi.StringArrayInput `pulumi:"policyArns"`
```
t0yv0, line 2132: This is certainly breaking.

sdk/java/src/main/java/com/pulumi/awsx/awsx/inputs/RoleWithPolicyArgs.java
```diff
@@ -412,11 +412,21 @@ public Builder permissionsBoundary(String permissionsBoundary) {
...
             return this;
         }
 
+        /**
+         * @param policyArns ARNs of the policies to attach to the created role.
+         * 
+         * @return builder
+         * 
+         */
+        public Builder policyArns(List<String> policyArns) {
```
t0yv0, line 426: This is not breaking because a new overload is generated with the same sig.

flostadler:
@t0yv0  You're absolutely correct about the breaking changes. I had started a thread about this on the original issue after I noticed it (sorry for not updating the PR yet!): https://github.com/pulumi/pulumi-awsx/issues/1197#issuecomment-2075380330

I'm thinking about adding a new parameter and deprecating the old one, do you have any other ideas to add this quality of life improvement without causing pain for other users?

awsx/role.ts
```diff
@@ -69,7 +69,7 @@ export function defaultRoleWithPolicies(
 ): {
   roleArn?: pulumi.Output<string>;
   role?: aws.iam.Role;
-  policies?: aws.iam.RolePolicyAttachment[];
+  policies?: pulumi.Output<aws.iam.RolePolicyAttachment[]>;
```
t0yv0, line 72: I thought there was some magic way to do this change for the SDKs while still receiving un-wrapped values in the provider but I may be mistaken. If this works it's good. 

I slightly worry about provisioning resources under apply inside the component resource, but I might need some platform team help to find references to how this works. 

https://www.pulumi.com/docs/concepts/resources/components/#registering-component-outputs

Normally there is a call registerOutputs at the end of the constructor.

We don't seem to be calling it for ec2TaskDefinition:

https://github.com/pulumi/pulumi-awsx/blob/master/awsx/ecs/ec2TaskDefinition.ts#L32

If we're doing apply then children may register out-of-order that is after the component resource is done constructing and I vaguely recall this may cause bugs.

t0yv0:
Since examples/tests aren't changing looks like this may not be under test yet, would it be possible to add a quick end-to-end example using this?

t0yv0:
Re: slightly breaking changes like this, I need a reminder CC @mjeffryes on what our breaking change policy is here. It sounds like change is not breaking for TypeScript which may account for most of the user base. I agree with you that if we're strict on breaking changes then a new added property is a good alternative.

sdk/dotnet/Awsx/Inputs/RoleWithPolicyArgs.cs
mjeffryes: I *think* dotnet has an implict conversion from `List<T>` to `InputList<T>`

sdk/nodejs/types/input.ts
```diff
@@ -310,7 +310,7 @@ export namespace awsx {
         /**
          * ARNs of the policies to attach to the created role.
          */
-        policyArns?: string[];
+        policyArns?: pulumi.Input<pulumi.Input<string>[]>;
```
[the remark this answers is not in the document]
flostadler: Yeah in typescript it is fine

sdk/dotnet/Awsx/Inputs/RoleWithPolicyArgs.cs
flostadler: I will try it out just to be sure. I was also under the impression, but better safe than sorry.

flostadler:
> Since examples/tests aren't changing looks like this may not be under test yet, would it be possible to add a quick end-to-end example using this?

Yes for sure!

sdk/dotnet/Awsx/Inputs/RoleWithPolicyArgs.cs
t0yv0: Oh that's true, I forgot about implicit conversions. That may be nice.

flostadler:
We're planning this for the next major version upgrade. Closing the PR for now, we can pick it up again once we're working on the next major version

Issue #1072, opened 2023-08-14
ecr.NewImage CacheFrom adds --target to docker build


afreakk:
### What happened?

When adding  `CacheFrom` attribute on `ImageArgs` it also adds  `--target` to `docker build.
Which causes:
```
error: Error: The target "123.dkr.ecr.eu-central-1.amazonaws.com/myImage:latest" was not found in the provided Dockerfile
```

```
 Error: 
docker build --cache-from 123.dkr.ecr.eu-central-1.amazonaws.com/myImage /a/b/c/myImage -t \
46b0e6fe-container-123.dkr.ecr.eu-central-1.amazonaws.com/myImage:latest \
--target 123.dkr.ecr.eu-central-1.amazonaws.com/myImage:latest
ailed with exit code 125
        at /home/afreak/.pulumi/plugins/resource-awsx-v1.0.0-beta.2/node_modules/@pulumi/docker.ts:592:15
        at Generator.next (<anonymous>)
        at fulfilled (/home/afreak/.pulumi/plugins/resource-awsx-v1.0.0-beta.2/node_modules/@pulumi/docker/docker.js:18:58)
        at processTicksAndRejections (node:internal/process/task_queues:95:5)

```

This is the code being  used:
```
package main

import (
	"github.com/pulumi/pulumi-awsx/sdk/go/awsx/ecr"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		_, err := ecr.NewImage(ctx, "test", &ecr.ImageArgs{
			RepositoryUrl: pulumi.String("123.dkr.ecr.eu-central-1.amazonaws.com/myImage"),
			Path:          pulumi.String("/home/afreak/myImage"),
			CacheFrom:     pulumi.StringArray{pulumi.String("123.dkr.ecr.eu-central-1.amazonaws.com/myImage:latest")},
		})
		return err
	})
}

```

### Expected Behavior

`docker build` should not have the `--target` flag, when I have only used the `CacheFrom` attribute.

### Steps to reproduce

```
package main

import (
	"github.com/pulumi/pulumi-awsx/sdk/go/awsx/ecr"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		_, err := ecr.NewImage(ctx, "test", &ecr.ImageArgs{
[...]

PR #1278, opened 2024-04-26
Use new docker build provider

blampe:
This replaces the old Docker provider with the new Docker Build provider.

User-facing AWSX types are unchanged.

Behavior is unchanged, except we will now automatically store and use an inline cache if the user hasn't configured anything explicitly.

Fixes https://github.com/pulumi/pulumi-awsx/issues/1349
Fixes https://github.com/pulumi/pulumi-awsx/issues/1072


awsx/package.json
```diff
@@ -22,8 +22,9 @@
   },
   "//": "Pulumi sub-provider dependencies must be pinned at an exact version because we extract this value to generate the correct dependency in the schema",
   "dependencies": {
-    "@pulumi/aws": "6.32.0",
+    "@pulumi/aws": "6.9.0",
     "@pulumi/docker": "4.5.1",
```
lukehoban, line 26: Can we remove this dependency?

lukehoban (commented):
> SDK types are unchanged -- those remain the old Docker API.

The `@pulumi/docker` typess don't show up in the public API of this component, right?  So what do you mean that "SDK types are unchanged"?  I assume we can remove the dependency on `@pulumi/docker` from both the implementation, and update the implied depdendency in the schema (which I assume is only to hint at the internal dependnecy - even though not strictly needed?).  Feel like maybe I'm missing something that motivated this comment though?

awsx/package.json
blampe: I think removing this will break anyone who indirectly depends on it -- is that OK? In other words I can specify only awsx in package.json but still use docker in my program.

blampe:
> So what do you mean that "SDK types are unchanged"?

Sorry, this meant that awsx's own wrapper types around Docker are unchanged. Only the internals are changing, the user-facing API remains the same.

awsx/ecr/image.ts
```diff
@@ -62,37 +62,52 @@ export function computeImageFromAsset(
...
+  if (cacheFrom.length === 0) {
+    cacheFrom.push({ registry: { ref: canonicalImageName } });
+  }
+
+  let context = ".";
+  if (dockerInputs.context !== undefined) {
+    context = dockerInputs.context;
+  }
+
   const dockerImageArgs: docker.ImageArgs = {
```
flostadler, line 91: Are there any new input properties we should expose as well?

flostadler (commented):
LGTM, just two small questions.
I was thinking that we should add an upgrade test for going from the old to the new docker provider, what do you think?

awsx/ecr/image.ts
```diff
@@ -62,37 +62,52 @@ export function computeImageFromAsset(
...
-      context: dockerInputs.context,
-      dockerfile: dockerInputs.dockerfile,
-      platform: dockerInputs.platform,
-      target: dockerInputs.target,
-    },
-    registry: registryCredentials,
+    tags: [canonicalImageName],
+    buildArgs: dockerInputs.args,
+    cacheFrom: cacheFrom,
+    cacheTo: [{ inline: {} }],
```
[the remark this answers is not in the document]
blampe: Great question! It's actually full cache artifacts, so it can increase the size a bit. I think it would be best to take this out and stay as close to the existing behavior of not doing any caching stuff unless `cacheFrom` is set.

However this makes me realize we probably _do_ want inline caching when `cacheFrom` is set. Currently it's using a registry cache output, but this isn't compatible with the existing behavior because users currently point this at an image (not a cache artifact) due to the way the old Docker provider did caching. (It would simply pull images to get layers into your local build cache.) In other words we don't want to export a cache to `image:latest`, but it would be reasonable to push an image with inline caching.

Caching Docker images is hard enough as it is and the old provider's behavior was also easy to get wrong. For example `cacheFrom` might not do anything without a magic env var:
```ts
new docker.Image("my-app-image", {
    build: {
        args: {
            BUILDKIT_INLINE_CACHE: "1",
        },
        cacheFrom: {
            images: ["foo:latest"],
        },
        context: "app/",
        dockerfile: "app/Dockerfile",
    },
    imageName: "foo:latest",
});
```
So for us to stay as close to the existing behavior while also allowing caching to _somewhat_ work, I think we basically just want to enable this `BUILDKIT_INLINE_CACHE` behavior when `cacheFrom` is present.

awsx/ecr/image.ts
blampe: Probably! There's a lot more we can do now, e.g. multiple tags, multiple platforms, more refined caching behavior, etc.

I didn't want to add any of those right now because (a) these are opinionated APIs and opinions take more time to get concensus on, and (b) it's not clear how much (if any) we want to continue investing in this wrapper versus giving users more composable building blocks (like your example of registry auth helpers). The latter is my preference, but it's not my call to make.

t0yv0:
Are we going forward with this @blampe or should we pick up on AWS squad?

What is your confidence that the upgrade is going to be "invisible" to users? Does new Docker Build provider Image have an alias to the old image and is ready to accept old image states and rewire them? Or is there going to be a replacement during upgrade but we believe it is benign?

> I was thinking that we should add an upgrade test for going from the old to the new docker provider, what do you think?

This would be good but I also wonder if we know enough to cover the upgrade space well enough to test it properly.

Perhaps instead we can just include this change in the next major version?

t0yv0:
This likely will also fix https://github.com/pulumi/pulumi-awsx/issues/1401

flostadler:
@t0yv0 In my opinion we shouldn't wait for the next major with this. Even if the image does get rebuilt it shouldn't be a problem. It's not a breaking change like infrastructure being replaced.

As part of another change I'm going to add a function to the provider for retrieving ECR creds. If there's users that do not want to go with docker-build, they could use that and pipe it into the `docker.Image` resource.

awsx/ecr/image.ts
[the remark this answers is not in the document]
flostadler: I tried a couple of different sample images and there was no substantial increase in image size. For some the image was even smaller. So the current approach sounds good to me

flostadler:
I resolved the merge conflicts with the main branch, updated to the latest docker-build (0.0.8) and added an upgrade test. I think this is good to go now!

corymhall (approved):

flostadler:
@blampe I updated the PR, could you have a look as well?

awsx/package.json
```diff
@@ -25,7 +25,7 @@
   "//": "Pulumi sub-provider dependencies must be pinned at an exact version because we extract this value to generate the correct dependency in the schema",
   "dependencies": {
     "@pulumi/aws": "6.65.0",
-    "@pulumi/docker": "4.5.1",
+    "@pulumi/docker-build": "0.0.8",
```
blampe, line 28: Is this possible a breaking change? Like if I pull in `p/docker` indirectly via `awsx` and then reference it from my code, my imports would break?

I don't know if that's what would actually happen or if it's avoidable. Would you leave in the old p/docker dependency even if it's unused?

Mostly just asking to make sure we've thought through any edge cases.

blampe:
> @blampe I updated the PR, could you have a look as well?

@flostadler it looks great to me! Just one small question re: dependencies but not blocking. Feel free to merge whenever you want!

awsx/package.json
flostadler: IMO it's not a breaking change. Users shouldn't depend on transitive dependencies directly. They should add those to their own dependencies.

That being said, I'm currently workin on another awsx feature that'll add back the docker provider for push without build scenarios so we're gonna be covered here

Issue #1282, opened 2024-05-08
add security group option for network load balancer


w0aw:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

i have an k3s cluster running on private subnet. with 2 k3s master nodes behind an NLB and now if i try to join a new worker node using NLB dns then it fails. like the dns resolves to some private ips which are not reachable and curretly i am using a single master node's private ip join new worker nodes. if somehow i can add a security group (that allow all ingress traffic from same vpc) then this issue can be solved. 

### Affected area/feature

thank you. i hope i am on right place

flostadler:
Hey @w0aw, thanks for bringing this up! I assume you're referring to the `NetworkLoadBalancer` component resource, right?

You're correct that it (and the ALB resource) is lacking a way to configure security groups for the NLB right now. I'm gonna work on prioritizing this.

As a workaround you could use transformations (https://www.pulumi.com/docs/concepts/options/transformations/) to inject the `securityGroups` parameter on the underlying `aws:lb/loadBalancer:LoadBalancer` resource while we work on adding exposing the parameter here.

w0aw:
hi @flostadler, thanks for resolving this issue. can you please confirm how long it will take to get merged and npm release?

Issue #1114, opened 2023-10-23
TestExamples/ts-nlb-simple/upgrade-preview-only fails on CI


t0yv0:
### What happened?

There is an issue with a replay-based provider upgrade test. It seems to pass for me locally but fails in CI.  The failure reads something like this:

```
warning: using pulumi-resource-awsx from $PATH at /home/runner/work/pulumi-awsx/pulumi-awsx/bin/pulumi-resource-awsx
  warning: using pulumi-resource-awsx from $PATH at /home/runner/work/pulumi-awsx/pulumi-awsx/bin/pulumi-resource-awsx
    pulumi:pulumi:Stack: (same)
      [urn=urn:pulumi:p-it-fv-az491-5-ts-nlb-sim-b0cccd6e::ts-nlb-simple::pulumi:pulumi:Stack::ts-nlb-simple-p-it-fv-az491-5-ts-nlb-sim-b0cccd6e]
  warning: using pulumi-resource-awsx from $PATH at /home/runner/work/pulumi-awsx/pulumi-awsx/bin/pulumi-resource-awsx
          +-aws:lb/loadBalancer:LoadBalancer: (replace)
              [id=arn:aws:elasticloadbalancing:us-west-2:616138583583:loadbalancer/net/nginx-lb-c00afb4/821fc191ad125fbe]
              [urn=urn:pulumi:p-it-fv-az491-5-ts-nlb-sim-b0cccd6e::ts-nlb-simple::awsx:lb:NetworkLoadBalancer$aws:lb/loadBalancer:LoadBalancer::nginx-lb]
              [provider: urn:pulumi:p-it-fv-az491-5-ts-nlb-sim-b0cccd6e::ts-nlb-simple::pulumi:providers:aws::default_5_35_0::fcb88445-5634-43ad-9dd9-d9d032832b45 => urn:pulumi:p-it-fv-az491-5-ts-nlb-sim-b0cccd6e::ts-nlb-simple::pulumi:providers:aws::default_6_5_0::output<string>]
            ~ subnets: [
                ~ [0]: "subnet-0016572b" => "subnet-03711d3b9b21b3a8e"
                ~ [1]: "subnet-d7e7fe9c" => "subnet-06e8296c053e2b952"
                ~ [2]: "subnet-c7d926bf" => "subnet-0fc2dc8f8ba906919"
                ~ [3]: "subnet-43f43a1e" => "subnet-037f366816336db85"
              ]
[...]

PR #1285, opened 2024-05-08
Add Security Group to NLBs

flostadler:
Since August 23 NLBs support Security Groups as well now (see [AWS blog post](https://aws.amazon.com/blogs/containers/network-load-balancers-now-support-security-groups/)).

This change adds the parameters for configuring security groups to the NLB component as well. The one notable difference to the ALB security group is that I refrained from opening up the whole security group by default.

Components should be secure by default, but right now a fully open security group (ingress&egress) is created when users create an ALB without specifying a security group. What makes this issue worse is that the load balancer is placed into a public subnet by default, ergo automatically exposing what's behind the LB to the whole world.

Fixes https://github.com/pulumi/pulumi-awsx/issues/1282

### DIFF
--- a/awsx/lb/applicationLoadBalancer.ts
+++ b/awsx/lb/applicationLoadBalancer.ts
@@ -117,50 +117,6 @@ export class ApplicationLoadBalancer extends schema.ApplicationLoadBalancer {
       }
     }
 
-    if (listener && listeners) {
-      throw new Error("Only one of [listener] and [listeners] can be specified");
-    }
-
-    if (!lbArgs.securityGroups && !defaultSecurityGroup?.skip) {
-      if (defaultSecurityGroup?.args && defaultSecurityGroup.securityGroupId) {
-        throw new Error(
-          "Only one of [defaultSecurityGroup] [args] or [securityGroupId] can be specified",
-        );
-      }
-      const securityGroupId = defaultSecurityGroup?.securityGroupId;
-      if (securityGroupId) {
-        lbArgs.securityGroups = [securityGroupId];
-      } else {
-        const securityGroup = new aws.ec2.SecurityGroup(
-          name,
-          defaultSecurityGroup?.args ?? {
-            // TO-DO: we default to open SG rules at this point - we should review this!
-            ingress: [
-              {
-                fromPort: 0,
-                toPort: 0,
-                protocol: "-1",
-                cidrBlocks: ["0.0.0.0/0"],
-                ipv6CidrBlocks: ["::/0"],
-              },
-            ],
-            egress: [
-              {
-                fromPort: 0,
-                toPort: 65535,
-                protocol: "tcp",
-                cidrBlocks: ["0.0.0.0/0"],
-                ipv6CidrBlocks: ["::/0"],
-              },
-            ],
-          },
-          { parent: this },
-        );
-        this.defaultSecurityGroup = securityGroup;
-        lbArgs.securityGroups = [securityGroup.id];
-      }
-    }
-
     // this is an application loadbalancer so we set this explicitly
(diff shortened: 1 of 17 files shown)


awsx/lb/applicationLoadBalancer.ts
```diff
@@ -117,50 +117,6 @@ export class ApplicationLoadBalancer extends schema.ApplicationLoadBalancer {
       }
     }
 
-    if (listener && listeners) {
```
flostadler, line 120: This whole section was duplicated in the function. It's defined here first: https://github.com/pulumi/pulumi-awsx/blob/adae402632bd5a2b5907745ad1c3b699f2165755/awsx/lb/applicationLoadBalancer.ts#L75-L118

Luckily it didn't have a bad effect because `lbArgs.securityGroups` is set within the function.

flostadler:
I just noticed that this will be trickier than I thought.

If an NLB has no security groups, then adding one causes a replacement.
I think this change might lead to replacements when doing an upgrade because we now default to adding a security group.

I will check that, but we might have to skip the default security group for NLBs alltogether

flostadler:
> I just noticed that this will be trickier than I thought.
> 
> If an NLB has no security groups, then adding one causes a replacement. I think this change might lead to replacements when doing an upgrade because we now default to adding a security group.
> 
> I will check that, but we might have to skip the default security group for NLBs alltogether

Yeah that's what I ended up doing. Default security groups do not play nicely with NLBs because that would cause replacements during upgrades.

corymhall (approved):
@flostadler looks good!

> The one notable difference compared to the ALB component is that I refrained from adding the default security group. For NLBs security groups cannot be added if none are currently present, and cannot all be removed once added. Adding a default security group to NLBs would cause replacements during upgrades.

cc @t0yv0 this is a good example of how it is difficult to keep evolving abstractions. In a perfect world we would be able to create the security group by default _only_ for new load balancers.

awsx/schema-types.ts
```diff
@@ -346,6 +346,7 @@ export interface NetworkLoadBalancerArgs {
     readonly name?: pulumi.Input<string>;
     readonly namePrefix?: pulumi.Input<string>;
     readonly preserveHostHeader?: pulumi.Input<boolean>;
+    readonly securityGroups?: pulumi.Input<pulumi.Input<string>[]>;
```
t0yv0, line 349: Interesting that this is added but consuming code didn't have to have any changes. Feels a bit magical.

t0yv0 (approved):

t0yv0:
> In a perfect world we would be able to create the security group by default only for new load balancers.

Hm that's true, this level of control is possible in a Custom Resource that can manage the lifecycle with full access to the details, but it is harder in the Component Resource level.

awsx/schema-types.ts
flostadler: I did check the different language SDKs to be sure, but all of them seem to be able to correctly handle adding optional types without breaking changes.

examples/ts-lb-simple/index.ts
```diff
@@ -2,14 +2,51 @@ import * as pulumi from "@pulumi/pulumi";
 import * as aws from "@pulumi/aws";
 import * as awsx from "@pulumi/awsx";
 
+// Create VPC for the LB and ECS service
```
flostadler, line 5: The changes to the examples were required because they heavily used invokes  (e.g. to get the default VPC) which are not compatible with upgrade tests because they do call out to the cloud. This caused the tests to fail in CI because it was running in a different AWS account.

This is what caused https://github.com/pulumi/pulumi-awsx/issues/1114

corymhall (approved):

Issue #1191, opened 2024-01-03
Would like to use a different version of the pulumi-aws package then the one pinned by pulumi-awsx


lakkeger:
### What happened?

Current implementation ignores installed package version and falls back to some hardcoded pulumi-aws version.

### Example

Given the below sample project:
```python
import pulumi
import pulumi_aws as aws
import pulumi_awsx as awsx

lb = awsx.lb.ApplicationLoadBalancer("lb")
cluster = aws.ecs.Cluster("cluster")

service = awsx.ecs.FargateService("service",
    cluster=cluster.arn,
    assign_public_ip=True,
    desired_count=2,
    task_definition_args=awsx.ecs.FargateServiceTaskDefinitionArgs(
        container=awsx.ecs.TaskDefinitionContainerDefinitionArgs(
            name="my-service",
            image="nginx:latest",
            cpu=128,
            memory=512,
            essential=True,
            port_mappings=[awsx.ecs.TaskDefinitionPortMappingArgs(
                container_port=80,
                target_group=lb.default_target_group,
            )],
        ),
    ))

pulumi.export("url", pulumi.Output.concat("http://", lb.load_balancer.dns_name))
```
I ran preview:
```
$ pulumilocal preview
Updating this Stack with LocalStack config
Previewing update (dev):
     Type                                Name            Plan       Info
 +   pulumi:pulumi:Stack                 ecs-sample-dev  create     1 error
 +   ├─ awsx:lb:ApplicationLoadBalancer  lb              create
 +   ├─ aws:ecs:Cluster                  cluster         create
     └─ pulumi:providers:aws             default_6_9_0              1 error

Diagnostics:
  pulumi:providers:aws (default_6_9_0):
    error: pulumi:providers:aws resource 'default_6_9_0' has a problem: could not validate provider configuration: Invalid or unknown key. Check `pulumi config get aws:endpoints`.

  pulumi:pulumi:Stack (ecs-sample-dev):
    error: Resource monitor has terminated, shutting down
```

### Output of `pulumi about`

```
CLI
Version      3.99.0
Go Version   go1.21.5
Go Compiler  gc

Plugins
NAME    VERSION
aws     6.17.0
awsx    2.4.0
docker  4.5.0
local   1.2.1
python  unknown

Host
OS       darwin
Version  14.2.1
Arch     arm64

This project is written in python: executable='/Users/lakkeger/Workspace/pulumi/ecs-sample/venv/bin/python3' version='3.11.5'

[...]

PR #1294, opened 2024-05-30
Upgrade AWS to v6.37.1

t0yv0:
Dependency updates:

```
@pulumi/aws: 6.32.0 to 6.37.1
@pulumi/sdk: v3.112.0 to v3.116.1
```

In particular AWS v6.36.0 has a change that makes it work better with LocalStack as it no longer fails opaquely when given an workspacesweb endpoint override in `aws:endpoints`.

This does not solve https://github.com/pulumi/pulumi-awsx/issues/1191 yet but alleviates most of the underlying trouble.


scripts/upgrade-aws.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
t0yv0, line 1: This seems to be all that's needed? Or any missed steps here?

mjeffryes: I think that's correct. (Or at least that's all I had to do to update the version of the docker plugin a while back.)

mjeffryes (approved):

PR #1295, opened 2024-05-30
Document AWS version resolution

t0yv0:
Documents how AWS provider version is selected and gives an example of overriding it.

Fixes https://github.com/pulumi/pulumi-awsx/issues/1191


mjeffryes (approved):
Awesome! Thanks for digging this up and documenting it!

t0yv0:
https://github.com/pulumi/ci-mgmt/issues/950 looks like I need to regenerate the SDKs.

PR #1297, opened 2024-06-03
Embed SDK version

danielrbradley:
Part of https://github.com/pulumi/ci-mgmt/issues/915

Bring workflows in line with the approach used in other providers.

1. Use action for checking for unexpected changes - and allow files with the version embedded to change once we've enabled `respectSchemaVersion`.
2. Fix how we calculate the version we're building. Stop relying on the local git checkout to contain all tags. Use a placeholder locally and override the placeholder in CI with the version calculated from our action.
    - Remove fetching all tags.
3. (Pending) enable `respectSchemaVersion`.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -32,11 +32,6 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
-      - name: Checkout Scripts Repo
-        uses: actions/checkout@v4
-        with:
-          path: ci-scripts
-          repository: pulumi/scripts
       - name: Unshallow clone for tags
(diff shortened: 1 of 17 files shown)


flostadler (approved):
Nice! 

I think this one here also needs those changes: https://github.com/pulumi/pulumi-awsx/blob/master/.github/workflows/weekly-pulumi-update.yml
It was added recently, if you want I can port your changes to that one

danielrbradley:
> I think this one here also needs those changes: https://github.com/pulumi/pulumi-awsx/blob/master/.github/workflows/weekly-pulumi-update.yml It was added recently, if you want I can port your changes to that one

Cool - will rebase and update that too 👍

PR #1298, opened 2024-06-04
Automate weekly pulumi/pulumi upgrade

### LINKED ISSUE #1264: Automate keeping pulumi/pkg reference updated on a weekly cadence
## Hello!

## Issue details

Avoid P1s like https://github.com/pulumi/pulumi-awsx/issues/1262 

Keep the latest fixes flowing

flostadler:
This sets up a cron job to bump pulumi/pulumi versions every week. The cron job updates the dependencies, rebuilds the provider and then opens a PR. The PR is set to auto merge.

Closes #1264

### DIFF
--- a/.github/workflows/weekly-pulumi-update.yml
+++ b/.github/workflows/weekly-pulumi-update.yml
@@ -3,8 +3,10 @@ name: weekly-pulumi-update
 #   schedule:
 #   - cron: 35 12 * * 4
 #   workflow_dispatch: {}
-"on":
-  pull_request_target:
+on:
+  pull_request:
+    branches:
+      - master
 
 permissions:
   pull-requests: write
@@ -60,5 +62,6 @@ jobs:
       run: |
         gh_pr_up() { gh pr create $* || gh pr edit $* }
         gh_pr_up --title "Update pulumi/pulumi version to $PULUMI_VERSION"
-        gh pr merge --auto
+        Disable auto-merge for testing
+        # gh pr merge --auto
     name: weekly-pulumi-update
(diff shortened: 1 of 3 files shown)


.github/workflows/weekly-pulumi-update.yml
```diff
@@ -0,0 +1,89 @@
+name: weekly-pulumi-update
+on:
+  schedule:
+  - cron: 0 11 * * 1
```
flostadler, line 4: Every monday at 11AM UTC

iwahbe: If you need to explain it, why not put it in the code.
```suggestion
  # Every Monday at 11AM UTC
  - cron: 0 11 * * 1
```

scripts/bump-pulumi-deps.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
iwahbe, line 1: I'm having trouble understanding why this script isn't just
```sh
go get -u github.com/pulumi/pulumi/pkg/v3
go get -u github.com/pulumi/pulumi/sdk/v3
go mod tidy
```

iwahbe (commented):
This looks like a good idea.

t0yv0 (approved):
Err on the side of merging, can fixup or edit as needed I think. Thank you! One day perhaps a common way to do this in ci-mgmt but it doesn't have to be now.

iwahbe:
> Err on the side of merging, can fixup or edit as needed I think. Thank you! One day perhaps a common way to do this in ci-mgmt but it doesn't have to be now.

CC @danielrbradley who is working on unifying theses repos in ci-mgmt.

Issue #1322, opened 2024-06-21
Unable to use pulumi-awsx v2.12.x in Golang


rquitales:
### What happened?

Attempt to use `v2.12.0` or `v2.12.1` of `pulumi-awsx` in a Golang program and observe that it fails.

Error message:
```shell
Previewing update (dev)

     Type                      Name                       Plan       Info
 +   pulumi:pulumi:Stack       awsx-go-dev                create     
     └─ pulumi:providers:awsx  default_2_0_0_alpha_0_dev             1 error

Diagnostics:
  pulumi:providers:awsx (default_2_0_0_alpha_0_dev):
    error: Could not automatically download and install resource plugin 'pulumi-resource-awsx' at version v2.0.0-alpha.0+dev, install the plugin using `pulumi plugin install resource awsx v2.0.0-alpha.0+dev`: error downloading provider awsx to file: failed to download plugin: awsx-2.0.0-alpha.0+dev: 403 HTTP error fetching plugin from https://get.pulumi.com/releases/plugins/pulumi-resource-awsx-v2.0.0-alpha.0%2Bdev-darwin-arm64.tar.gz
```

It appears that we are not correctly embedding the version for our Golang SDK (https://github.com/pulumi/pulumi-awsx/blob/d26a60c3c313fa20da1fd8ed0d23e6ce48ce9cb0/sdk/go/awsx/internal/pulumiUtilities.go#L168). This line should be `v2.12.1`.

It appears that we are missing the download of the golang SDK in our publish/tagging step of the release workflow.

### Example

Pulumi Go program
```go
package main

import (
	"github.com/pulumi/pulumi-aws/sdk/v6/go/aws/s3"
	"github.com/pulumi/pulumi-awsx/sdk/v2/go/awsx/ec2"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		// Fetch the default VPC for the current AWS region.
		vpc, err := ec2.NewDefaultVpc(ctx, "default-vpc", nil)
		if err != nil {
			return err
		}
		ctx.Export("vpcId", vpc.VpcId)
		ctx.Export("privateSubnetIds", vpc.PrivateSubnetIds)
		ctx.Export("publicSubnetIds", vpc.PublicSubnetIds)

		// Create an AWS resource (S3 Bucket)
		bucket, err := s3.NewBucket(ctx, "my-bucket", nil)
[...]

PR #1323, opened 2024-06-21
ci: fix go sdk publish to embed version

rquitales:
Follows the workflow defined in ci-mgmt (https://github.com/pulumi/ci-mgmt/blob/7c1f0521b8f04aff46732c6656659c85d60fd9c4/provider-ci/test-workflows/aws/.github/workflows/release.yml#L149-L186) as the basis for this change. This ensures that during the publishing step, we download the Go SDK artifact which contains the correctly embedded version. We then use the `pulumi/publish-go-sdk-action` action to commit/push the new files and cut a new sdk tag for this.

Fixes: #1322

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -344,9 +344,28 @@ jobs:
         uses: jaxxstorm/action-install-gh-release@v1.10.0
         with:
           repo: pulumi/pulumictl
-      - name: Add SDK version tag
-        run: git tag "sdk/v$PROVIDER_VERSION" && git push origin
-          "sdk/v$PROVIDER_VERSION"
+      - name: Download Go SDK
+        uses: actions/download-artifact@v4
+        with:
+          name: go-sdk.tar.gz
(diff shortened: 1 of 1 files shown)


blampe (commented):
Do we need a similar change in pre-release.yml?

rquitales:
> Do we need a similar change in pre-release.yml?

Good question, I opted not to update prerelease since it appears that this action is mostly unused? https://github.com/pulumi/pulumi-awsx/actions/workflows/pre-release.yml Last run was 8 months ago.

blampe (approved):
> > Do we need a similar change in pre-release.yml?
> 
> Good question, I opted not to update prerelease since it appears that this action is mostly unused? https://github.com/pulumi/pulumi-awsx/actions/workflows/pre-release.yml Last run was 8 months ago.

I'd suggest updating it for good measure - just so the next person to run it X months from now doesn't get surprised by this as well. 

But no need to block on that -- better to fix the published version ASAP.

PR #1325, opened 2024-06-22
Use v3 of actions/download-artifact

rquitales:
https://github.com/pulumi/pulumi-awsx/pull/1323 used v4 of the download artifact action, which is incompatible with uploads using v3 of the upload action. See [FAQ](https://github.com/actions/toolkit/blob/main/packages/artifact/docs/faq.md#which-versions-of-the-artifacts-packages-are-compatible) for compatibility matrix.

This resulted in the release step [failing](https://github.com/pulumi/pulumi-awsx/actions/runs/9620941727/job/26540766895#step:4:36) to download the uploaded artifact.

This PR also updates the pre-release workflow while we're at it.

Note: I've opted to use v3 of the download action to maintain consistency with the rest of our GHA workflows. We can update to v4 in the future.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -346,9 +346,28 @@ jobs:
         uses: jaxxstorm/action-install-gh-release@v1.10.0
         with:
           repo: pulumi/pulumictl
-      - name: Add SDK version tag
-        run: git tag "sdk/v$PROVIDER_VERSION" && git push origin
-          "sdk/v$PROVIDER_VERSION"
+      - name: Download Go SDK
+        uses: actions/download-artifact@v4
+        with:
+          name: go-sdk.tar.gz
+          path: $${{ github.workspace }}/sdk/
+      - name: Uncompress Go SDK
+        run: tar -zxf ${{ github.workspace }}/sdk/go.tar.gz -C
+          ${{ github.workspace }}/sdk/go
+        shell: bash
+      - uses: pulumi/publish-go-sdk-action@v1
+        with:
+          repository: ${{ github.repository }}
+          base-ref: ${{ github.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -345,10 +345,10 @@ jobs:
         with:
           repo: pulumi/pulumictl
       - name: Download Go SDK
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v3
         with:
           name: go-sdk.tar.gz
-          path: $${{ github.workspace }}/sdk/
+          path: ${{ github.workspace }}/sdk/
```
rquitales, line 351: Extra "$" should also not be included.

.github/workflows/pre-release.yml
```diff
@@ -346,9 +346,28 @@ jobs:
         uses: jaxxstorm/action-install-gh-release@v1.10.0
         with:
           repo: pulumi/pulumictl
-      - name: Add SDK version tag
-        run: git tag "sdk/v$PROVIDER_VERSION" && git push origin
-          "sdk/v$PROVIDER_VERSION"
+      - name: Download Go SDK
+        uses: actions/download-artifact@v3
```
iwahbe, line 350: Most of our actions outside of this repo (including ci-mgmt) use v4. Could we fail forward into using v4 everywhere?

rquitales: I agree that upgrading to v4 should be the long term goal. However, our Go SDK is unusable currently which is why I opted to use v3 to avoid any other unexpected failures during the release workflow and unbork the Go SDK ASAP.

justinvp (approved):
(Drive by review) looks reasonable to me to fix the Go SDK

Issue #1250, opened 2024-03-29
`awsx.ecs.FargateService` resources don't update in parallel, even on separate clusters


markalexander:
### What happened?

I have a number of ECS clusters with Fargate services and tasks.  I can't see any dependencies between them, but these never update in parallel, rather each waiting in turn for the last to update.  This happens both for services on the same cluster and for services on different clusters.

### Example

Here's a minimal example with:

  - Two clusters: Cluster A and Cluster B.
  - Cluster A has two services: Service 1 and Service 2.
  - Cluster B has one Service: Service 1.

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";

const image = "docker.io/memcached:1.6.26"

const clusterA = new aws.ecs.Cluster("cluster-a", {});

new awsx.ecs.FargateService("cluster-a-service-1", {
  cluster: clusterA.arn,
  assignPublicIp: true,
  desiredCount: 1,
  taskDefinitionArgs: {
    container: {
      name: "cluster-a-service-1",
      image,
      cpu: 128,
      memory: 512,
      essential: true,
    },
  },
  forceNewDeployment: true,
});

new awsx.ecs.FargateService("cluster-a-service-2", {
  cluster: clusterA.arn,
  assignPublicIp: true,
  desiredCount: 1,
  taskDefinitionArgs: {
    container: {
      name: "cluster-a-service-2",
      image,
      cpu: 128,
      memory: 512,
      essential: true,
    },
  },
  forceNewDeployment: true,
});


const clusterB = new aws.ecs.Cluster("cluster-b", {});

new awsx.ecs.FargateService("cluster-b-service-1", {
  cluster: clusterB.arn,
  assignPublicIp: true,
  desiredCount: 1,
  taskDefinitionArgs: {
    container: {
      name: "cluster-b-service-1",
      image,
      cpu: 128,
      memory: 512,
      essential: true,
    },
  },
  forceNewDeployment: true,
});
```

**Note that `forceNewDeployment: true` doesn't seem to work (at least for me per #1249), so you may have to e.g. change the tag in `image` to make it update the services on subsequent runs.**

With this, run `pulumi up`.  This results in each service being updated in series, e.g waiting here:

```

     Type                                  Name                 Status               Info
     pulumi:pulumi:Stack                   pulumi_ecs_test-dev  running..            
[...]

PR #1334, opened 2024-06-27
Create example for testing parallel ECS updates

flostadler:
Adding an example to test parallel updates to components like ECS services. Currently component updates are serialized (#1250), but https://github.com/pulumi/pulumi/issues/7629 fixed that.

I used this test for verifying the changes locally, now that the change is released we can use this as a regression test


examples/examples_nodejs_test.go
```diff
@@ -201,6 +202,24 @@ func TestVpcMultipleSimilarSubnetSpecArgs(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccEcsParallel(t *testing.T) {
+	maxDuration(15*time.Minute, t, func(t *testing.T) {
```
flostadler, line 206: Deploying and updating 10 of these ECS services took 1+ hours before. Now that it's all parallel it runs in under 10min

corymhall (approved):
Looks good! My only question is whether we need this test in `pulumi-awsx`. I feel like a regression test belongs in `pu/pu` since that is where the breaking change would happen and where the fix would be needed.

flostadler:
> Looks good! My only question is whether we need this test in `pulumi-awsx`. I feel like a regression test belongs in `pu/pu` since that is where the breaking change would happen and where the fix would be needed.

Generally I agree with you, but this aspect of pu/pu is under-tested right now. I tried adding some tests for this area (provider server) there, but this would've required to refactor the provider server, which would be quite the rabbit hole.

PR #1352, opened 2024-07-30
Permit aws.ec2.Vpc to use IPAM-allocated cidrBlock ranges

### LINKED ISSUE #872: ec2.Vpc does not work with IPAM
### What happened?

`VpcArgs` has `ipv4IpamPoolId` and `ipv4NetmaskLength` fields, but it doesn't look like the code is set up to handle them.

https://github.com/pulumi/pulumi-awsx/blob/dfe6be74d8142c56ea7f00ff1618de7bda23bcb0/awsx/ec2/vpc.ts#L80-L83

This code assumes that a static `cidrBlock` will be used and applies a default if one is not provided.

### Steps to reproduce

Create a `Vpc` with `ipv4IpamPoolId` set.

### Expected Behavior

The stack deploys.

### Actual Behavior

```
  aws:ec2:Vpc (vpc-primary):
    error: aws:ec2/vpc:Vpc resource 'vpc-primary' has a problem: Conflicting configuration arguments: "ipv4_netmask_length": conflicts with cidr_block. Examine values at 'Vpc.Ipv4NetmaskLength'.
    error: aws:ec2/vpc:Vpc resource 'vpc-primary' has a problem: Conflicting configuration arguments: "cidr_block": conflicts with ipv4_netmask_length. Examine values at 'Vpc.CidrBlock'.
```

### Versions used

```
CLI          
Version      3.34.1
Go Version   go1.17.11
Go Compiler  gc

Plugins
NAME    VERSION
aws     5.8.0
awsx    1.0.0-beta.8
docker  3.2.0
nodejs  unknown
random  4.8.0

Host     
OS       ubuntu
Version  20.04
Arch     x86_64

This project is written in nodejs: executable='/home/codespace/.nodejs/current/bin/node' version='v16.14.2'

Backend        
Name           codespaces-3c53a8
URL            s3://[REDACTED]
User           codespace
Organizations  

Dependencies:
NAME            VERSION
@pulumi/aws     5.8.0
@pulumi/awsx    1.0.0-beta.8
@pulumi/pulumi  3.34.1
@pulumi/random  4.8.0
@types/node     14.18.21
```

### Additional context

_No response_

### Contributing

t0yv0:
Fixes issues with supporting the ipv4IpamPoolId parameter that ties a VPC to an IPAM pool. 

You should now be able to write the following to allows IPAM to allocate and manage a cidrBlock range. The VPC component now uses that dynamically allocated block to automatically configure subnets.

```typescript
new awsx.ec2.Vpc("myVpc", {
  ipv4IpamPoolId: myVpcIpamPool.id,
  ipv4NetmaskLength: 24,
  subnetStrategy: "Auto",
});
```

It is also possible to constrain the allocated subnets with subnetSpecs, while still using IPAM to manage the overall cidrBlock range:

```typescript
new awsx.ec2.Vpc("myVpc", {
  numberOfAvailabilityZones: 3,
  subnetStrategy: "Auto",
  ipv4IpamPoolId: myVpcIpamPool.id,
  ipv4NetmaskLength: 22,
  subnetSpecs: [
    {
      type: "Private",
      name: "private",
      cidrMask: 25,
    },
    {
      type: "Public",
      name: "public",
      cidrMask: 27,
    },
  ],
  tags: tags,
});
```

Fixes #872 

Note that `subnetStrategy: "Auto"` is required with this functionality, and "Legacy" strategy is not supported.

### DIFF
--- a/awsx/ec2/subnetDistributorNew.ts
+++ b/awsx/ec2/subnetDistributorNew.ts
@@ -1,4 +1,4 @@
-// Copyright 2016-2022, Pulumi Corporation.
+// Copyright 2016-2024, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,7 +20,6 @@ import { SubnetSpecInputs } from "../schema-types";
 import { Netmask } from "netmask";
 import { SubnetSpec, SubnetSpecPartial } from "./subnetSpecs";
 
-// Like getSubnetSpecs but tolerates partially known vpcCidr.
 export function getSubnetSpecs(
   vpcName: string,
   vpcCidr: pulumi.Input<string>,
@@ -28,21 +27,19 @@ export function getSubnetSpecs(
   subnetInputs: SubnetSpecInputs[] | undefined,
   azCidrMask?: number,
 ): SubnetSpecPartial[] {
-  const allocatedCidrBlocks = allocateSubnetCidrBlocksInput(
-    vpcName,
-    vpcCidr,
-    azNames,
-    subnetInputs,
-    azCidrMask,
-  );
+  const allocatedCidrBlocks = pulumi
+    .output(vpcCidr)
+    .apply((vpcCidr) =>
+      allocateSubnetCidrBlocks(vpcName, vpcCidr, azNames, subnetInputs, azCidrMask),
+    );
   const subnetSpecs = subnetInputs ?? defaultSubnetInputsBare();
   return azNames.flatMap((azName, azIndex) => {
     const azNum = azIndex + 1;
     return subnetSpecs.map((subnetSpec, subnetIndex) => {
       const subnetAllocID = subnetAllocationID(vpcName, subnetSpec, azNum, subnetIndex);
-      const allocated = allocatedCidrBlocks[subnetAllocID];
+      const cidrBlock = allocatedCidrBlocks.apply((t) => pulumi.output(t[subnetAllocID].cidrBlock));
       return {
-        cidrBlock: allocated.cidrBlock,
+        cidrBlock,
         type: subnetSpec.type,
         azName,
         subnetName: subnetName(vpcName, subnetSpec, azNum),
@@ -62,36 +59,6 @@ function subnetAllocationID(
 ): SubnetAllocationID {
   const name = subnetName(vpcName, subnetSpec, azNum);
   return `${name}#${subnetSpecIndex}`;
-}
-
-// Like allocateSubnetCidrBlocks but accepts pulumi.Input for vpcCidr.
-export function allocateSubnetCidrBlocksInput(
-  vpcName: string,
-  vpcCidr: pulumi.Input<string>,
-  azNames: string[],
-  subnetInputs: SubnetSpecInputs[] | undefined,
-  azCidrMask?: number,
-): Record<SubnetAllocationID, { cidrBlock: pulumi.Input<string> }> {
-  if (typeof vpcCidr === "string") {
-    return allocateSubnetCidrBlocks(vpcName, vpcCidr, azNames, subnetInputs, azCidrMask);
-  }
-  const alloc = pulumi
-    .output(vpcCidr)
-    .apply((vpcCidr) =>
-      allocateSubnetCidrBlocks(vpcName, vpcCidr, azNames, subnetInputs, azCidrMask),
-    );
-  const subnetSpecs = subnetInputs ?? defaultSubnetInputsBare();
-  const result: Record<SubnetAllocationID, { cidrBlock: pulumi.Input<string> }> = {};
-  azNames.forEach((_, azIndex) => {
-    const azNum = azIndex + 1;
-    return subnetSpecs.map((subnetSpec, subnetIndex) => {
-      const subnetAllocID = subnetAllocationID(vpcName, subnetSpec, azNum, subnetIndex);
-      result[subnetAllocID] = {
-        cidrBlock: alloc.apply((a) => pulumi.output(a[subnetAllocID].cidrBlock)),
-      };
-    });
-  });
-  return result;
 }
 
 function allocateSubnetCidrBlocks(
--- a/awsx/ec2/subnetDistributorNew.ts
+++ b/awsx/ec2/subnetDistributorNew.ts
@@ -16,7 +16,7 @@
 // and used in accordance with MPL v2.0 license
 
 import * as pulumi from "@pulumi/pulumi";
-import { SubnetSpecInputs, SubnetTypeInputs } from "../schema-types";
+import { SubnetSpecInputs } from "../schema-types";
 import { Netmask } from "netmask";
 import { SubnetSpec, SubnetSpecPartial } from "./subnetSpecs";
 
@@ -27,6 +27,45 @@ export function getSubnetSpecs(
   subnetInputs: SubnetSpecInputs[] | undefined,
   azCidrMask?: number,
 ): SubnetSpecPartial[] {
+  const allocatedCidrBlocks = allocateSubnetCidrBlocks(vpcName, vpcCidr, azNames, subnetInputs, azCidrMask);
+  const subnetSpecs = subnetInputs ?? defaultSubnetInputsBare();
+  return azNames.flatMap((azName, azIndex) => {
+    const azNum = azIndex + 1;
+    return subnetSpecs.map((subnetSpec, subnetIndex) => {
+      const subnetAllocID = subnetAllocationID(vpcName, subnetSpec, azNum, subnetIndex);
+      const allocated = allocatedCidrBlocks[subnetAllocID];
+      return {
+        cidrBlock: allocated.cidrBlock,
+        type: subnetSpec.type,
+        azName,
+        subnetName: subnetName(vpcName, subnetSpec, azNum),
+        tags: subnetSpec.tags,
+      };
+    });
+  });
+}
+
+type SubnetAllocationID = string;
+
+function subnetAllocationID(
+  vpcName: string,
+  subnetSpec: SubnetSpecInputs,
+  azNum: number,
+  subnetSpecIndex: number,
+): SubnetAllocationID {
+  const name = subnetName(vpcName, subnetSpec, azNum);
+  return `${name}#${subnetSpecIndex}`;
+}
+
+function allocateSubnetCidrBlocks(
+  vpcName: string,
+  vpcCidr: string,
+  azNames: string[],
+  subnetInputs: SubnetSpecInputs[] | undefined,
+  azCidrMask?: number
+): Record<SubnetAllocationID, {cidrBlock: pulumi.Input<string>}> {
+
+  const allocation: Record<string, {cidrBlock: pulumi.Input<string>}> = {};
   const vpcNetmask = new Netmask(vpcCidr);
   const azBitmask = azCidrMask ?? vpcNetmask.bitmask + newBits(azNames.length);
 
@@ -51,11 +90,11 @@ export function getSubnetSpecs(
   }
 
   let currentAzNetmask = new Netmask(`${vpcNetmask.base}/${azBitmask}`);
-  const subnets: SubnetSpec[] = [];
+
   for (let azIndex = 0; azIndex < azNames.length; azIndex++) {
-    const azName = azNames[azIndex];
     const azNum = azIndex + 1;
     let currentSubnetNetmask: Netmask | undefined;
+    let subnetIndex = 0;
     for (const subnetSpec of subnetSpecs) {
       if (currentSubnetNetmask === undefined) {
         currentSubnetNetmask = new Netmask(
@@ -69,19 +108,24 @@ export function getSubnetSpecs(
         );
       }
       const subnetCidr = currentSubnetNetmask.toString();
-      subnets.push({
+      allocation[subnetName(vpcName, subnetSpec, azNum)+"#"+subnetIndex] = {
         cidrBlock: subnetCidr,
-        type: subnetSpec.type,
-        azName,
-        subnetName: subnetName(vpcName, subnetSpec, azNum),
-        tags: subnetSpec.tags,
-      });
+      };
+
+      subnetIndex++;
     }
 
     currentAzNetmask = currentAzNetmask.next();
   }
 
-  return subnets;
+  return allocation;
+}
+
+function defaultSubnetInputsBare(): SubnetSpecInputs[] {
+  return [
+    {type: "Private"},
+    {type: "Public"},
+  ];
 }
 
 export function defaultSubnetInputs(azBitmask: number): SubnetSpecInputs[] {
@@ -101,16 +145,10 @@ export function defaultSubnetInputs(azBitmask: number): SubnetSpecInputs[] {
   // Even if we've got more than /16, only use the first /16 for the default subnets.
   // Leave the rest for the user to add later if needed.
   const maxBitmask = Math.max(azBitmask, 16);
-  return [
-    {
-      type: "Private",
-      cidrMask: maxBitmask + 1,
-    },
-    {
-      type: "Public",
-      cidrMask: maxBitmask + 2,
-    },
-  ];
+  return defaultSubnetInputsBare().map((t, i) => ({
+    type: t.type,
+    cidrMask: maxBitmask + i + 1,
+  }));
 }
 
 export function nextNetmask(previous: Netmask, nextBitmask: number): Netmask {
--- a/awsx/ec2/subnetDistributorNew.ts
+++ b/awsx/ec2/subnetDistributorNew.ts
@@ -28,7 +28,13 @@ export function getSubnetSpecsWithPartialCidr(
   subnetInputs: SubnetSpecInputs[] | undefined,
   azCidrMask?: number,
 ): SubnetSpecPartial[] {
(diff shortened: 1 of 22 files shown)


examples/vpc/nodejs/vpc-ipam-ipv4-auto-cidrblock/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const myVpcIpamPool = new aws.ec2.VpcIpamPool("myVpcIpamPool", {
+  addressFamily: "ipv4",
+  ipamScopeId: myVpcIpam.privateDefaultScopeId,
+  locale: currentRegion.name,
+  tags: tags,
+});
+
+const myVpcIpamPoolCidr = new aws.ec2.VpcIpamPoolCidr("myVpcIpamPoolCidr", {
+    ipamPoolId: myVpcIpamPool.id,
```
t0yv0, line 30: Note to self: reindent properly (Emacs confused on indentation level).

awsx/ec2/vpc.ts
```diff
@@ -343,14 +314,100 @@ export class Vpc extends schema.Vpc<VpcData> {
...
+    const parsedSpecs = validateAndNormalizeSubnetInputs(
+      args.subnetSpecs,
+      availabilityZones.length,
+    );
+
+    const thisResource = this;
+
+    const subnetSpecs = (() => {
+
+      // cidrBlock is allocated dynamically but there are no explicit specs; use subnetDistributorNew.
```
t0yv0, line 348: The case analysis here is a bit uncertain for me, appreciate some help formulating which test cases we want to add. 

One particular complication is that `getSubnetSpecsLegacy` cannot yet accept a `cidrBlock: Input<string>` and therefore this does not support IPAM-allocated cidrBlock with Legacy strategy properly. I've only done the Input threading for `getSubnetSpecs` which implements the newer Auto strategy.

awsx/ec2/vpc.ts
```diff
@@ -604,3 +687,9 @@ export function validateNoGaps(vpcCidr: string, subnetSpecs: SubnetSpec[]) {
     `There are gaps in the subnet ranges. Please fix the following gaps: ${gaps.join(", ")}`,
   );
 }
+
+function rewrap(
+  x: pulumi.Output<pulumi.UnwrappedArray<schema.ResolvedSubnetSpecOutputs>>
+): pulumi.Output<schema.ResolvedSubnetSpecOutputs[]> {
+  return x as any;
```
t0yv0, line 694: This is a code-smell and potentially unsafe, need to double-check how we get to need this.

awsx/ec2/vpc.ts
```diff
@@ -157,6 +105,29 @@ export class Vpc extends schema.Vpc<VpcData> {
...
+    const { subnetSpecs: decidedSpecs, subnetLayout } = this.decideSubnetSpecs(
+      name,
+      actualCidrBlock,
+      subnetStrategy,
+      availabilityZones,
+      args,
+    );
+
+    const subnetSpecs = validatePartialSubnetSpecs(decidedSpecs, (subnetSpecs) => {
+      validateSubnets(subnetSpecs, getOverlappingSubnets);
```
t0yv0, line 120: Hmm, self-reviewing this change, there's a problem here that validations now happen after the new aws.ec2.Vpc call, and in the case when they used to terminate the flow promptly, that's a visible change, as now `new aws.ec2.Vpc` attempts to provision before we reject the validations. It's a bit of a chicken and egg problem.

awsx/ec2/vpc.ts
```diff
@@ -343,14 +314,119 @@ export class Vpc extends schema.Vpc<VpcData> {
...
+          const legacySubnetSpecs = getSubnetSpecsLegacy(
+            name,
+            cidrBlock,
+            availabilityZones,
+            parsedSpecs?.normalizedSpecs,
+          );
+          return legacySubnetSpecs;
+        } else {
+          pulumi.log.warn(
+            `Unable to allocate subnets. Try setting subnetStrategy="Auto"`,
```
t0yv0, line 386: Should this fail fast?

awsx/ec2/subnetDistributorNew.ts
```diff
@@ -16,17 +16,52 @@
...
+): SubnetSpecPartial[] {
+  const allocatedCidrBlocks = allocateSubnetCidrBlocksInput(
+    vpcName,
+    vpcCidr,
+    azNames,
+    subnetInputs,
+    azCidrMask,
+  );
+  const subnetSpecs = subnetInputs ?? defaultSubnetInputsBare();
+  return azNames.flatMap((azName, azIndex) => {
```
corymhall, line 39: Can we combine this with the `allocateSubnetCidrBlocksInput`/`allocationSubnetCidrBlocks` function? It looks like we perform the same type of loop a bunch of different places which makes it a little hard to follow what is happening.

awsx/ec2/subnetSpecs.ts
```diff
@@ -0,0 +1,61 @@
...
+    check(xs);
+    return xs;
+  });
+  return specs.map((s, index) => ({ ...s, cidrBlock: checked.apply((cs) => cs[index].cidrBlock) }));
+}
+
+function detectPromptSubnetSpecs(specs: SubnetSpecPartial[]): SubnetSpec[] | undefined {
+  if (specs.every((s) => typeof s.cidrBlock === "string")) {
+    return specs.map((s) => {
+      const cidrBlock: string = typeof s.cidrBlock === "string" ? s.cidrBlock : "";
```
corymhall, line 55: Is this one of those typescript issues where you've already established the type in the outer loop, but it forgets?

awsx/ec2/vpc.ts
corymhall: Would it be acceptable to only support IPAM when using the `Auto` strategy? Users won't be adding IPAM to an existing VPC, so I don't think we should have to support "legacy" stuff.

awsx/ec2/vpc.ts
corymhall: I would vote yes. It will eventually fail and I would rather be told before I deploy than after.

awsx/ec2/vpc.ts
corymhall: I think it's the best that we can do. From the user's perspective I would have to have created the VPC resource in order to have the cidr allocated.

How does pulumi run the validation in this case? Are the creation of the other resources dependent on the validation completing or would they be run in parallel?

awsx/ec2/vpc.ts
```diff
@@ -85,67 +88,12 @@ export class Vpc extends schema.Vpc<VpcData> {
...
-
-    if (subnetStrategy === "Exact") {
-      validateNoGaps(cidrBlock, subnetSpecs);
-    }
-
-    validateNatGatewayStrategy(natGatewayStrategy, subnetSpecs);
 
     const sharedTags = { Name: name, ...args.tags };
 
+    const cidrBlock = Vpc.decideCidrBlockVpcInput(args);
```
corymhall, line 95: Should we also validate here that if the user provides the ipamPoolId then they also provide the netmask?

corymhall (commented):
Had a first pass and mostly things look good!

awsx/ec2/vpc.ts
t0yv0: It runs after the internal vpc resource registers but before subnets do. If there's error reported by the internal vpc resource they now will be reported in addition to validation failures reported by awsx.Vpc. At a cost of a little bit of extra code complexity we can get the previous behavior back while using the new validation ordering behavior for the IPAM case.

awsx/ec2/vpc.ts
t0yv0: This validation is done by the underlying awx.ec2.Vpc resource. It probably is slightly better UX to lift the validation to the awsx.ec2.Vpc layer. We can do this.

awsx/ec2/vpc.ts
t0yv0: Yeah pulumi runs resource allocations in parallel, so when we say new Vpc() we fork off a thread creating that VPC. Validations can run in parallel or they can block resource creation.

awsx/ec2/subnetSpecs.ts
t0yv0: Exactly.

awsx/ec2/subnetDistributorNew.ts
t0yv0: Yeah, I don't know how to make this cleaner. allocateSubnetCidrBlocksInput is very intentionally broken out. Our problem here is that we're trying to keep the allocateSubnetCidrBlocks preserve the original algo as-is. But we need to interface with this algo when`vpcCidr: pulumi.Input<string>` is not known yet. Even in this case where it's not know yet, we can't assume that the entire set of `SubnetSpecPartial[]` is unknown - we know how many there will be, and we know parts of information about them right away at preview, just not their respective cidrBlock values.

This indirection is accomplished by the means of `Record<SubnetAllocationID, { cidrBlock: pulumi.Input<string> }>` helper data structure.

- allocateSubnetCidrBlocks is needed to preserve original algo in as simple code as possible
- allocateSubnetCidrBlocksInput is needed to do tricks to make allocateSubnetCidrBlocks callable with vpcCidr that is now an Input but return a Record, not an `Output<Record>` - tricks are used to not lose information
- getSubnetSpecsWithPartialCidr/getSubnetSpecs is needed to do the rest of SubnetSpecPartial computation, not just the cidrBlock

awsx/ec2/vpc.ts
t0yv0: It now only supports the Auto strategy. Indeed cleaner.

awsx/ec2/vpc.ts
t0yv0: Done.

t0yv0:
The tests pass but they take a very long time, and they start timing out the token used for other tests. I think it's actually the destroy bit that's slow for some reason, likely for the aws Ipam resource. I wonder what's a practical way forward here, is there a way to keep an IPAM resource alive for testing and import it instead of create/delete on each test. Could also trivially combine the tests into one to run against only one IPAM resource.

```
 TestVpcIpam (49m0.33s)
✅ TestVpcIpam/vpc-ipam-ipv4-auto-cidrblock (29m0.69s)
✅ TestVpcIpam/vpc-ipam-ipv4-auto-cidrblock-with-specs (19m59.64s)
```

corymhall:
>I wonder what's a practical way forward here, is there a way to keep an IPAM resource alive for testing and import it instead of create/delete on each test. Could also trivially combine the tests into one to run against only one IPAM resource.


IPAM has a [free tier](https://aws.amazon.com/vpc/pricing/) so I  wonder if it would be better to just create one and leave it there and have any tests that need it import it.

t0yv0:
It turns out the slowness comes from https://github.com/pulumi/pulumi-aws/issues/4346 which makes deleting a VPC wait for IPAM to remove allocations. This aspect will not be helped by keeping an always-created IPAM for testing. Looks like customTimeout setting does not work either.

t0yv0:
Suggesting to go ahead with a workaround to the problem simply admitting longer running tests for the moment: https://github.com/pulumi/pulumi-awsx/pull/1357

awsx/ec2/subnetDistributorNew.ts
corymhall: What if we used union types and type checking? Something like this:

```typescript
export function getSubnetSpecsWithPartialCidr(
  vpcName: string,
  vpcCidr: pulumi.Input<string>,
  azNames: string[],
  subnetInputs: SubnetSpecInputs[] | undefined,
  azCidrMask?: number,
): SubnetSpecPartial[] {
  let alloc:
    | pulumi.Output<Record<SubnetAllocationID, { cidrBlock: pulumi.Input<string> }>>
    | Record<SubnetAllocationID, { cidrBlock: pulumi.Input<string> }>;
  if (typeof vpcCidr === "string") {
    alloc = allocateSubnetCidrBlocks(vpcName, vpcCidr, azNames, subnetInputs, azCidrMask);
  } else {
    alloc = pulumi
      .output(vpcCidr)
      .apply((vpcCidr) =>
        allocateSubnetCidrBlocks(vpcName, vpcCidr, azNames, subnetInputs, azCidrMask),
      );
  }
  const subnetSpecs = subnetInputs ?? defaultSubnetInputsBare();
  return azNames.flatMap((azName, azIndex) => {
    const azNum = azIndex + 1;
    return subnetSpecs.map((subnetSpec, subnetIndex) => {
      const subnetAllocID = subnetAllocationID(vpcName, subnetSpec, azNum, subnetIndex);
      let cidrBlock: pulumi.Input<string>;
      if (pulumi.Output.isInstance(alloc)) {
        const allocOutput = alloc as pulumi.Output<
          Record<SubnetAllocationID, { cidrBlock: pulumi.Input<string> }>
        >;
        cidrBlock = allocOutput.apply((a) => pulumi.output(a[subnetAllocID].cidrBlock));
      } else {
        cidrBlock = alloc[subnetAllocID].cidrBlock as pulumi.Input<string>;
      }
      return {
        cidrBlock: cidrBlock,
        type: subnetSpec.type,
        azName,
        subnetName: subnetName(vpcName, subnetSpec, azNum),
        tags: subnetSpec.tags,
      };
    });
  });
}
```

awsx/ec2/vpc.ts
```diff
@@ -347,14 +297,217 @@ export class Vpc extends schema.Vpc<VpcData> {
...
+  ): {
+    vpc: aws.ec2.Vpc;
+    subnets: {
+      subnetSpecs: SubnetSpecPartial[];
+      subnetLayout: pulumi.Output<schema.ResolvedSubnetSpecOutputs[]>;
+    };
+  } {
+    if (cidrBlock) {
+      // If cidrBlock is known because the user provided it, validations should run as early as possible. Failing
+      // validations will short-circuit trying to create the inner aws.ec2.Vpc resource.
```
corymhall, line 356: Nice!

corymhall (changes requested):
Just a couple of comments, otherwise looks good!

awsx/ec2/subnetDistributorNew.ts
```diff
@@ -35,7 +70,48 @@ export function getSubnetSpecs(
   azNames: string[],
   subnetInputs: SubnetSpecInputs[] | undefined,
   azCidrMask?: number,
-): SubnetSpec[] {
+): SubnetSpecPartial[] {
+  return getSubnetSpecsWithPartialCidr(vpcName, vpcCidr, azNames, subnetInputs, azCidrMask);
```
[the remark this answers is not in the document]
t0yv0: You might be right, I must have made a mistake of thinking about other functions. This can just be getSubnetSpecs.

awsx/ec2/subnetDistributorNew.ts
t0yv0: This is a nice idea, I think you are right. This lets me delete one of the functions. I opted for an even simpler form as I think we do not need to `if (typeof vpcCidr === "string") {` for cidrBlock we can just completely commit to pulumi.Output without loss of pertinent information in this case. Have a look at the latest version and thanks for pushing on this 🙏

t0yv0:
Looks like unit tests do not like it if we lift prompt values to pulumi.Output around cidrBlocks and that may delay validations as well. To be safe I've coded up an inputApply that allows to use apply on pulumi.Input without losing prompt-ness.

corymhall (approved):

awsx/ec2/subnetDistributorNew.ts
```diff
@@ -285,3 +321,18 @@ export const validSubnetSizes: readonly number[] = [
   8388608, 4194304, 2097152, 1048576, 524288, 262144, 131072, 65536, 32768, 16384, 8192, 4096, 2048,
   1024, 512, 256, 128, 64, 32, 16, 8, 4,
 ];
+
+// This utility function is like pulumi.output(x).apply(fn) but tries to stay in the Input layer so that prompt
+// validations and test cases are not disturbed. wrap* functions are usually identity. Ideally something like this could
+// be handled in the core Pulumi Node SDK.
+function inputApply<T, U>(
```
flostadler, line 328: Uff, nice!

awsx/ec2/vpc.ts
flostadler: Awesome!

awsx/ec2/vpcConverters.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as pulumi from "@pulumi/pulumi";
+import * as schema from "../schema-types";
+
+export function toResolvedSubnetSpecOutputs(
+  s: schema.SubnetSpecInputs[],
+): schema.ResolvedSubnetSpecOutputs[] {
+  return s.map(convSubnetSpecInputsToResolvedSubnetSpecOutputs);
+}
+
+function convSubnetSpecInputsToResolvedSubnetSpecOutputs(
```
flostadler, line 28: Yikes, inputs, outputs, unwrap... Good job with taming that complexity!

flostadler (approved):
Awesome!

The only thing I'm wondering is if we could add some unit tests like these here to make the testing easier&faster: https://www.pulumi.com/docs/using-pulumi/testing/unit/. Might be a cool&fast hackathon project to test those

t0yv0:
> The only thing I'm wondering is if we could add some unit tests

Indeed, testing via the integration loop was an absolute bear.  I'll leave this out of this PR but return to it if doing any more serious work around this repo.

awsx/ec2/vpcConverters.ts
t0yv0: I've fixed Haskell singletons code before, this is fun and games.

PR #1355, opened 2024-08-01
Fix VPC support for Auto subnet layout strategy

t0yv0:
The awsx.ec2.Vpc resource supports three subnet layout strategies: Auto, Exact, and Legacy. Consider the following program that does not specify the subnet specs:

    new awsx.ec2.Vpc("vpc", {
        subnetStrategy: "Auto",
    });

Prior to this change, the code would implicitly use the Legacy strategy for this program even though Auto is requested. This is now fixed. The Auto strategy already supports allocating subnets without any explicit specs provided.

### DIFF
--- a/awsx/ec2/vpc.test.ts
+++ b/awsx/ec2/vpc.test.ts
@@ -303,10 +303,9 @@ describe("picking subnet allocator", () => {
     expect(a.allocator).toBe("LegacyAllocator");
   });
 
-  // This behavior is suspect as NewAllocator supports working without any subnet specs defined.
-  it("picks legacy allocator for the auto strategy with no specs", () => {
+  it("picks NewAllocator for the auto strategy with no specs", () => {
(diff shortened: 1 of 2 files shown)


corymhall (commented):
This is technically a breaking change right? Or does `Legacy` and `Auto` produce the same result from an undefined subnetStrategy?

t0yv0:
I can try to see if actual allocation without any subnetSpecs differs between legacy and auto allocator.

t0yv0:
It looks like legacy and auto strategies are generating the same layout. I've confirmed by provisioning this program on the prod version of the provider and testing upgrading the provider with this change. It's a no-change plan:

```
Resources:
    34 unchanged
```

t0yv0:
@corymhall can I have a quick review here? This simplifies case analysis in the PR that supports using IPAM with awsx.ec2.Vpc.

corymhall (approved):

Issue #1372, opened 2024-09-04
If no subnet strategy specified, you get the "outputs are not strings" error message


pierskarsenbarg:
### What happened?

I didn't add a `subnet_strategy` input to my VPC and got the following warning message:

```
warning: The default subnetStrategy will change from "Legacy" to "Auto" in the next major version. Please specify the subnetStrategy explicitly. The current subnet layout can be specified via "Auto" as:

    "Calling [toJSON] on an [Output<T>] is not supported.\n\nTo get the value of an Output as a JSON value or JSON string consider either:\n    1: o.apply(v => v.toJSON())\n    2: o.apply(v => JSON.stringify(v))\n\nSee https://www.pulumi.com/docs/concepts/inputs-outputs for more details.\nThis function may throw in a future version of @pulumi/pulumi."
```

which suggests that we're trying to write out an output<string> as a string.

### Example

```python
import pulumi_aws as aws
import pulumi_awsx as awsx

vpc = awsx.ec2.Vpc("gpu-vpc",
          cidr_block="10.0.0.0/16",
          subnet_specs=[awsx.ec2.SubnetSpecArgs(
              type=awsx.ec2.SubnetType.PUBLIC,
              name="gpu-public"
          )],
          nat_gateways=awsx.ec2.NatGatewayConfigurationArgs(
              strategy=awsx.ec2.NatGatewayStrategy.NONE
          ),
        #   subnet_strategy=awsx.ec2.SubnetAllocationStrategy.AUTO
          )
```

(note the subnet strategy line is commented out)

### Output of `pulumi about`

```
CLI
Version      3.130.0
Go Version   go1.22.6
Go Compiler  gc

Plugins
KIND      NAME    VERSION
resource  aws     6.50.1
resource  awsx    2.14.0
resource  docker  4.5.5
language  python  unknown

Host
OS       darwin
Version  14.6.1
Arch     arm64

This project is written in python: executable='/Users/piers/go/src/github.com/pierskarsenbarg/demos/python-ecs-gpu/venv/bin/python' version='3.12.5'

Current Stack: demo/python-ecs-gpu/dev

Found no resources associated with demo/dev

Found no pending operations associated with demo/dev

Backend
Name           pulumi.com
URL            https://app.pulumi.com/pierskarsenbarg
User           pierskarsenbarg
Organizations  pierskarsenbarg, team-ce, gitlab-test-piers, demo
Token type     personal

Dependencies:
NAME         VERSION
[...]

PR #1374, opened 2024-09-04
Rewrite a warning to avoid printing an output directly

mikhailshilkov:
We can't use JSON.stringify on an output directly, we need to do that inside an apply. 

Fix https://github.com/pulumi/pulumi-awsx/issues/1372

### DIFF
--- a/awsx/ec2/vpc.ts
+++ b/awsx/ec2/vpc.ts
@@ -489,13 +489,13 @@ export class Vpc extends schema.Vpc<VpcData> {
             .output(parsedSpecs?.normalizedSpecs)
             .apply(vpcConverters.toResolvedSubnetSpecOutputs);
 
-    // Only warn if they're using a custom, non-explicit layout and haven't specified a strategy.
-    if (
-      args.subnetStrategy === undefined &&
-      parsedSpecs !== undefined &&
-      parsedSpecs.isExplicitLayout === false
-    ) {
-      subnetLayout.apply((sl) => {
+    const verifiedSubnetLayout = subnetLayout.apply((sl) => {
+      // Only warn if they're using a custom, non-explicit layout and haven't specified a strategy.
+      if (
+        args.subnetStrategy === undefined &&
+        parsedSpecs !== undefined &&
+        parsedSpecs.isExplicitLayout === false
+      ) {
         pulumi.log.warn(
           `The default subnetStrategy will change from "Legacy" to "Auto" in the next major version. Please specify the subnetStrategy explicitly. The current subnet layout can be specified via "Auto" as:\n\n${JSON.stringify(
             sl,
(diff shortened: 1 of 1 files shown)


awsx/ec2/vpc.ts
```diff
@@ -495,14 +495,16 @@ export class Vpc extends schema.Vpc<VpcData> {
...
     ) {
-      pulumi.log.warn(
-        `The default subnetStrategy will change from "Legacy" to "Auto" in the next major version. Please specify the subnetStrategy explicitly. The current subnet layout can be specified via "Auto" as:\n\n${JSON.stringify(
-          subnetLayout,
-          undefined,
-          2,
-        )}`,
-        this,
-      );
+      subnetLayout.apply(sl => {
```
flostadler, line 498: Good catch!
Is this `apply` guaranteed to execute/complete? What if we return the `subnetLayout` again in the apply and thread that through to the function's return statement to ensure it's executed?

Something like:
```// Only warn if they're using a custom, non-explicit layout and haven't specified a strategy.
    const verifiedSubnetLayout = subnetLayout.apply(sl => {
      if (
        args.subnetStrategy === undefined &&
        parsedSpecs !== undefined &&
        parsedSpecs.isExplicitLayout === false
      ) {
        pulumi.log.warn(
          `The default subnetStrategy will change from "Legacy" to "Auto" in the next major version. Please specify the subnetStrategy explicitly. The current subnet layout can be specified via "Auto" as:\n\n${JSON.stringify(
            sl,
            undefined,
            2,
          )}`,
          this,
        );
      }

      return sl;
    });
    
    return { subnetLayout: verifiedSubnetLayout, subnetSpecs };
```

flostadler (commented):
Good Catch!

awsx/ec2/vpc.ts
mikhailshilkov: It won't execute until the Output is resolved, but then it will. I don't think that your change will make a difference in practice - but maybe it's a win in terms of readability and not provoking future engineers to ask this question again.

flostadler (approved):
LGTM

awsx/ec2/vpc.ts
t0yv0: Eh, I wouldn't be sure. Programs await all outstanding apply threads before exiting the program, but I'm not 100% sure the same is certain of code inside MLC contexts . So it might be possible this introduces a race. 

If it is the case however the best place to fix it would be the node SDK not the program here.

t0yv0 (approved):

awsx/ec2/vpc.ts
t0yv0: In cases where it's' critical to await some effect and we're still not sure about the above, one trick is to enforce ordering by making the effect output an antecedent of some awaited output. 

```
foo(): Output<X> {
    effectOutput = something.apply(xValue => doSometingImportant(xValue))
    outputX = ...
    return effectOutput.apply(_ => outputX)
}
```

Issue #1373, opened 2024-09-04
VPC with endpoints does not associate the endpoints with subnets


omercnet:
### What happened?

created a vpc with an endpoint, the endpoint was created but was not associated with the subnets created by awsx

### Example

```typescript
new awsx.ec2.Vpc(
    args.name,
    {
      enableDnsHostnames: true,
      enableDnsSupport: true,
      vpcEndpointSpecs: [
        { serviceName: `com.amazonaws.${region}.ecr.dkr`, vpcEndpointType: "Interface" }
      ],
    },
    opts,
  );
```

### Output of `pulumi about`

```
CLI          
Version      3.129.0
Go Version   go1.22.6
Go Compiler  gc

Plugins
KIND      NAME    VERSION
language  nodejs  unknown

Host     
OS       darwin
Version  14.5
Arch     arm64
```

### Additional context

I'm expecting awsx to attach the private/public subnets it creates to the endpoint
it's impossible to reference those subnets before they're created..

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Thanks for reporting @omercnet ! Looks like we have an omission here, and VPC endpoints will respect manually provided subnets but not an automatically computed subnet layout. My team will have a look as time permits, something along the lines of https://github.com/pulumi/pulumi-awsx/pull/1375 can be done to fix (I'm sure there's some more subtleties here).

PR #1375, opened 2024-09-04
Toward #1373: respect auto-allocated subnets in vpcEndpoints

t0yv0:


t0yv0:
This will need some careful testing. Closing for now to revisit as time permits.

PR #1395, opened 2024-10-11
Rewrite a warning to avoid printing sub-properties as outputs

mikhailshilkov:
Another fix for https://github.com/pulumi/pulumi-awsx/issues/1372 while https://github.com/pulumi/pulumi-awsx/pull/1374 only partially succeeded.

The previous attempt was successful in resolving the top-level output but child outputs were still unresolved. I was a bit surprised that was the case... outputs are hard!

The new attempt relies on `pulumi.jsonStringify` which is made exactly for this purpose. This drops JSON formatting but hopefully that not too bad for a small JSON. Example of the result:

```
 warning: The default subnetStrategy will change from "Legacy" to "Auto" in the next major version. Please specify the subnetStrategy explicitly. The current subnet layout can be specified via "Auto" as:
    
    [{"cidrMask":20,"type":"Public"}]
```

### DIFF
--- a/awsx/ec2/vpc.ts
+++ b/awsx/ec2/vpc.ts
@@ -489,7 +489,7 @@ export class Vpc extends schema.Vpc<VpcData> {
             .output(parsedSpecs?.normalizedSpecs)
             .apply(vpcConverters.toResolvedSubnetSpecOutputs);
 
-    const verifiedSubnetLayout = subnetLayout.apply((sl) => {
+    const verifiedSubnetLayout = pulumi.jsonStringify(subnetLayout).apply((sl) => {
       // Only warn if they're using a custom, non-explicit layout and haven't specified a strategy.
       if (
         args.subnetStrategy === undefined &&
@@ -497,15 +497,11 @@ export class Vpc extends schema.Vpc<VpcData> {
         parsedSpecs.isExplicitLayout === false
       ) {
         pulumi.log.warn(
-          `The default subnetStrategy will change from "Legacy" to "Auto" in the next major version. Please specify the subnetStrategy explicitly. The current subnet layout can be specified via "Auto" as:\n\n${JSON.stringify(
-            sl,
-            undefined,
-            2,
-          )}`,
+          `The default subnetStrategy will change from "Legacy" to "Auto" in the next major version. Please specify the subnetStrategy explicitly. The current subnet layout can be specified via "Auto" as:\n\n${sl}`,
           this,
         );
       }
-      return sl;
+      return subnetLayout;
     });
 
     return { subnetLayout: verifiedSubnetLayout, subnetSpecs };


awsx/ec2/vpc.ts
```diff
@@ -489,23 +489,19 @@ export class Vpc extends schema.Vpc<VpcData> {
             .output(parsedSpecs?.normalizedSpecs)
             .apply(vpcConverters.toResolvedSubnetSpecOutputs);
 
-    const verifiedSubnetLayout = subnetLayout.apply((sl) => {
+    const verifiedSubnetLayout = pulumi.jsonStringify(subnetLayout).apply(sl => {
```
flostadler, line 492: Why does `pulumi.jsonStringify` handle this better than a regular apply?

All it seems to be doing is calling apply on the output: https://github.com/pulumi/pulumi/blob/e34b753f070b2862263f80882ceb88ce69a14272/sdk/nodejs/output.ts#L1170-L1172

Is it that it's wrapping it in `pulumi.output()` again? If so, why do we need double wrapping in output here?

mikhailshilkov: Yes, here is the doc for `pulumi.output()`:

```
output takes any Input value and converts it into an Output, deeply unwrapping nested Input values as necessary.

The expected way to use this function is like so:

var transformed = pulumi.output(someVal).apply(unwrapped => {
   // Do whatever you want now.  'unwrapped' will contain no outputs/promises inside
   // here, so you can easily do whatever sort of transformation is most convenient.
});
```

Interestingly, my initial implementation works just fine for a Node.js program (that's how I tested it) but doesn't work for a Python program (provided by Piers). I don't quite understand where this difference comes from, but it must be about the exact shape of the output that we need to log.

flostadler: So that means by doing an `apply` on a `pulumi.output` we're loosing some of the lifting abilities?
My confusing here comes from the fact that we are doing `pulumi.output` in 482-490. It's just that we `apply` on it again.

flostadler: That almost seems like a bug to me, tbh

mikhailshilkov: I think it's by design, although it's very confusing and subtle. Note the types of the `apply` argument in each case:

```
// value.apply:
func: (t: { type: pulumi.Output<string>;}) =>
    pulumi.OutputInstance<{type: pulumi.Output<string>;}>

// pulumi.output(value).apply:
(t: pulumi.UnwrappedObject<{  type: pulumi.Output<string>; }>) =>     pulumi.UnwrappedObject<{ type: pulumi.Output<string>; }>

// pulumi.jsonStringify(value).apply:
func: (t: string) => string
```

`UnwrappedObject` is a Mapped Type that removes outputs from properties, so it's plain. But in the first case, you see you get something that isn't safe to serialize.

flostadler (approved):

PR #1405, opened 2024-11-05
First pass at automating keeping pulumi-aws updated

t0yv0:
TBD


flostadler (approved):
This looks good to me, but as a follow up we should automate updating the pu-docker dependency as well.

PR #1458, opened 2025-01-02
chore: fix renovate

corymhall:
It looks like renovate cannot run any of the build targets because it
doesn't have the language tools installed. Splitting out the generate
steps from the build steps

re #1457


flostadler (approved):
LGTM! This also aligns the components, EKS seems to already have this generate/build split!

Issue #1203, opened 2024-01-11
ECR Image: push without build


mattfysh:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently the image being pushed to ECR must be also built by Pulumi, which is a rather uncommon setup in projects where the build steps are separated from the deployment steps. The schema of the resource allows for no build args to be provided, but fails when it cannot locate a Dockerfile

It would be great if there were a way to push an image to the ECR registry if it has already been built and tagged on the local host.

See https://github.com/pulumi/pulumi-docker/issues/54 for a similar discussion on the docker provider. It is suggested that if no build is required, a `docker.RegistryImage` should be created instead of `docker.Image`

### Affected area/feature

A possible solution is to support an `imageName` input for the Image, and branch the internal logic based on its presence:

```ts
import * as awsx from "@pulumi/awsx";
const repository = new awsx.ecr.Repository("repository", {
    forceDelete: true,
});
const image = new awsx.ecr.Image("image", {
    repositoryUrl: repository.url,
    imageName: 'my-locally-built-image',
});
```

mjeffryes:
Thanks for this suggestion @mattfysh; we've been talking a bit about this use case in the context of some of our docker work too recently. We'll try to think about how to enable this scenario in awsx as well. cc @blampe

mattfysh:
let me know if this is off topic, but I was looking into the implementation of `ecr.Image` and noticed that a new image is built and pushed on each program run only because `aws.ecr.getCredentialsOutput` will return new credentials each time. If (for example) the same creds token was returned then pulumi wouldn't invoke the resource, so a new image wouldn't be built/pushed even if the context or Dockerfile has changed

flostadler:
I started working on this and created a prototype. There's still some rough edges though; this is affected by https://github.com/pulumi/pulumi-docker/issues/952.

I'm working on an enhancement to the docker provider to ignore changes to the volatile parts of the auth data (username & password)

flostadler:
This is currently blocked on https://github.com/pulumi/pulumi-docker/issues/952. We'll first need to first iron out issues with the provider showing updates on every run because of the changing credentials

flostadler:
We made progress on https://github.com/pulumi/pulumi-docker/issues/952 and once it is resolved we can tackle this here.

[...]

PR #1464, opened 2025-01-06
Add ecr.RegistryImage resource for pushing existing images to ECR

flostadler:
The current `awsx.ecr.Image` resource is limited to building and pushing new images to ECR. However, many CI/CD workflows need to push pre-built images that may come from separate build processes or workflows.
This PR introduces a new `awsx.ecr.RegistryImage` component that enables pushing existing Docker images to ECR. The component extends the functionality of `docker.RegistryImage` with improved change detection capabilities.
The component handles scenarios where the source image tag remains constant (e.g., `my-image:latest`) but the underlying image content has changed. Instead of relying solely on the image name for change detection, it uses the image's unique ID to determine when updates are needed. This ensures proper triggering of infrastructure updates when the source image was rebuilt.

Resolves https://github.com/pulumi/pulumi-awsx/issues/1203
Blocked by https://github.com/pulumi/pulumi-docker/issues/952


awsx/ecr/registryImage.ts
```diff
@@ -0,0 +1,83 @@
...
+    args: schema.RegistryImageArgs,
+    opts?: pulumi.ComponentResourceOptions,
+  ) {
+    super(name, args, opts);
+
+    const creds = pulumi
+      .output(args.repositoryUrl)
+      .apply((url) => getDockerCredentials({ registryUrl: url }, { parent: this }));
+    const provider = new docker.Provider(name, {
+      registryAuth: [creds],
```
flostadler, line 32: Note to myself, this should also be parented

awsx/ecr/registryImage.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  const parts = str.split(":");
+  if (parts.length < 2) {
+    // if there is no tag, use the latest tag
+    return "latest";
+  }
+
+  // the tag is the last part of the image, tags are not allowed to contain colons so we can just take the last part
+  return parts[parts.length - 1];
```
flostadler, line 82: Can we detect whether that's the actual tag or just the port number of the registry host

corymhall (approved):
🚀

awsx/ecr/auth.ts
```diff
@@ -0,0 +1,61 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+export interface CredentialArgs {
```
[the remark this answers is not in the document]
flostadler: Done

t0yv0:
Curious about this as I don't have a good intuition on our Docker support. My one question is adding new resources creates apparent complexity for the user, no?  The matrix of possibilities is:

```
Image source: (1) prebuilt image; (2) need to build it
Destination: (1) ECR; (2) somewhere else? (3) nowhere, just build it 
```

Does this matrix have very different Pulumi resources involved depending on what you are doing, and does this new resource make it even more complicated? 

https://www.pulumi.com/registry/packages/docker-build/#example seems to be able to push to ECR, can it push a prebuilt image to ECR? I think yes?

flostadler:
> Curious about this as I don't have a good intuition on our Docker support. My one question is adding new resources creates apparent complexity for the user, no? The matrix of possibilities is:
> 
> ```
> Image source: (1) prebuilt image; (2) need to build it
> Destination: (1) ECR; (2) somewhere else? (3) nowhere, just build it 
> ```
> 
> Does this matrix have very different Pulumi resources involved depending on what you are doing, and does this new resource make it even more complicated?
> 
> https://www.pulumi.com/registry/packages/docker-build/#example seems to be able to push to ECR, can it push a prebuilt image to ECR? I think yes?

The only resource that can push prebuilt images is docker.RegistryImage. (docker-build cannot)

Right now users would need to handle the auth bit and re-triggering the auth bit when the source image changes. This is what this PR addresses. Given that the feature request has 14 upvotes I’d argue that users have an appetite for this.

awsx/ecr/auth.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (args.registryId) {
+    registryId = args.registryId;
+  } else {
+    // add protocol to help parse the url
+    const registryUrl = args.registryUrl?.startsWith("https://")
+      ? args.registryUrl
+      : "https://" + args.registryUrl;
+    const parsedUrl = new URL(registryUrl);
+    // the registry id is the AWS account id. It's the first part of the hostname
+    registryId = parsedUrl.hostname.split(".")[0];
```
t0yv0, line 76: Should this get a bit more error handling just in case, or we are certain it succeeds?

corymhall (approved):

awsx/ecr/auth.ts
flostadler: I've taken this from the existing implementation of `Image`, but you have a good point!
What could fail in theory is parsing the URL (i.e. `new URL(...)`), doesn't hurt to add some error handling for that to return better errors.

awsx/ecr/registryImage.ts
```diff
@@ -0,0 +1,65 @@
...
+    name: string,
+    args: schema.RegistryImageArgs,
+    opts?: pulumi.ComponentResourceOptions,
+  ) {
+    super(name, args, opts);
+
+    const creds = pulumi
+      .output(args.repositoryUrl)
+      .apply((url) => getDockerCredentials({ registryUrl: url }, { parent: this }));
+    const provider = new docker.Provider(name, {
```
t0yv0, line 31: There is no way for the user to customize https://www.pulumi.com/registry/packages/docker/#remote-hosts or anything else about this provider. Possibly OK for now just noting.

awsx/schema-types.ts
```diff
@@ -118,6 +119,20 @@ export interface ImageArgs {
     readonly repositoryUrl: pulumi.Input<string>;
     readonly target?: pulumi.Input<string>;
 }
+export abstract class RegistryImage<TData = any> extends (pulumi.ComponentResource)<TData> {
+    public image!: unknown | pulumi.Output<unknown>;
```
t0yv0, line 123: What is `unknown` here exactly?

examples/ts-ecr-registry-image/README.md
```diff
@@ -0,0 +1,5 @@
+# Pushing a local image to an ECR registry
+
+This example demonstrates how to push a local image to an ECR registry using the `awsx.ecr.RegistryImage` component.
```
t0yv0, line 3: Nice

schemagen/pkg/gen/docs/ecr/registry-image.md
```diff
@@ -0,0 +1,117 @@
+Manages the lifecycle of a docker image in a registry. You can upload images to a registry (= `docker push`) and also delete them again. In contrast to [`awsx.ecr.Image`](/registry/packages/awsx/api-docs/ecr/image/), this resource does not require to build the image, but can be used to push an existing image to an ECR repository. The image will be pushed whenever the source image changes or is updated.
+
+{{% examples %}}
+## Example Usage
+{{% example %}}
+### Pushing an image to an ECR repository
+
+```typescript
```
t0yv0, line 8: Could be cool to source from examples but that's for another time :)

t0yv0 (approved):

awsx/schema-types.ts
flostadler: Uh that's interesting, good catch!
Seems like the script for generating provider internal types has a bug or cannot properly handle references to docker??

The SDK and schema look good, but I'm gonna dig into this

awsx/ecr/registryImage.ts
flostadler: Good point, I can quickly expose those. I don't think those are gonna be useful for most users (e.g. remote host) as part of pushing a local image, but it doesn't hurt to export them to support future use cases

awsx/schema-types.ts
t0yv0: Thanks!!

awsx/ecr/registryImage.ts
t0yv0: Users can still reach this code with transforms possibly, if really needed.

awsx/ecr/auth.ts
flostadler: Added error handling + tests for it

awsx/schema-types.ts
flostadler: Yeah, the type-gen for provider internal types was only configured to handle aws. I've added docker as well now

PR #1471, opened 2025-01-09
wip: test CI changes

t0yv0:


.upgrade-config.yml
```diff
@@ -0,0 +1,6 @@
+# WARNING: This file is autogenerated - changes will be overwritten when regenerated by https://github.com/pulumi/ci-mgmt
+
+---
+upstream-provider-name: terraform-provider-awsx
```
rquitales, line 4: Let's have a follow-up work item to remove referencing `terraform`. This could be confusing for users since awsx is not a bridged provider.

rquitales (approved):
LGTM. Changes came mostly from other PRs that were reviewed with scrutiny. A sanity check of this PR looks good.

.upgrade-config.yml
t0yv0: Good catch, looking.

t0yv0:
Looks like we need to disable branch protections temporarily and reset them via ci-mgmt.

PR #1472, opened 2025-01-09
chore: alias schema to a new location

t0yv0:
This is a temporary change to facilitate testing https://github.com/pulumi/pulumi-awsx/pull/1471

The schema-tools expect schema to be found under a certain path but the PR is moving the schema. The new location is enabled to reference the old location.


guineveresaenger (approved):
oh boy 😂

provider/cmd/pulumi-resource-awsx/schema.json
```diff
@@ -0,0 +1 @@
+../../../schema.json
```
corymhall, line 1: TIL!

corymhall (approved):

provider/cmd/pulumi-resource-awsx/schema.json
t0yv0: `ln -s`

PR #1485, opened 2025-01-15
chore: remove custom Make targets

t0yv0:
Applies https://github.com/pulumi/ci-mgmt/pull/1292 to this repository to remove the need for custom Make targets.


awsx/scripts/build.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
t0yv0, line 3: We might want to wait with this one until we have reliable x-platform verification of the binaries.

t0yv0:
The changes in ci-mgmt to ensure release verification on mac are now included.

guineveresaenger (approved):

PR #1515, opened 2025-02-07
feat: Add option to store docker build cache in ECR image resource

larrywax:
Add `cacheTo` arg to configure the underlying docker cacheTo option, a list of ECR registries is expected.
I added some sane defaults to the config, but I would love a feedback on that.

Note that `imageManifest: true` and `ociMediaTypes: true` are required for ECR registries, but I don't really know if they can cause issues with other types of registries. In any case, I don't think it's a big deal, this is a component tailor-made for AWS solutions.

It falls back to the old behaviour (inline cache) if no parameter is specified

### DIFF
--- a/awsx/ecr/image.ts
+++ b/awsx/ecr/image.ts
@@ -53,6 +53,7 @@ export function computeImageFromAsset(
     { parent },
   );
 
+  // CacheFrom setup
   let cacheFrom: docker.types.input.CacheFromArgs[] = [];
   if (dockerInputs.cacheFrom !== undefined) {
     cacheFrom = dockerInputs.cacheFrom.map((c) => {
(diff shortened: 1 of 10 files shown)


larrywax:
Hi, can someone give feedback on this?
Thanks

larrywax:
@blampe Thanks for your feedback!
My use case is exactly what you pointed as an issue in the default behaviour. I have a pulumi project that builds and pushes a multi-stage docker image and I want to speed up the build phase by leveraging docker cache. To do so effectively, I need to store the cache somewhere (ECR repository in my case).
I don't have strong requirements, I just implemented the 'MVP' without disrupting the current behaviour.

I'll be happy to update the PR based on your feedback. For instance, I can bridge the `cacheTo` option from the underlying provider to avoid imposing too many limitations.
Let me know!

blampe:
@larrywax 

> For instance, I can bridge the cacheTo option from the underlying provider to avoid imposing too many limitations.

This sounds reasonable to me, but I would also run it by @flostadler. He recently made some improvements to the Image resource and probably has a vision for how it should evolve.

guineveresaenger:
Hi @larrywax - are you still working on this?

larrywax:
> Hi @larrywax - are you still working on this?

Hi @guineveresaenger, I'm not unfortunately

PR #1580, opened 2025-04-04
[vpc/vpcEndpoints] use shared tags & expose endpoints

lukealbao:
👋 I had some internal tagging policies fail when using the awsx.ec2.Vpc. I included the shared tags for the failing children (vpc endpoints) here. When testing it, I discovered that the vpc weren't being added to the component data, so I added that too. Along the way, I wrote some tests for other things that might have had the same problem, and I left them because they seem benign enough, and maybe even helpful.

I'm not familiar with this codebase, but it looks reasonable that `build_sdks` produced no output.

### DIFF
--- a/awsx/ec2/vpc.test.ts
+++ b/awsx/ec2/vpc.test.ts
@@ -414,13 +414,6 @@ describe("child resource api", () => {
   });
 
   it("internetGateway", async () => {
-    const vpc = new Vpc("test", {
-      tags: {
-        share1: "parent",
-        share2: "parent",
-      },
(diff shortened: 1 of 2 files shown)


mjeffryes (approved):
Good catch! Looks like the shared tags and endpoints outputs were just overlooked when this was originally written. Thanks for the fix and for the new tests!

lukealbao:
@mjeffryes 👋 when should we expect this to be released?

mjeffryes:
Thanks for the reminder @lukealbao I've kicked off a release, it should be published within an hour or so

PR #1626, opened 2025-05-27
sort TaskDefinitionContainerDefinitionInputs by container name to reduce non-needed deploys

whutchinson98:
My team has been running into issues with unnecessary container task definition updates for our services which causes lost time in waiting for these changes to be deployed.

The cause of this was having multiple containers where the containers would be getting mapped to the task definition in a non-predictable manor which lead to false positive changes.

This PR addresses that issue by ensuring we sort the containers for a task definition by their name in an effort to reduce these false positive changes from occurring.

### DIFF
--- a/awsx/ecs/containers.ts
+++ b/awsx/ecs/containers.ts
@@ -48,9 +48,11 @@ export function computeContainerDefinitions(
 ): pulumi.Output<schema.TaskDefinitionContainerDefinitionInputs[]> {
   const computed = containers.apply((c) => {
     return pulumi.all(
-      Object.entries(c).map(([containerName, container]) =>
-        computeContainerDefinition(parent, containerName, container, logGroupId),
-      ),
+      Object.entries(c)
+        .sort(([a], [b]) => a.localeCompare(b)) // sort by container name to ensure deterministic order and reduce false positive redeploys
+        .map(([containerName, container]) =>
+          computeContainerDefinition(parent, containerName, container, logGroupId),
+        ),
     );
   });
   return computed;


blampe:
Hi @whutchinson98, thank you for the contribution!

I've run into a similar issue in the past -- https://github.com/pulumi/pulumi-aws/issues/4576 and https://github.com/pulumi/pulumi-aws/issues/1738.

Can you tell us a little more about what might be causing your nondeterminism? For example do you have any null (or empty) environment variables? (That would point to https://github.com/hashicorp/terraform-provider-aws/issues/39533.) Are you programmatically assembling your list of containers?

If you're able to share any related issues or sample code it would be very helpful!

whutchinson98:
> Hi @whutchinson98, thank you for the contribution!
> 
> I've run into a similar issue in the past -- [pulumi/pulumi-aws#4576](https://github.com/pulumi/pulumi-aws/issues/4576) and [pulumi/pulumi-aws#1738](https://github.com/pulumi/pulumi-aws/issues/1738).
> 
> Can you tell us a little more about what might be causing your nondeterminism? For example do you have any null (or empty) environment variables? (That would point to [hashicorp/terraform-provider-aws#39533](https://github.com/hashicorp/terraform-provider-aws/issues/39533).) Are you programmatically assembling your list of containers? Do you still have the issue if you sort your containers before passing them to Pulumi?
> 
> If you're able to share any related issues or sample code it would be very helpful!

Hi @blampe! Happy to provide more info

- I do not have any null/empty environment variables
- I am not programmatically assembling the list of containers
- I have tried sorting the containers which has made it less flaky but due to the `Object.entries` the order is not guaranteed so it'll still cause false positives.

I don't have any examples I can share off the top of my head but can create one if necessary

whutchinson98:
@blampe here is an example of the containers we use. Basically all our services have the "service" then a datadog sidecar setup

We've tried ordering them in various ways as mentioned above and as you see they all have env vars set that are not null

```typescript
const service = new awsx.ecs.FargateService(
  'service',
  {
    tags,
    cluster: ecsClusterArn,
    networkConfiguration: {
      subnets: vpc.privateSubnetIds,
      securityGroups: [this.serviceSg.id],
    },
    taskDefinitionArgs: {
      taskRole: {
        roleArn: this.role.arn,
      },
      containers: {

        log_router: {
          essential: true,
          image: 'amazon/aws-for-fluent-bit:latest',
          name: 'log_router',
          firelensConfiguration: {
            type: 'fluentbit',
            options: {
              'config-file-type': 'file',
              'config-file-value': '/fluent-bit/configs/parse-json.conf',
              'enable-ecs-log-metadata': 'true',
            },
          },
          environment: [
            {
              name: 'ECS_FARGATE',
              value: 'true',
            },
            {
              name: 'DD_API_KEY',
              value: DATADOG_API_KEY,
            },
          ],
          memoryReservation: 50,
        },
        datadog_agent: {
          name: 'datadog-agent',
          image: 'public.ecr.aws/datadog/agent:latest',
          environment: [
            {
              name: 'ECS_FARGATE',
              value: 'true',
            },
            {
              name: 'DD_SITE',
              value: 'us5.datadoghq.com',
            },
            {
              name: 'DD_API_KEY',
              value: DATADOG_API_KEY,
            },
          ],
          memoryReservation: 256,
        },
        service: {
          name: 'service',
          image: image.image.imageUri,
          cpu: 256,
          memory: 512,
          environment: [
            ...(containerEnvVars ?? []),
            {
              name: 'BASE_URL',
              value: 'url',
            },
          ],
          logConfiguration: {
            logDriver: 'awsfirelens',
            options: {
              Name: 'datadog',
              Host: 'HOST',
              apikey: DATADOG_API_KEY,
              dd_service: 'service',
              dd_source: 'fargate',
              dd_tags: 'project:service',
              provider: 'ecs',
            },
          },
          portMappings: [
            {
              appProtocol: 'http',
              name: 'tcp',
              hostPort: 8080,
              containerPort: 8080,
              targetGroup,
            },
          ],
        },
      },
      runtimePlatform: {
        operatingSystemFamily,
        cpuArchitecture,
      },
    },
    desiredCount: 1,
  },
  { parent: this }
);
```

whutchinson98:
> Sorry @whutchinson98 just now catching up on notifications. Thanks for posting that code, looking at it now it makes sense. I had forgotten about the "containers" arg and it makes sense that this map wouldn't reliably get serialized into a list of container definitions for AWS.
> 
> I do think it would be reasonable to sort the array computed by awsx.
> 
> The tricky piece is that this change will trigger a redeployment of existing services. That's not ideal... but I suppose they would be non-deterministically updating anyway, so it's a small price to pay for stability.

whutchinson98:
sorry @blampe I fat fingered and closed the PR

PR #1631, opened 2025-06-05
feat: Make fargate launchtype configurable

TheJokersThief:
# What

* Make the launch type for fargate services configurable

# Why

Enables us to use `FARGATE_SPOT` in our non-production environments instead of a hardcoded `FARGATE` :)


blampe:
Hi @TheJokersThief, can you please explain the motivation behind this change? There is already an [EC2Service](https://www.pulumi.com/registry/packages/awsx/api-docs/ecs/ec2service/) available. Is that not sufficient?

PR #1640, opened 2025-06-24
[TEST]: Upgrade to aws v7.0.0-alpha.1

corymhall:


examples/ecs-parallel/package.json
```diff
@@ -4,7 +4,7 @@
         "@types/node": "^10.0.0"
     },
     "dependencies": {
-        "@pulumi/aws": "^6.0.4",
+        "@pulumi/aws": "7.0.0-alpha.1",
```
mjeffryes, line 7: here and below, looks like we need to update to `v7.0.0`

Issue #1049, opened 2023-07-20
subnets created by awsx.ec2.vpc do not get ipv6 netblocks assigned


ritzk:
### What happened?

subnets created by awsx.ec2.vpc do not get ipv6 netblocks assigned.

Sample

```
const vpc = new awsx.ec2.Vpc(`k3s-vpc-${currentStack}`, {
    assignGeneratedIpv6CidrBlock: true,
    cidrBlock: cidrBlock,
    enableDnsHostnames: true,
    enableDnsSupport: true,
    natGateways: { strategy: "None" },
    subnetSpecs: [
        { type: awsx.ec2.SubnetType.Private, cidrMask: 24 },
        { type: awsx.ec2.SubnetType.Public, cidrMask: 24 }
    ],
});
```

### Expected Behavior

subnets created have an ipv6 address associated with them

### Steps to reproduce

run the code

### Output of `pulumi about`

CLI          
Version      3.75.0
Go Version   go1.20.5
Go Compiler  gc

Plugins
NAME        VERSION
aws         5.42.0
aws-native  0.69.0
awsx        1.0.2
docker      3.6.1
nodejs      unknown

Backend        
Name           pulumi.com

Dependencies:
NAME                          VERSION
@pulumi/aws                   5.42.0
@pulumi/awsx                  1.0.2
@pulumi/pulumi                3.76.0
@types/node                   18.16.19
ip-num                        1.5.1
typescript-string-operations  1.5.0
typescript                    5.1.6
@pulumi/aws-native            0.69.0

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Hi @ritzk thanks for reporting this issue. and apologies that you're experiencing this.

I've inspected the codebase closer, and it looks like v1.x of AWSx does not have any references to ipv6 logic, whereas the logic for [AWSx v0.40.0 does](https://github.com/pulumi/pulumi-awsx/blob/d798487f062748c4bc944c2f178f176fd617f029/awsx-classic/ec2/vpc.ts#L110-L121C30).

I'll surface this to the team and see if they have any input on this as well.

Jimmy89:
@rquitales is there any progress on this ticket? I have created a VPC with subnets through awsx as well and suffer from the same issue. Also, now lambda starts supporting IPv6, I want my subnet to have IPv6 support so I can attach an egress only internet gateway.

[...]

PR #1646, opened 2025-07-03
IPv6 Subnet CIDR generation and assignment

daniel-sheward:
Fixes https://github.com/pulumi/pulumi-awsx/issues/1049.

This restores functionality that appears to exist in previous versions.

I've modified the arguments to take `assignIpv6AddressOnCreation` which decides if a subnet should get an IPv6 CIDR.

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -356,9 +356,9 @@ func TestVpc(t *testing.T) {
 
 func TestIPv6Assignment(t *testing.T) {
 	validate := func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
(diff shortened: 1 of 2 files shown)


blampe:
>     TSError: ⨯ Unable to compile TypeScript:
>    index.ts(7,7): error TS2322: Type 'string' is not assignable to type '"Private" | "Public" | "Isolated" | "Unused"'.
>    index.ts(11,7): error TS2322: Type 'string' is not assignable to type '"Private" | "Public" | "Isolated" | "Unused"'.

I'm not able to repro this failure locally. Very odd, especially because there are several other tests with similar subnet specs (e.g. TestLbSimple).

PR #1678, opened 2025-08-24
Adding security groups to NLB args.

andreujuanc:
Noticed how it was not added, and explicitly removed from the NLB args.
Docs mentioned that it's available: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-security-groups.html
I guess it's something that it was not possible before.
I have also tried it myself, receipt:
image

No need to add any other lies as all the args are already passed to the underlying constructor. 
Would love to get this merged, such that I can remove my type hack.

Best,

J

### DIFF
--- a/awsx-classic/lb/loadBalancer.ts
+++ b/awsx-classic/lb/loadBalancer.ts
@@ -158,8 +158,7 @@ export interface LoadBalancerArgs {
     tags?: pulumi.Input<aws.Tags>;
 
     /**
-     * A list of security group IDs to assign to the LB. Only valid for Load Balancers of type
-     * `application`.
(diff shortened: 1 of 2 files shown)


andreujuanc:
CC: [corymhall](https://github.com/corymhall) 🙏

corymhall:
@andreujuanc thanks for taking the time to create the PR! Unfortunately the `awsx-classic` library has been replaced with `awsx` and we are no longer accepting contributions to it.

> The classic namespace will remain until the next major version release but will only receive updates for critical security fixes.

andreujuanc:
Hi @corymhall even if it's a bugfix? :(

Issue #1816, opened 2025-12-18
Model awsx-upgrade-aws and awsx-weekly-pulumi-update using Renovate


pose:
Currently, we are experiencing workflow failures related to runners running out of resources. See:
* https://github.com/pulumi/pulumi-awsx/issues/1786
* https://github.com/pulumi/pulumi-awsx/issues/1811

We decided we want to pursue managing these updates through Renovate instead of using custom scripts.

pose:
`<rant>` I'm extremely frustrated at how hard is to use renovate. Even when I pass `--help`, I get bombarded with so many options, modes, dry run, platform, tokens, etc. that it's hard to understand how to even invoke the CLI for very basic local use cases. I spent hours trying to make sense of it by playing with it, reading the docs and going over as much resources as I could.`</rant>`

My next action is to keep trying to make it work with my local fork by pushing change by change (see last section of this comment).

Here is all I tried:

###  `--platform=local` and `github>`

Full command: `LOG_LEVEL=debug npx renovate --platform=local --repository-cache=reset`

It failed to fetch the private config repo where we host the base configuration. I tried passing [`hostRules`](https://docs.renovatebot.com/configuration-options/#hostrules) to use a GitHub token with required permissions:

```
  "hostRules": [
    {
      "hostType": "github",
      "matchHost": "https://github.com",
      "token": "YOUR_SECRET_TOKEN"
    },
```

But I kept getting the error: `DEBUG: hostRules: no authentication for api.github.com (repository=local)`.

Also, it seems that [they don't recommend having a public repo pointing to a private config](https://docs.renovatebot.com/getting-started/private-packages/#private-config-presets). [This](https://github.com/renovatebot/renovate/discussions/21818) discussion recommendation didn't do the trick for me.

### `--platform=local` and  `local>` 

I pointed the `local>` part to the private shared-config repo. Then, I started getting the following error _"The platform you're using (local) does not support local presets."_.

Also, confusingly enough, Renovate's definition of local [means "same GitHub/gitlab host" and not local file system or local repo (according to its creator)](https://github.com/renovatebot/renovate/discussions/13408), but [the official docs](https://docs.renovatebot.com/modules/platform/local) mention: _"With the "local" platform you can perform dry runs of Renovate against the local file system. This can be handy when testing a new Renovate configuration for example"_

###  [`--require-config=ignored`](https://docs.renovatebot.com/self-hosted-configuration/#requireconfig)

[...]

PR #1829, opened 2025-12-23
Introduce Renovate for aws and pulumi version updates

pose:
Part of #1816

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -12,5 +12,27 @@
         executionMode: "branch", // Only run once.
       },
     },
+    {
+      matchManagers: ["gomod"],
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -12,5 +12,27 @@
...
+    {
+      matchManagers: ["gomod"],
+      matchPackageNames: ["github.com/pulumi/pulumi/pkg/v3", "github.com/pulumi/pulumi/sdk/v3"],
+      postUpgradeTasks: {
+        commands: ["make renovate"],
+        executionMode: "branch", // Only run once. 
+      },
+      // automatically bump these, don't require manual trigger
+      dependencyDashboardApproval: false,
+    },
```
pose: It's not clear to me (yet) how this is going to interact with the `pulumi` CLI version. I suspect this may not be enough.

Issue #1913, opened 2026-03-18
How to create a public-only VPC with SubnetAllocationStrategy.Auto after v3.2.0?


prryb:
## Context

We're running a public-only VPC (ECS Fargate with `assignPublicIp: true`, no NAT gateways) and recently upgraded from v3.1.0 to v3.2.0.

After the upgrade, `pulumi preview` shows new private subnets being created that weren't there before. We traced it to #1862 which merges partial `subnetSpecs` with defaults.

## Our configuration

```typescript
const vpc = new awsx.ec2.Vpc("vpc", {
  cidrBlock: "10.0.0.0/16",
  numberOfAvailabilityZones: 2,
  natGateways: { strategy: awsx.ec2.NatGatewayStrategy.None },
  subnetStrategy: awsx.ec2.SubnetAllocationStrategy.Auto,
  subnetSpecs: [
    { type: "Public", cidrMask: 20 },
  ],
});
```

**v3.1.0:** Only public subnets created — expected.

**v3.2.0+:** Private subnets (with route tables) are added automatically, even though we only specified public.

## Question

Is there a recommended way to maintain a public-only VPC with `Auto` strategy in v3.2.0+? Or is switching to `Exact` the intended path for this use case?

For now we've stayed on v3.1.0 to keep our existing infrastructure stable. Happy to upgrade once we know the right approach — thanks!

Issue #1746, opened 2025-10-25
Unable to apply tags to automatically generated subnets


atrauzzi:
### Describe what happened

When using Awsx and an auto mode VPC, I can't tag a subnet for EKS auto as per: https://docs.aws.amazon.com/eks/latest/userguide/tag-subnets-auto.html

It seems like perhaps providing any kind of specs for subnets results in all auto behaviour being suppressed?

### Sample program

```csharp
var vpc = new Vpc("application", new()
    {
        SubnetStrategy = SubnetAllocationStrategy.Auto,
        SubnetSpecs =
        {
            new()
            {
                Type = SubnetType.Public,
                Tags =
                {
                    { "kubernetes.io/role/elb", "1" },
                },
            },
        },
    });
```

### Output of `pulumi about`

CLI          
Version      3.193.0
Go Version   go1.25.1
Go Compiler  gc

Plugins
KIND      NAME          VERSION
resource  aws           7.7.0
resource  awsx          3.0.0
resource  docker        4.8.0
resource  docker-build  0.0.12
language  dotnet        unknown

Host     
OS       fedora
Version  42
Arch     x86_64

This project is written in dotnet: executable='/usr/bin/dotnet' version='9.0.110'

### Additional context

I need a way to apply tags to subnets without necessarily taking full control over them.  Ideally anything I provide should just be merged in with whatever definitions Awsx is generating for me.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

atrauzzi:
I was somewhat lead to believe that AWSX was a convenience library for common patterns and best practices by the Pulumi docs.  Is this library being deprecated?

The issue seems to be more around just the absence of the ability to flow tags to the resources AWSX creates, rather than any specific challenges of the provider, no?

Graham-Pedersen:
AWSX Is not currently under active development and I do not know very much about the best practices niche it may have once held. Sadly this means it. takes us a long time to get to support requests like this.

Also I am realizing I may not understand what you are trying to do. You say that you "need a way to apply tags to subnets without necessarily taking full control over them".

By this do you mean that the subnet is not modeled or managed in your Pulumi program? So you need a way to identify and apply a tag? After glancing again at the AWSX provider, I can create a new subnet and set the kubernetes tag when I create it just fine.

corymhall:
[...]

PR #1917, opened 2026-03-25
Add AutoMerge subnet allocation strategy

corymhall:
## Summary

This adds a new `AutoMerge` subnet allocation strategy for `awsx.ec2.Vpc`.

The motivation is to support the use case from #1746: users want to keep AWSX's generated default subnet layout, but still customize the generated subnet specs, for example by adding EKS load balancer discovery tags such as `kubernetes.io/role/elb`.

We originally tried to make `Auto` merge partial subnet specs with the default layout, but that changed the meaning of existing `Auto` programs. In particular, #1913 showed that public-only VPCs using `Auto` started planning new private subnets after upgrading. This PR keeps `Auto` backward-compatible and makes the default-layout-plus-overrides behavior explicit through the new `AutoMerge` strategy.

## Behavior

With `subnetStrategy: "AutoMerge"`, AWSX starts from the canonical generated public/private layout and overlays user-provided settings onto matching default subnet types.

For example, this keeps the default private subnets while tagging the generated public subnets:

```ts
const vpc = new awsx.ec2.Vpc("application", {
  subnetStrategy: "AutoMerge",
  subnetSpecs: [
    {
      type: "Public",
      tags: {
        "kubernetes.io/role/elb": "1",
      },
    },
  ],
});
```

`Auto` remains unchanged: partial `subnetSpecs` continue to mean "allocate exactly these subnet specs automatically," which preserves existing public-only and custom-layout programs.

## Compatibility

- `Auto` keeps its existing semantics.
- `AutoMerge` is opt-in and additive in the schema/SDK surface.
- The implementation preserves the canonical default subnet ordering and CIDR allocation when user settings are merged.
- Explicit `cidrBlocks` are rejected for `AutoMerge`; users who want to fully own the exact CIDR layout should continue using `Auto` or `Exact`.

## Testing

- `yarn --cwd awsx test --runInBand ec2/subnetDistributorNew.test.ts ec2/vpc.test.ts`
- `make test_provider`
- `yarn --cwd awsx tsc --noEmit --incremental false`
- `make generate`


awsx/ec2/subnetDistributorNew.test.ts
```diff
@@ -433,6 +436,126 @@ describe("explicit subnet layouts", () => {
   });
 });
 
+describe("merging SubnetSpecs with defaults for AutoMerge", () => {
```
pose, line 439: (minor) these tests are a bit repetitive, I wonder if we could use `.each`

awsx/ec2/vpc.test.ts
```diff
@@ -593,3 +599,201 @@ describe("child resource api", () => {
   });
 
 });
+
+describe("AutoMerge strategy merges partial SubnetSpecs with defaults", () => {
+  function unwrap<T>(x: pulumi.Output<T> | T): Promise<T> {
```
pose: (nitpick) I wonder if Pulumi offers something out of the box like that

pose: (also seems that this is copied from above as well, we could potentially factor it out)

awsx/ec2/vpc.test.ts
```diff
@@ -593,3 +599,201 @@ describe("child resource api", () => {
...
+    expect(subnets).toHaveLength(3);
+
+    for (const subnet of subnets) {
+      const tags = await unwrap(subnet.tags);
+      expect(tags?.SubnetType).toBe("Public");
+      expect(tags?.["kubernetes.io/role/elb"]).toBe("1");
+    }
+  });
+
+  it("keeps Auto backward-compatible for the #1913 public-only cidrMask case", async () => {
```
pose, line 698: I wonder if these tests should be snapshot tests where we could just diff the output of the subnets and have less lengthy test code

pose (approved):
Nothing blocking. Just readability comments.

PR #1927, opened 2026-04-08
Relax NAT Gateway validation to allow more flexible subnet configurations

### LINKED ISSUE #966: NAT Gateway validation too strict
## Hello!

## Issue details

As the original author of the multi-language VPC component, I am willing to admit defeat and say that the validation on NAT Gateway strategy is a bad idea.

I have a scenario where I want to create isolated and public subnets with a NAT Gateway. I want isolated subnets because I want to create a route to the NAT Gateway for the first part of a workshop, but later change it to a route to a firewall. This is not currently possible because NAT Gateway strategy != None will throw an exception if there are not both public and private subnets specified. This is overly rigid.

jkodroff:
The previous validation required both private AND public subnets when using NAT Gateway strategy (OnePerAz or Single). This was overly restrictive.

The NAT Gateway only needs to be hosted in a public subnet. Both private and isolated subnets can route through it. This change relaxes the validation to only require public subnets, allowing configurations like:
- Public + Isolated subnets (the use case described in #966)
- Public + Private subnets (still works)
- Public subnets only (also valid)

Fixes #966

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/awsx/ec2/vpc.test.ts
+++ b/awsx/ec2/vpc.test.ts
@@ -91,14 +91,17 @@ describe("validateNatGatewayStrategy", () => {
       it("should succeed if there's public and private subnets", () =>
         runTest(strategy, ["Public", "Private"], false));
 
+      it("should succeed if there's public and isolated subnets", () =>
+        runTest(strategy, ["Public", "Isolated"], false));
+
+      it("should succeed if there are only public subnets", () =>
+        runTest(strategy, ["Public"], false));
+
       it("should throw an exception if there are only isolated subnets", () =>
-        runTest(strategy, ["Isolated"], true, "both private and public subnets"));
+        runTest(strategy, ["Isolated"], true, "public subnets must be declared"));
 
       it("should throw an exception if there are only private subnets", () =>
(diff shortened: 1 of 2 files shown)


awsx/ec2/vpc.test.ts
```diff
@@ -91,14 +91,17 @@ describe("validateNatGatewayStrategy", () => {
...
+        runTest(strategy, ["Public"], false));
+
       it("should throw an exception if there are only isolated subnets", () =>
-        runTest(strategy, ["Isolated"], true, "both private and public subnets"));
+        runTest(strategy, ["Isolated"], true, "public subnets must be declared"));
 
       it("should throw an exception if there are only private subnets", () =>
-        runTest(strategy, ["Private"], true, "both private and public subnets"));
-
-      it("should throw an exception if there are only public subnets", () =>
```
corymhall, line 100: I think we should still throw an error if the user has only configured public subnets. While it is technically possible to create a NAT gateway with only public subnets I think it is not a _valid_ configuration

corymhall (changes requested):

corymhall (approved):

PR #1943, opened 2026-04-14
Fix awsx:ecr:Image refresh crash after 2.19 upgrade

### LINKED ISSUE #1926: awsx:ecr:Image: pulumi refresh fails with `Cannot read properties of undefined (reading 'replace')`
### Describe what happened

When running `pulumi refresh` I'm encountering the following error:

```
error: awsx:ecr:Image resource 'myimage' has a problem: Cannot read properties of undefined (reading 'replace')
    error: TypeError: Cannot read properties of undefined (reading 'replace')
        at removeTagFromRef (/snapshot/awsx/bin/ecr/image.js:121:16)
        at applyHelperAsync (/snapshot/awsx/node_modules/@pulumi/pulumi/output.js:327:31)
        at /snapshot/awsx/node_modules/@pulumi/pulumi/output.js:222:63
        at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

It's actually blocking us from upgrading pulumi-aws provider from 6.x to 7.x, as [it's required to run `pulumi up --refresh --run-program` with the new provider](https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration/#prerequisites-to-upgrade-to-v700), but this also fails with the same issue.

I tried to run `pulumi refresh --run-program` with `--exclude` pointing to the problematic resource, but it wasn't helping (it seems `--exclude` is ignored when `--run-program` is specified).

### Sample program

```
lambda_repo = pulumi_awsx.ecr.Repository(
    "lambda",
)


lambda_image = pulumi_awsx.ecr.Image(
    "lambda",
    repository_url=lambda_repo.url,
    image_name="lambda",
    context="../lambda",
    platform="linux/amd64",
    opts=pulumi.ResourceOptions(parent=lambda_repo),
)
```

### Log output

_No response_

### Affected Resource(s)

`awsx:ecr:Image`

### Output of `pulumi about`

```
➜  pulumi about
CLI
Version      3.224.0
Go Version   go1.26.0
Go Compiler  gc

Plugins
KIND      NAME          VERSION
resource  aws           7.23.0
resource  awsx          3.3.0
resource  docker        4.11.1
resource  docker-build  0.0.15
resource  gcp           8.41.1
resource  opsgenie      1.3.20
language  python        unknown
resource  random        4.19.1
resource  tls           5.3.0

Host
OS       darwin
Version  26.3.1
Arch     arm64
```

### Additional context

_No response_

### Contributing

corymhall:
## Summary
- fix the `awsx.ecr.Image` refresh crash when upgrading stacks from `awsx` 2.19.0
- persist `imageUri` on the component and avoid rebuilding child resources during component rehydration
- add an upgrade repro that covers `2.19.0 -> current` with both `refresh --run-program` and `up --refresh --run-program`

## Motivation

This PR fixes the scenario reported in #1926.

The issue only shows up on an upgrade path. Fresh `awsx` 3.x stacks do not reproduce it.

The relevant sequence is:

1. A stack is originally deployed with `awsx` 2.19.0.
2. `awsx.ecr.Image` creates a `docker:index/image:Image` child resource.
3. That old child stores `repoDigest`, and the component returns a digest-form image URI like:
   `repo@sha256:...`
4. The stack is later upgraded to current `awsx`.
5. The user runs `pulumi refresh --run-program`.

At that point, the new implementation expects the `docker-build:index:Image` world (`digest` / `ref`), but the old stack still contains the legacy `docker:index/image:Image` child.

## What Was Happening

There are really two separate issues here:

### 1. Refresh crash on upgrade from 2.19.0

Current `awsx.ecr.Image` derives its output from `docker-build.Image.ref`.

On the `2.19.0 -> current` upgrade path during `refresh --run-program`, the child resource has not yet been migrated to `docker-build:index:Image`, so there is no usable `ref` from the new child shape.

That led to:
- `image.ref` being `undefined`
- `removeTagFromRef(image.ref)` throwing
- refresh failing

### 2. Missing persisted component output

Older `awsx.ecr.Image` did not call `registerOutputs`, so the component’s own `imageUri` output was not persisted in checkpoint state.

That means that on upgrade, the engine cannot rehydrate a previously stored `imageUri` for the component itself and instead has to recompute it from current program behavior.

### Fresh current-version stacks are fine

A fresh stack using current `awsx` does not reproduce this issue.

### `2.20.0 -> current` is also fine

`awsx` 2.20.0 had already switched to `docker-build:index:Image`, so those stacks already have `digest` / `ref` state and refresh behaves normally.

### The problematic path is specifically `2.19.0 -> current`

That is the version boundary where the child resource implementation changed from:
- `docker:index/image:Image`
to:
- `docker-build:index:Image`

### `refresh --run-program` and `up --refresh --run-program` behave differently

For stacks originating on `2.19.0`:

- `refresh --run-program` does not complete the child-resource migration
- `up --refresh --run-program` does replace the old child with the new `docker-build:index:Image`

So the crash happens on refresh, but a full update is able to converge the stack.

## Root Cause

The root cause is the combination of:

- a resource type migration from `docker.Image` to `docker-build.Image`
- old `2.19.0` state containing `repoDigest` rather than `digest` / `ref`
- `awsx.ecr.Image` assuming `ref` exists in the new world
- the component not having previously persisted its own `imageUri`

## Fix

This PR takes the narrow, safe fix:

1. If the component is being rehydrated (`opts.urn`), do not reconstruct child resources.
2. Call `registerOutputs({ imageUri })` so the component output is persisted going forward.
3. When deriving `imageUri` from `docker-build.Image`, prefer:
   - `digest`
   - then `ref`
   - then a final fallback to the canonical tag form

This is enough to eliminate the refresh crash and make the stack converge correctly on the subsequent update path.

## Behavioral Note

This PR is intentionally scoped to removing the refresh error.

For stacks that originated on `awsx` 2.19.0, a `refresh --run-program` can still temporarily rewrite the component output from digest form to tag form until a later:

`pulumi up --refresh --run-program`

completes the child-resource migration to `docker-build:index:Image`.

That behavior already self-heals once the update runs; this PR ensures refresh no longer crashes on the way there.

### Unit
- `yarn --cwd awsx test ecr/image.test.ts --runInBand`
- `yarn --cwd awsx tsc --noEmit`

### Repro coverage
Added a targeted upgrade repro that exercises:
- baseline deploy with `awsx` 2.19.0
- upgrade to current local provider
- `pulumi refresh --run-program`
- `pulumi up --refresh --run-program`

fixes #1926


examples/ts-ecr-refresh-repro-v3/index.ts
```diff
@@ -0,0 +1,9 @@
+import * as awsx from "@pulumi/awsx";
+
+const repository = new awsx.ecr.Repository("repository", { forceDelete: true });
+
+export const repositoryName = repository.repository.name;
+export const image = new awsx.ecr.Image("image", {
```
pose: @corymhall  Could this Pulumi program be made into a YAML program or since the component code is in TS it would be problematic?

pose (changes requested):

pose (approved):
Thanks for making the test a YAML test @corymhall

Issue #2010, opened 2026-05-28
Add nightly CI and pin sdk/nodejs dependencies so upstream type drift surfaces promptly


pose:
## Summary

`pulumi-awsx`'s `run-acceptance-tests.yml` only triggers on `pull_request` and `repository_dispatch`, never on push to master. Combined with caret-ranged unlocked deps in the generated `sdk/nodejs`, that means upstream typings can break the SDK build on master and stay invisible until the next PR forces a fresh install. Concretely this has happened: master CI has been red since around July 2025 and the breakage only surfaced on PR #2007 in late May 2026 (see #2008 for the surfacing errors and #2009 for the proposed code fix).

## Concrete evidence and impact

- Last green master `run-acceptance-tests-command` dispatch: 2025-06-16 (`cee5326`).
- Four consecutive red master dispatches: 2025-07-07 through 2025-07-11 (`4f67108` then `bdef4a2`).
- No master dispatch since 2025-07-11.
- Today's master tip: `182ec9d` (workflow update from 2026-05-12).
- PR #2007 is the first thing in roughly 11 months to re-trigger acceptance tests; that is when the existing breakage finally became visible.

### 1. Opt in to nightly CI via `ci-mgmt`

`pulumi/ci-mgmt` already generates a nightly workflow as an opt-in. Sibling provider repos enable it with a single line in `.ci-mgmt.yaml`:

```yaml
generate-nightly-test-workflow: true
```

When set, the next `ci-mgmt` regeneration produces `.github/workflows/nightly-test.yml` (cron-triggered). `pulumi-eks` is the directly comparable example: its `.ci-mgmt.yaml` has the flag and the autogenerated `nightly-test.yml` exists in its workflows directory. `pulumi-awsx`'s `.ci-mgmt.yaml` does not set the flag, so no nightly is emitted.

**Action:** add the flag to `pulumi-awsx/.ci-mgmt.yaml`, then regenerate via `ci-mgmt` to produce the nightly workflow. The PR will be the usual ci-mgmt-driven workflow update PR; reviewers should expect a new `nightly-test.yml` file in the diff.

### 2. Commit a lockfile for `sdk/nodejs` (or pin exact versions)

`sdk/nodejs/package.json` currently uses caret ranges:

```
"@pulumi/aws": "^7.24.0",
"@pulumi/pulumi": "^3.142.0",
```

[...]

PR #2011, opened 2026-05-28
Add nightly acceptance test workflow

pose:
## Summary

Enable the cron-triggered nightly acceptance run by setting `generate-nightly-test-workflow: true` in `.ci-mgmt.yaml`; regeneration produces `.github/workflows/nightly-test.yml` (daily cron `0 6 * * *`), matching `pulumi-eks`.

## Why

`run-acceptance-tests.yml` only fires on `pull_request` and `repository_dispatch`. With caret-ranged sub-provider deps and no committed lockfile in the generated `sdk/nodejs/`, an upstream typings change can rot an identical-source SDK build and stay invisible on master until the next PR forces a fresh install. A nightly run surfaces that drift within a day.

The dependency-pinning approach originally bundled here has been dropped: pinning the consumer-facing `sdk/nodejs/package.json` to exact sub-provider versions would prevent npm deduping and could leave consumers with multiple Pulumi provider/runtime copies in one tree. The nightly gives the drift protection without that downside.

## What changed

| File | Why |
| --- | --- |
| `.ci-mgmt.yaml` | Opt in to the nightly workflow generator (one-line flag). |
| `.github/workflows/nightly-test.yml` | Autogenerated from the flag above; cron `0 6 * * *`. |

Part of pulumi/pulumi-awsx#2010


provider/pkg/schemagen/schema.go
```diff
@@ -82,11 +82,15 @@ func GenerateSchema(packageDir string) schema.PackageSpec {
...
 			"nodejs": rawMessage(map[string]interface{}{
+				// Pulumi sub-provider deps are pinned to exact versions so that an upstream
+				// patch release with a breaking type change can't silently rot identical-source
+				// SDK builds between intentional bump PRs.
 				"dependencies": map[string]string{
 					"@aws-sdk/client-ecs":  "^3.405.0",
-					"@pulumi/aws":          "^" + dependencies.Aws,
-					"@pulumi/docker":       "^" + dependencies.Docker,
-					"@pulumi/docker-build": "^" + dependencies.DockerBuild,
+					"@pulumi/aws":          dependencies.Aws,
```
corymhall: I'm not sure that this is the change that we want to make. This will have an impact on consumers of awsx.

For example, a program that already depends on `@pulumi/aws@7.31.0` can currently satisfy `@pulumi/awsx`'s `^7.24.0` dependency with that same version. With this change, npm installs a nested `@pulumi/aws@7.24.0` under `@pulumi/awsx`. That freezes AWSX's dependency path to one provider/runtime patch and can leave consumers with multiple Pulumi provider/runtime versions in the tree.

corymhall (changes requested):
I'm not sure there is a good way around this other than maybe using lockfiles in `sdk/nodejs`.

provider/pkg/schemagen/schema.go
pose: Good catch, agreed. I've dropped the dependency-pinning change entirely from this PR rather than pin the consumer-facing `sdk/nodejs/package.json`, since exact pins would block npm deduping and risk multiple Pulumi provider/runtime copies in a consumer's tree. This PR is now nightly-CI only, which gives the drift protection without that downside. Reduced to two files and re-requested your review.

corymhall (approved):

PR #2028, opened 2026-06-05
feat(ec2): add VPC endpoint auto strategy

### LINKED ISSUE #1373: VPC with endpoints does not associate the endpoints with subnets

corymhall:
## Summary

Adds an opt-in `vpcEndpointStrategy: "Auto"` mode to `awsx.ec2.Vpc` while keeping the default behavior as `"Legacy"`.

Auto fills omitted VPC endpoint fields from resources generated by the same VPC component:

- Gateway endpoints get generated route table IDs when `routeTableIds` is omitted.
- Interface endpoints get generated subnet IDs when `subnetIds` and `subnetConfigurations` are omitted.
- Interface endpoints default `privateDnsEnabled` to `true` and set omitted VPC DNS attributes to `true`.
- Interface endpoints create a default endpoint security group and ingress rule when `securityGroupIds` is omitted.
- Other endpoint types remain pass-through.

This also adds the checked-in strategy spec, mock coverage for Legacy and Auto behavior, non-Java generated SDK updates, and a targeted AWS-backed acceptance example for DynamoDB Gateway plus SQS Interface endpoint traffic from isolated subnets.

## Compatibility

- Omitted `vpcEndpointStrategy` remains `Legacy`, preserving current pass-through behavior.
- `privateDnsEnabled` remains the existing plain boolean component input to avoid the Go SDK source break found during review.
- Java SDK generation is intentionally not included in this draft per follow-up; this PR includes schema plus NodeJS, Python, .NET, and Go generated outputs.

## Validation

- `make -B schema`
- `yarn --cwd awsx gen-types`
- `make -B generate_nodejs generate_python generate_dotnet generate_go`
- `yarn --cwd awsx test vpc.test.ts --runInBand`
- `make test_provider`
- `go test -tags nodejs ./... -run '^$'` from `examples/`
- `schema-tools compare -p awsx --old-path /tmp/awsx-schema-old.json --new-path provider/cmd/pulumi-resource-awsx/schema.json --summary`
- `git diff --check`
- `awsx/node_modules/.bin/tsc -p examples/vpc/nodejs/vpc-with-auto-service-endpoints/tsconfig.json --noEmit --skipLibCheck`

fixes https://github.com/pulumi/pulumi-awsx/issues/1373


corymhall:
@pulumi-eon review

docs/specs/vpc-endpoint-strategy.md
guineveresaenger, line 1: What is the intended use/audience for this file?

Is it a design/design decision doc? If so, are we developing a new practice of keeping those in-repo?
Is it meant to be user-facing? The placement and naming say yes; the content says no.
Is it for future bot use to quickly establish context?

My preference would be to place this into an _issue_, not an in-repo doc, but that may make discovery by coding assistants more difficult.

I'm happy with wherever we land but I'd like us to be intentional.

guineveresaenger (approved):
non-blocking: question about how we organize design and decision documentation 💟

docs/specs/vpc-endpoint-strategy.md
corymhall: Honestly I'm not sure. I've seen a lot of projects start putting these in repo somewhere which does allow agents better access. I think I'm going with this for now since the decision seems low risk. It's easy enough to convert them to issues, delete them, etc if we change our mind.

Issue #1933, opened 2026-04-13
Track region passthrough gaps in modern awsx components


corymhall:
## Summary
Several modern `awsx/` components now expose a top-level `region` input, but do not consistently pass that region through to all child AWS resources and region-sensitive invokes/lookups.

This surfaced concretely in `awsx:ec2:Vpc`, where child resources and AZ discovery could be evaluated in the ambient provider region instead of the component's `region` input.

## Background
`VpcArgs.region` was added as part of the AWS v7 schema surface refresh. The same pattern now appears to exist in other modern components that:
- expose a top-level `region` input
- create more than one child resource or perform region-sensitive lookups/invokes

## Scope
This tracking issue is for auditing and fixing the remaining modern components so that top-level `region` is consistently threaded through child resources and region-aware invokes.

## Known instance
- #1928

## Inventory to audit/fix
- `awsx:cloudtrail:Trail`
- `awsx:ecr:Repository`
- `awsx:ecs:EC2TaskDefinition`
- `awsx:ecs:FargateTaskDefinition`
- `awsx:ecs:EC2Service`
- `awsx:ecs:FargateService`
- `awsx:lb:ApplicationLoadBalancer`
- `awsx:lb:NetworkLoadBalancer`

## Notes
For this pass, the intended fix shape is simple:
- if a component accepts `region`, pass it into child AWS resource args that support `region`
- also pass it into AWS invokes/lookups whose args support `region`
- avoid introducing provider synthesis just to model this behavior

## Acceptance criteria
- Every affected modern component above is audited.
- Any child resources created by those components receive the top-level `region` when present.
- Any region-aware invokes/lookups performed by those components receive the top-level `region` when present.
- Repros/tests exist for at least the components that were changed.

PR #2058, opened 2026-06-25
fix(ecr): pass top-level region through to Repository lifecycle policy

### LINKED ISSUE #1935: Pass top-level region through Repository lifecycle resources
Parent: #1933

`awsx:ecr:Repository` exposes a top-level `region` input, but the implementation does not consistently thread that region through all child resources.

Known child/defaulted paths to audit:
- root `aws.ecr.Repository` resource
- auto-created `aws.ecr.LifecyclePolicy` resource

Implementation references:
- `awsx/schema-types.ts` (`RepositoryArgs.region`)
- `awsx/ecr/repository.ts`

Acceptance criteria:
- top-level `region` is passed to repository child AWS resources that support it
- targeted tests cover the changed behavior

mvanhorn:
## Summary
A user who sets `region` on an `awsx.ecr.Repository` expects every child resource to land in that region. Today the repository itself is created in the requested region, but its auto-created lifecycle policy is silently created in the ambient provider region instead. That inconsistency is the bug.

The component exposes a top-level `region` input, and the root `aws.ecr.Repository` already receives it via the `repoArgs` spread. The auto-created `aws.ecr.LifecyclePolicy`, however, was constructed with only `{ repository, policy }`, so it never saw `region`. This threads `args.region` through to the `LifecyclePolicy` when it is set, and leaves behavior unchanged when `region` is omitted.

Concretely, `awsx/ecr/repository.ts` now spreads `region: args.region` into the `aws.ecr.LifecyclePolicy` args only when `args.region` is defined, so the existing no-region path is byte-for-byte unchanged. A new `awsx/ecr/repository.test.ts` uses the `@pulumi/pulumi/runtime` mock harness (same style as `awsx/ec2/vpc.test.ts`) to cover: region threads to the lifecycle policy; region threads through the auto-generated default "remove untagged images" policy; `skip: true` creates no policy; and omitting `region` adds no `region` key (no regression). This change does not touch `provider/pkg/schemagen/**`, so no schema regeneration or committed SDK diffs are required.

Fixes #1935. This is one item in the region-passthrough audit tracked by #1933; it mirrors the merged Vpc fix in #1942.

## Compatibility
- [x] No public behavior/API change

## Generated Artifacts
- [x] N/A

## Validation
- [x] `awsx` Jest suite for the touched files (`yarn jest ecr/repository.test.ts`) - 4 passing
- [x] `yarn tsc --noEmit` - clean
- [x] No new lint errors introduced in the touched files

## Risk
Low blast radius. The change adds `region` to a single child resource only when the component's `region` input is set; the no-region path produces identical args to before, so existing stacks see no diff. The lifecycle policy already supports the `region` argument (the same provider-wide argument the Vpc fix used). Worst case if wrong, the lifecycle policy would target an unexpected region, which the new tests guard against.

## Rollback
Revert this commit. The lifecycle policy reverts to being created in the ambient provider region, restoring the prior behavior. No state migration or schema regeneration is involved.

AI was used for assistance.

### DIFF
--- a/awsx/ecr/repository.test.ts
+++ b/awsx/ecr/repository.test.ts
@@ -1,0 +1,112 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as runtime from "@pulumi/pulumi/runtime";
+import * as pulumi from "@pulumi/pulumi";
+import * as pulumiAws from "@pulumi/aws";
+import { Repository } from "./repository";
+
+function unwrap<T>(x: pulumi.Output<T> | T): Promise<T> {
+  return new Promise((resolve) => (pulumi.Output.isInstance(x) ? x.apply(resolve) : resolve(x)));
+}
+
+describe("Repository region passthrough", () => {
+  let newResources: any[] = [];
+
+  beforeAll(async () => {
+    await runtime.setMocks({
+      call(args) {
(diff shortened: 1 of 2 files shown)


corymhall (approved):

mvanhorn:
Thanks @corymhall! Region passthrough to the ECR lifecycle Repository was a subtle one.

mvanhorn:
Appreciate the merge, @corymhall! Region now reaching the Repository lifecycle policy closes that cross-region gap.

PR #2072, opened 2026-07-08
Allow configuring capacity providers on ECS Services

### LINKED ISSUE #1093: Reintroduce CapacityProviderService lost in 1.0.0
## Hello!

## Issue details

In #599 / #644, a new `CapacityProviderService` was introduced to allow using e.g. FARGATE_SPOT with ECS. [In 1.0.0-beta.1](https://github.com/pulumi/pulumi-awsx/compare/v0.40.1...v1.0.0-beta.1#diff-c1b4860a454225c1c49a136f374a070fdf53a0183cf348c9f6458bf95a5abe35), it was moved to "awsx-classic", and then never brought into awsx itself. That means it's now impossible again to use FARGATE_SPOT with this provider (unless one downgrades all the way to classic).

corymhall:
## Summary

This adds the ability to configure capacity providers on Fargate ECS services. It also fixes a bug in the EC2 ECS Service where you could not configure the capacityProviderStrategies due to the launchType always being provided.

In additional to being able to configure a capacity provider at the service level, you can also configure the cluster to have a default capacity provider. When this is set all services that are launched in the cluster will get that capacity provider. You can override this by setting the `launchType` on the service (which we were always providing). To allow this functionality I've added a new config option `useClusterDefaultCapacityProviderStrategy` which allows this setup.

> Note: The default capacity provider strategy requires that the cluster
be configured separately and if you forget that step then the service will fail to launch.

fixes #1093

## Compatibility
- [ ] No public behavior/API change
- [X] Public behavior/API changed (describe impact)
    - New options exposed, backwards compatible.

## Generated Artifacts
- [ ] N/A
- [X] Ran regeneration (`make generate` or scoped equivalent) and committed generated diffs

## Validation
- [X] `make test_provider`
- [X] `make lint`
- [X] `yarn --cwd awsx-classic lint` (if `awsx-classic/**` changed)
- [X] `make test` (if integration-impacting; requires AWS creds/resources)
- [ ] Other targeted checks listed in PR body


pose:
@pulumi-eon Any chance we could use YAML-based tests instead of increasing our surface area of JS/TS examples/integration tests?

pose:
@corymhall let me know if you agree with me. I think the trade-off is acceptable ☝️

corymhall:
@pose looks good to me!

pose (approved):

PR #2135, opened 2026-09-03
Replace vercel/pkg with enhanced Node.js SEA

### LINKED ISSUE #2129: Use Node.js 24 to build AWSX
## Change

Set Node.js 24 in the root `mise.toml` so it overrides the older Node.js version generated in `.config/mise.toml`.

## Why

AWSX should use the current Node.js runtime for repository builds and tests. A root Mise override lets AWSX update independently without editing generated `ci-mgmt` configuration.

### LINKED ISSUE #2130: Replace vercel/pkg and update the embedded provider runtime
## Change

Replace the archived `vercel/pkg` build and the embedded Node.js 16 runtime.

Consider the options already identified:

- native Node.js SEA, which can embed Node.js 26 but requires bundling and cross-platform packaging;
- `@yao-pkg/pkg`, the maintained `vercel/pkg` replacement, which currently documents Node.js 22 and 24 targets;
- Bun `--compile`, which supports standalone executables but embeds Bun instead of Node.js.

Select and implement the first option that supports the complete AWSX provider and release-platform matrix.

## Why

Node.js 16 is no longer supported, and `vercel/pkg` is archived. The provider currently ships an obsolete runtime that no longer receives security or compatibility updates.

corymhall:
We have been using `vercel/pkg` to create provider binaries for awsx. `vercel/pkg` has been deprecated for a while now and only supports up to Node.js 16. This prevents us from upgrading past `v3.200.0` of pulumi (which requires Node.js 22) and means that we are shipping the provider with a version of nodejs that is no longer supported.

This PR replaces it with `@yao-pkg/pkg` (a maintained fork of `vercel/pkg`) enhanced SEA and Node.js 24.20.0. This comes with some tradeoffs:

**Benefits**
- SEA mode uses the official Node.js executables and Node's SEA mechanism
- Native Node.js SEA can embed one JavaScript entrypoint into an executable, but it does not handle the packaging (e.g. it can't create that single JavaScript entrypoint). `@yao-pkg/pkg` handles the packaging (e.g. finds dependencies through `require` calls, embeds application files like schema.json, etc) while still using native Node.js SEA.

**Downsides**
- `@yao-pkg/pkg` is still a 3rd party project, but this is no worse than `vercel/pkg`
- Official Node.js SEA builds dynamically linked linux executables, not static (see more on this below)

**Dynamic vs Static**

`@yao-pkg/pkg` supports two methods of packaging, `standard` and `SEA`. 

- `standard` mode works the same way as `vercel/pkg` and depends on a separate `pkg-fetch` project that publishes a patched version of Node.js. This allows `pkg` to build a `linuxstatic` executable.
- `SEA` uses the official Node.js which means it is not able to convert the official Linux executable into a static one.

This doc has more details on the difference between the two modes https://yao-pkg.github.io/pkg/guide/sea-vs-standard. `SEA` seems like the better option, but it does mean that for Linux we now require some extra dependencies.

**Linux Requirements**
- Linux kernel 4.18 or newer and glibc 2.28 or newer. 
- Standard Debian, Ubuntu, RHEL, and similar glibc distributions should work.
- A musl-only system like Alpine does not work without installing additional glibc dependencies

**MacOS Requirements**
- macOS 13.5 or newer

**Windows Requirements**
- Windows 10 or newer, or Windows Server 2016 or newer

I think this is an acceptable tradeoff because Pulumi already expects glibc-based Linux environments, and we expect most current users to meet these OS requirements. I've added a compatibility workflow that confirms that all six artifacts start on current native runners and that both Windows artifacts have valid signatures. This does not prove that no users will be affected. If this causes material customer problems, we will need to roll it back and reassess the packaging approach.

fixes #2130, fixes #2129


corymhall:

- #2135 ◀
    - #2138
        - #2140
            - #2141

julienp:
>  because Pulumi already expects glibc-based Linux environments

I don't think that is true, the CLI is built with `CGO_ENABLED=0`, and I can start the current release of `awsx` provider on `alpine:3.20`. I only tested to the point where it prints the port, but that should be enough to establish that it doesn't need glibc.

It might be that is still the right way to go about things, I think using Node.js standard standalone executables is an improvement.

Maybe we could have a build for the [alpine target](https://yao-pkg.github.io/pkg/guide/targets#alpine-musl), but we'd have to update the CLI's download to optionally consider libc/musl. Today we base it on `runtime.GOOS` and `runtime.GOARCH`

awsx/scripts/build.sh
```diff
@@ -50,4 +50,8 @@ yarn tsc
 cp ${SCHEMA} bin/schema.json
 cp package.json bin/package.json
 yarn --cwd bin version --new-version "${VERSION}" --no-git-tag-version
-yarn run pkg . --no-bytecode --public-packages "*" --public --target "${TARGET}" --output "${OUT}"
+yarn run pkg . --sea --target "${TARGET}" --output "${OUT}"
```
pose, line 53: @corymhall I would check that given that NodeJS 24 [already includes ways out of the box to do SEA](https://nodejs.org/api/single-executable-applications.html) if we need a dependency at all here and we can just simplify and use NodeJS built-in capabilities.

corymhall: Node’s SEA feature is not a complete packaging solution yet (and I don't know if it ever will be). For Node 24 it requires you to bundle a single JavaScript entrypoint yourself and pass that to the build. This means that without `@yao-pkg/pkg`, we would have to add our own bunder which would move a lot more packaging logic into this repo. `@yao-pkg/pkg --sea` does all that stuff for us while using the stock Node.js binaries.

corymhall:
> > because Pulumi already expects glibc-based Linux environments
> 
> I don't think that is true, the CLI is built with `CGO_ENABLED=0`, and I can start the current release of `awsx` provider on `alpine:3.20`. I only tested to the point where it prints the port, but that should be enough to establish that it doesn't need glibc.
> 
> It might be that is still the right way to go about things, I think using Node.js standard standalone executables is an improvement.
> 
> Maybe we could have a build for the [alpine target](https://yao-pkg.github.io/pkg/guide/targets#alpine-musl), but we'd have to update the CLI's download to optionally consider libc/musl. Today we base it on `runtime.GOOS` and `runtime.GOARCH`

I think I was basing it on the `pulumi/pulumi-base:latest` docker image, but yeah that doesn't mean that Pulumi itself doesn't support alpine.

pose (approved):

awsx/nodeCompatibility.ts
```diff
@@ -0,0 +1,273 @@
...
+      reason: "kernel-version",
+      message: `This Linux kernel is below the Node.js ${nodeRuntimePolicy.warningNodeMajor} support baseline (${nodeRuntimePolicy.minimumLinuxKernel}). Older kernels may still work.`,
+    });
+  }
+  if (facts.distribution === "alpine") {
+    risks.add({
+      reason: "musl",
+      message:
+        "Alpine was detected. Musl-only Linux environments, including Alpine, might no longer work.",
+    });
```
pgavlin: I think we can probably be more precise in detecting glibc availability. Looks like node settled on `/usr/binldd` for detecting `musl` specifically: https://github.com/nodejs/node/issues/48204

feels like the more accurate question to as is "is `glibc` available on the current system" (and maybe its version). maybe that's also doable with `ldd`?

PR #2138, opened 2026-09-04
Modernize AWSX provider TypeScript tooling

corymhall:
## Summary

The AWSX provider uses TypeScript 4, Node.js 18 definitions, TSLint, and Prettier. This moves the provider to TypeScript 6 and Node.js 24 definitions, replaces TSLint with Oxlint, and replaces Prettier with Oxfmt.

The type generator continues to use the TypeScript compiler API. Its factory calls are updated for the TypeScript 6 signatures, and the generated schema types are refreshed. The compiler now emits ES2022 for the Node.js 24 runtime introduced by the parent PR.

This PR is stacked on #2135.

Fixes #2131.

## Compatibility
- [x] No public behavior/API change
- [ ] Public behavior/API changed (describe impact)

## Generated Artifacts
- [ ] N/A
- [x] Ran scoped type regeneration (`yarn --cwd awsx gen-types`) and committed generated diffs

## Validation
- [x] `make test_provider`
- [ ] `make lint`
- [ ] `yarn --cwd awsx-classic lint` (if `awsx-classic/**` changed)
- [ ] `make test` (if integration-impacting; requires AWS creds/resources)
- [x] TypeScript compile, type generation, Oxlint, Oxfmt, frozen install, lockfile deduplication, and `git diff --check`

## Risk

The provider JavaScript emission target changes from ES2017 to ES2022. The parent PR embeds Node.js 24, which supports this target. The public schema and SDK API are unchanged.

## Rollback

Revert this PR to restore TypeScript 4, Node.js 18 definitions, TSLint, and Prettier.

PR #2140, opened 2026-09-04
Update pulumi/pulumi to v3.261.0

corymhall:
Update pulumi/pulumi to v3.261.0

Update the pinned Node.js dependency and the Pulumi Go modules together,
following the former weekly dependency update process. Refresh lockfiles,
module metadata, and generated SDK outputs for the new dependency set.

Pulumi now persists local component inputs. Update the provider type
generator to register public arguments instead of its internal options, and
adapt VPC initialization to the current component hook contract. This avoids
sending dependency-resource options through protobuf during remote component
construction.

Pulumi v3.261.0 raises the required Go version to 1.26.6. Update the
repository's golangci-lint override to a release built with Go 1.26.


corymhall:

- #2135
    - #2138
        - #2140 ◀
            - #2141

PR #2141, opened 2026-09-04
Use npm for AWSX provider tooling

corymhall:
Use npm for AWSX provider tooling

Replace the AWSX provider Yarn lock and commands with npm equivalents so
local builds, CI, Renovate, and AWS dependency updates use one reproducible
npm dependency tree.

Remove the Yarn-only deduplication tool and unused TypeDoc dependency. The
updated Pulumi peer range and removal of TypeDoc let npm enforce peer
requirements without legacy-peer-deps.


corymhall:

- #2135
    - #2138
        - #2140
            - #2141 ◀

